PR #78, opened 2013-04-28
Paginate tag

neilkinnish:
Hi,

I've noticed a couple of people asking for pagination support in dotliquid. It's close to the shopify implementation with some differences due to the fact that this could be used with any type of collection or array not just a product collection.

It does require that user pass through a request querystring for the page in the hash.

Thanks

Neil


Woland2k:
Anyone knows how pagination works in shopify? It basically does very similar thing, except no collection is getting sliced and somehow for tag knows about collection limits after. The issue is how do I know in C# code how large the collection is going to be? Or how does for tag knows that I only need to display 10 items out of 20 items collection?
(there are 20 products in collection)

{% paginate collection.products by 10 %}  
  {% for product in collection.products %}
    
  {% endfor %}
{% endpaginate %}

haf:
Hi there, sorry about the (very) slow response. Could you please rebase on master?

haf:
Closing due to lack of response.

PR #85, opened 2013-09-13
Refactor Variable.cs into some reusable expression evaluation classes

mikebridge:
I wanted to write some custom tags with the same sort of expressions that are used in a DotLiquid variable, but the expression-parsing and expression-evaluating code in Variable wasn't available outside that class and I didn't want to have to recreate all that code myself.

This modification breaks out the parsing and expression evaluation responsibilities from the variable class and makes them available for other custom tags so that I can do this:

```liquid
     {% my_custom_tag person.address | toupper %}
```

The tests in CustomExpressionTagTests show a few examples.

I also removed a try/catch block which was catching an exception that was no longer thrown, refactored some of the longer methods and moved and cleaned up some of the affected tests.

Note: I don't have VS2008 installed any more so I can't update or test the old VS2008 .sln file.

Also, the TestFormat("%W") test was failing before---I would have fixed it but I'm not sure it's testing a valid C# format string.

Hope this looks ok!


haf:
Hi there, sorry about the (very) slow response. Could you please rebase on master?

laedit:
@haf if the rebase is the only thing missing, I'd like to try the new "rebase and merge" button, is that ok with you?

PR #87, opened 2013-11-02
Added int overload for catch all method

tonybadguy:
Added support for drop[123] in addition to the existing drop["abc"].

Added overload: 
BeforeMethod(int index)

Previously, the catchall method only support strings:
BeforeMethod(string method)


tonybadguy:
@tgjones what do you think of this? 
We've been using this on our fork, but would be good if it could be incorporated into the official dotliquid

tgjones:
@tonybadguy I didn't merge this originally because I didn't fully understand it - my bad :) Specifically, I wasn't sure what DotLiquid does *now* if you try to do `drop[123]`. Does it pass `"123"` as a string to `BeforeMethod(string method)`? Or it just doesn't work at all?

tonybadguy:
@tgjones Right now, ```drop[123]``` will not work at all. It will return ```"Liquid error: Unable to cast object of type 'System.Int32' to type 'System.String'."``` 

After this change, it will call ```BeforeMethod(int index)```.

tgjones:
Ruby Liquid actually allows the key to be anything:
https://github.com/Shopify/liquid/blob/master/lib/liquid/drop.rb#L34

If it's not a string that is the name of an invocable method, then it gets passed as-is to `BeforeMethod`. So the correct thing to do in DotLiquid is probably to change the signature of `BeforeMethod` from `object BeforeMethod(string)` to `object BeforeMethod(object)`. Since that's a breaking change, it should wait until v2, if and when that arrives.

tonybadguy:
I see, and you want to maintain a single ```BeforeMethod``` without overloads like in Ruby Liquid. 

What if, for V1, we just do a ToString() to the key before invoking BeforeMethod(string)? This would be a non-breaking change, matches closer to what Ruby Liquid is doing, and allows for non-string keys.

We can make a one-line change here:
https://github.com/dotliquid/dotliquid/blob/master/src/DotLiquid/Drop.cs#L132-L134

From:
```
public object InvokeDrop(object name)
{
    string method = (string)name;
```

To:
```
public object InvokeDrop(object name)
{
    string method = name.ToString();
```

PR #94, opened 2014-02-18
Allow non-global template configuration

### LINKED ISSUE #93: Allow non-global template configuration
This proposal is intended to provide a way to use multiple rendering configurations in an application, so that you don't have to use the global tag/filter/safe type/naming convention/file system configuration.

Currently template configuration is done as follows:

```csharp
Template.RegisterTag<MyTag>("my");
Template.NamingConvention = new CSharpNamingConvention();
// etc.

var template = Template.Parse("some template text");
var result = template.Render();
```

My intention is to allow the following syntax, while keeping the former for compatibility and simplicity for those who only need global registration.

```csharp
var config = new LiquidConfiguration();
config.RegisterTag<MyTag>("my");
config.NamingConvention = new CSharpNamingConvention();
// etc.

var template = Template.Parse("some template text", config);
var result = template.Render();
```

I also thing it would be interesting to provide a more fluent configuration, such as:

```csharp
var config = new LiquidConfiguration()
    .RegisterTag<MyTag>("my")
    .SetNamingConvention(new CSharpNamingConvention());
```

Any ideas? I intend on sending a pull request ASAP.

rdumont:
Fixes #93. This is what I got:

### Before
```cs
Template.RegisterFilter(typeof(SomeFilters)); // Filter registration
Template.RegisterValueTypeTransformer(type, func); // Value type transformer
Template.RegisterSafeType(type, allowedMembers); // Safe type
Template.RegisterTag<MyTag>("my");  // Tag registration
Template.FileSystem = new MyFileSystem(); // File system

Template.NamingConvention = new SomeNamingConvention();

var template = Template.Parse("the template");
var result = template.Render();
```

### After
```cs
var config = new TemplateConfiguration();
config.RegisterFilter(typeof(SomeFilters)); // Filter registration
config.RegisterValueTypeTransformer(type, func); // Value type transformer
config.RegisterSafeType(type, allowedMembers); // Safe type
config.RegisterTag<MyTag>("my");  // Tag registration
config.FileSystem = new MyFileSystem(); // File system

Template.NamingConvention = new SomeNamingConvention(); // The naming convention was kept global for now, since it could too heavily impact performance and caching

var template = Template.Parse("the template", config);
var result = template.Render();
```

The global methods were all kept so that it doesn't break the interface. Also, the syntax for each configuration was kept the same (including overloads), except for being on the `TemplateConfiguration` instance.

I ran performance tests and no difference was detected on the time taken to run all the tests. Is this good to go?

### Motivation
In case anyone fails to understand the importance of this feature, here is my real-world scenario.

**1. Support different schemas**: I have different schemas of my application running at the same time on the same application. Each schema supports completely different structures for the entities and have tags with different behaviors, although both run on DotLiquid. Therefore, I need to be able to conditionally use a different DotLiquid configuration for each schema.

**2. Validation**: My application supports editing entities through a REST API, so I need to validate their contents so that they don't break at run time. I've found that the best way to do this is by registering fake tags that, instead of fully running their production code, just verify the integrity of their contents and configuration.

**3. Testing**: I don't want my unit tests to depend on a global, static, configuration. If I am testing a class, the test configuration should be as isolated and as close to my tests as possible.

### DIFF
--- a/src/DotLiquid.Tests/BlockTests.cs
+++ b/src/DotLiquid.Tests/BlockTests.cs
@@ -67,8 +67,8 @@ namespace DotLiquid.Tests
 		[Test]
 		public void TestWithCustomTag()
 		{
-			Template.RegisterTag<Block>("testtag");
-			Assert.DoesNotThrow(() => Template.Parse("{% testtag %} {% endtesttag %}"));
+		    var config = new TemplateConfiguration().RegisterTag<Block>("testtag");
+			Assert.DoesNotThrow(() => Template.Parse("{% testtag %} {% endtesttag %}", config));
 		}
 	}
 }
--- a/src/DotLiquid.Tests/ContextTests.cs
+++ b/src/DotLiquid.Tests/ContextTests.cs
@@ -278,9 +278,11 @@ namespace DotLiquid.Tests
 		[Test]
 		public void TestOverrideGlobalFilter()
 		{
-			Template.RegisterFilter(typeof(GlobalFilters));
-			Assert.AreEqual("Global test", Template.Parse("{{'test' | notice }}").Render());
-			Assert.AreEqual("Local test", Template.Parse("{{'test' | notice }}").Render(new RenderParameters { Filters = new[] { typeof(LocalFilters) } }));
+		    var config = new TemplateConfiguration();
+		    config.RegisterFilter(typeof (GlobalFilters));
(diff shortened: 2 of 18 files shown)


haf:
Closing due to lack of activity.

haf:
@axm Correct, feel free to have a go at it

axm:
@haf All right

PR #112, opened 2014-09-02
Break and continue tags

### LINKED ISSUE #100: Support break and continue statements in for loops
Implement `break` and `continue` tags to be used inside `for` loops. As seen in Shopify/liquid#140.

rdumont:
Fixes #100. Here is the change summary:

**New exception**: `InterruptException` - this allows interruptions to be sent from anywhere. This exception, and any other that inherits from it, will not be caught by the template rendering, no matter if `RethrowErrors` is set to `True` or `False`.

**New tag**: `break` - this will act as a normal `break` statement inside a `for` loop, interrupting its execution.

**New tag**: `continue` - this will act as a normal `continue` statement inside a `for` loop, interrupting the current iteration and continuing to the next.

Both `break` and `continue` tags, if used outside of a `for` loop, wiil abort the entire rendering, without any errors.

### DIFF
--- a/src/DotLiquid.Tests/ExceptionHandlingTests.cs
+++ b/src/DotLiquid.Tests/ExceptionHandlingTests.cs
@@ -17,6 +17,11 @@ namespace DotLiquid.Tests
 			{
 				throw new SyntaxException("syntax exception");
 			}
+
+		    public void InterruptException()
+		    {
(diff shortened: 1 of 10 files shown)


tgjones:
This looks great - thanks @rdumont. (And I see you based it on the Ruby implementation of these tags, which is always nice.)

PR #115, opened 2014-11-20
Update README.markdown

airtonix:
Present more compelling reasons.


haf:
Merged, thanks @airtonix

PR #125, opened 2015-01-14
Support Slice

jogibear9988:


tgjones:
@rdumont what do you think? This seems like a good change to me - it implements something already in Ruby Liquid, and it has tests. If you're happy with it, you could go ahead and merge it.

rdumont:
@tgjones yes, looks simple enough to me. Merging it!

@jogibear9988 thanks for this contribution! For future reference, though, was that merge from upstream really necessary? The PR would've been cleaner without it.

tgjones:
@rdumont Thanks!

jogibear9988:
@rdumont:
 Sorry, I did not correctly make a Branch for each of my change, I fixed the things I needed and didn't flollow the right workflow! Thanks for merging!

Can you also merge #127 and #128?

After that I've a few more patches in my Branch, most important for me is the change that string filters also work on other types.

PR #127, opened 2015-01-14
Dictionary access change

jogibear9988:


haf:
Hi @jogibear9988  can we do #128 instead?

jogibear9988:
Why did you close this? it's also not yet implemented!

haf:
@jogibear9988 Because it would seem #128 is identical?

jogibear9988:
No, the changes to Context.cs are missing in the other one

haf:
Ok, could you please rebase it against master then?

jogibear9988:
will do. monday i think

haf:

PR #130, opened 2015-01-16
Performance Enhancements

SteveLillis:
**Performance related changes:**
These changes make the template rendering process, on average, **2.7 times faster**.

A full write-up of the changes and their associated performance timings can be found here: 
http://stevelillis.blogspot.co.uk/search/label/DotLiquid

To avoid the use of exceptions as program flow, I had to make a minor breaking change to the Tag class.  Anyone who has written their own tags will have to make the following changes to their tag code:
- The return Type of the Render method is now ReturnCode.
- Wherever return is used, return ReturnCode.Return instead.
- Whenever the result of RenderAll is not ReturnCode.Return, return that result immediately.

An example of how a tag should behave regarding the new ReturnCode behaviour can be seen in Block.RenderAll:
https://github.com/SteveLillis/dotliquid/blob/master/src/DotLiquid/Block.cs

**Non-performance related changes:**
 - Fixed a minor bug in the TryOnline website, which was not decoding HTML and therefore failing to parse less than and greater than signs.
 - Added support for Break and Continue to .NET 3.5 solutions.

**Note:**
Inconsistent tab/space indentation across the solution and within files made it difficult to keep the pull request clean.  I recommend that we use spaces throughout as they render more consistently in non-IDE environments.  

Let me know if you agree and I'll untabify the whole solution as a separate pull request.


SteveLillis:
Closed request and resubmitted with less spaces vs tab noise

PR #131, opened 2015-01-16
Performance Enhancements (minimal request noise)

SteveLillis:
**Performance related changes:**
These changes make the template rendering process, on average, **2.7 times faster**.

A full write-up of the changes and their associated performance timings can be found here: 
http://stevelillis.blogspot.co.uk/search/label/DotLiquid

To avoid the use of exceptions as program flow, I had to make a minor breaking change to the Tag class.  Anyone who has written their own tags will have to make the following changes to their tag code:
- The return Type of the Render method is now ReturnCode.
- Wherever return is used, return ReturnCode.Return instead.
- Whenever the result of RenderAll is not ReturnCode.Return, return that result immediately.

An example of how a tag should behave regarding the new ReturnCode behaviour can be seen in Block.RenderAll:
https://github.com/SteveLillis/dotliquid/blob/master/src/DotLiquid/Block.cs

**Non-performance related changes:**
 - Fixed a minor bug in the TryOnline website, which was not decoding HTML and therefore failing to parse less than and greater than signs.
 - Added support for Break and Continue to .NET 3.5 solutions.

**Note re: tabs vs spaces**
Inconsistent tab/space indentation across the solution and within files made it difficult to keep the pull request clean.  I recommend that we use spaces throughout as they render more consistently in non-IDE environments.  

Let me know if you agree and I'll untabify the whole solution as a separate pull request.

[Review on Reviewable](https://reviewable.io/reviews/dotliquid/dotliquid/131)


rdumont:
@SteveLillis these are awesome changes, thank you! But unfortunately they break the library API and therefore must be postponed to version 2.0. What do you think about it?

SteveLillis:
@rdumont I'm on board with that, unexpectedly springing API changes on people is the last thing you'd want. Do you have an estimated delivery date for version 2.0?  Anything I can do to help?

rdumont:
If you can help? Yes. How? We don't know yet. :smile: 

So far 2.0 is still just a concept, but there are already a few PRs that should go into it. If you'd like, join the [Gitter room](https://gitter.im/dotliquid/dotliquid), where we can plan the next major release. I'm really looking forward to it.

nelero:
So, a year later, are those great improvments can be found in the v1.x branch ?
Only performance improvements part cause it seems to break the library API ?
And what about the 2.0 version ? :-) Can't wait ! :-)

SteveLillis:
Reviewed 54 of 54 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dotliquid/dotliquid/131)*

bryan07:
I been using @SteveLillis version for some time, the adjustments to my existing Tag classes where minimal.

haf:
If this is given a new PR to master I'll release it as soon as there's build code in the repo that pushes nugets. If you want this feature in v2.0, that release probably happens monday next week (2016-07-25)

PR #140, opened 2015-03-16
Added ability so Slice can deal with negative offsets

lubos:
When you go to https://github.com/Shopify/liquid/wiki/Liquid-for-Designers, you can see that "slice" function should be able to deal with negative offsets as well.


rdumont:
Thanks for the fix! :)

PR #145, opened 2015-04-30
Ensured that numbers are output using the format provider of the current...

bradphelan:
On my dev machine the culture is german and by default all numbers are printed like

    3,14159
    
however my template was output to a file that should use the invariant culture. Numbers
should use decimal points like

    3.14159
    
Previous to this fix numbers were being output using

   output.ToString()
   
with no format provider passed. Thus the current culture would be used.
    
This fix ensures the numbers are output using the same format provider as stored in the textwriter used to generate the output.

### DIFF
--- a/src/DotLiquid.Tests/OutputTests.cs
+++ b/src/DotLiquid.Tests/OutputTests.cs
@@ -1,3 +1,5 @@
+using System.Globalization;
+using System.IO;
 using NUnit.Framework;
 
 namespace DotLiquid.Tests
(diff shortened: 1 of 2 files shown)


haf:
Nice, I'd like to merge this. Could you rebase on master?

haf:

JKronberger:
Is it possible, that this fix has been removed with one of the latest versions?
We have some testcases which started to fail with version 2.0.135 and now show the following behaviour:
unbenannt

It seems to ignore the comma.

laedit:
From what I see of the history of this PR, the fix hasn't been merged to the master branch, so behavior of DotLiquid should have been the same.

Nevertheless if you want to propose a new PR with the fix, I will be happy to review it 😃

PR #148, opened 2015-05-14
Add aspnetcore50 project and build

tbolon:
Made DotLiquid compatible with aspnetcore50, by creating a new xproj and using conditional symbols to handle differences between the full .net framework and corefx.

I have edited the nuspec to include the new library. Please note that you have to use msbuild 14 to generate the dll : `msbuild DotLiquid.sln /p:Configuration=Release`.

The only limitation for now is that the ToTitleCase method does not exists in the CoreFX, and should be implemented manually in the corefx version.

*Background: I want to use DotLiquid in an corefx / aspnetcore project. My goal is to build something similar to jekyllrb using the new corefx multiplatform engine. I want to setup the same liquid/yaml/markdown setup to create pages. So, I will certainly use [Commonmark.NET](https://github.com/Knagis/CommonMark.NET/) which is already aspnetcore50 compatible. But DotLiquid was not.*


src/DotLiquid/Condition.cs
```diff
@@ -24,10 +24,10 @@ public class Condition
...
-			{ "<", (left, right) => left != null && right != null && Comparer.Default.Compare(left, Convert.ChangeType(right, left.GetType())) == -1 },
-			{ ">", (left, right) => left != null && right != null && Comparer.Default.Compare(left, Convert.ChangeType(right, left.GetType())) == 1 },
-			{ "<=", (left, right) => left != null && right != null && Comparer.Default.Compare(left, Convert.ChangeType(right, left.GetType())) <= 0 },
-			{ ">=", (left, right) => left != null && right != null && Comparer.Default.Compare(left, Convert.ChangeType(right, left.GetType())) >= 0 },
+			{ "<", (left, right) => left != null && right != null && Comparer<object>.Default.Compare(left, Convert.ChangeType(right, left.GetType())) == -1 },
+			{ ">", (left, right) => left != null && right != null && Comparer<object>.Default.Compare(left, Convert.ChangeType(right, left.GetType())) == 1 },
+			{ "<=", (left, right) => left != null && right != null && Comparer<object>.Default.Compare(left, Convert.ChangeType(right, left.GetType())) <= 0 },
+			{ ">=", (left, right) => left != null && right != null && Comparer<object>.Default.Compare(left, Convert.ChangeType(right, left.GetType())) >= 0 },
```
tbolon: `Comparer` does not exists in corefx. `Comparer<object>` is supposed to be equivalent.

src/DotLiquid/Context.cs
```diff
@@ -3,6 +3,9 @@
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
+#if NETCore
+using System.Reflection;
+#endif
```
tbolon: Required because in corefx, Advanced type operations are implemented with extensions.

src/DotLiquid/Exceptions/LiquidException.cs
```diff
@@ -2,9 +2,16 @@
...
 {
-	[Serializable]
-	public abstract class LiquidException : ApplicationException
-	{
+#if !NETCore
+    [Serializable]
+#endif
+    public abstract class LiquidException :
+#if NETCore
+        Exception
```
tbolon: ApplicationException does not exists in corefx.

src/DotLiquid/Properties/AssemblyInfo.cs
```diff
@@ -23,7 +23,9 @@
 
 // The following GUID is for the ID of the typelib if this project is exposed to COM
 
+#if !NETCore
 [assembly: Guid("75d1dc49-0097-4f60-b5fd-0e43d213b38b")]
```
tbolon: Not supported

src/DotLiquid/Properties/Resources.NETCore.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace DotLiquid.Properties
+{
+#if NETCore
+    // this class is only used to setup the ResourceManager in Liquid
```
tbolon: ResourceManager in Liquid is based on Type for resource lookup. But because the Resources.Designer.cs is not supported in corefx, I have replaced it with an empty class to help ResourceManager locate the .resources item in the assembly.

src/DotLiquid/StandardFilters.cs
```diff
@@ -81,10 +81,14 @@ public static string Capitalize(string input)
...
 				return input;
 
-			return string.IsNullOrEmpty(input)
-				? input
-				: CultureInfo.CurrentCulture.TextInfo.ToTitleCase(input);
-		}
+            return string.IsNullOrEmpty(input)
+                ? input
+#if NETCore
+                : input; // not supported, should implement on our own
```
tbolon: TODO

tgjones:
@tbolon I like the general direction you're going in here. I've been wondering, though, would it better to turn DotLiquid into a PCL? (or at least have it as an option) Then it should work on both CoreFX and the full .NET framework (I think).

tbolon:
@tgjones no PCL != corefx. Eventually, you could replace the .NET 4/4.5 release with a portable one, but often the best method is to add a specific portable project which targets all the most common denominator platforms. I could eventually create another csproj targeting a portable version, in another PR.

You can see how the CommonMark.net project has structured his source code for example.

tgjones:
> no PCL != corefx

I know that, but if we had a PCL, we could target most, if not all, of the platforms we want to target, with just a single DLL.

tbolon:
@tgjones Yes, sorry. For NET35 you will need a specific project. But you are right, the NET40/NET45 libraries could be merged in single a portable class library.

For me you could get rid of the 2008-2010 sln and convert DotLiquid-2008.csproj into DotLiquid-NET35.csproj and include it in the single remaining solution.

itlackey:
Thought I would throw my $0.02 in here. I give a +1 to the corefx and the pcl, both will be very useful.

itlackey:
Btw @tbolon it appears we have the same goal, a Jekyll (ish) aspnetcore project.
I have a start on this, maybe we should consider joining forces... Let me know if you are interested.

tbolon:
@itlackey yes, I have also a project, named [giki](https://github.com/tbolon/giki). It's only a prototype, because I plan using it to migrate a screwturn wiki into a static website using git as a way to accept contributions.

I have implemented also some custom markdown extensions to allow easily to link other pages using [[ ]] syntax. I have also implemented basic wikitable support. But all of this could be changed as optional plugins to allow a more strict rendering. Anyway, feel free to contact me directly at thomasbolon at gmail.

haf:
Hi @tbolon and @itlackey sorry about the delay in getting to this. Could you please rebase this on master and we start start to get it going?

haf:

tbolon:
Hi, I will certainly create a new PR, as there was a lot of changes between my original PR and the new Framework.

PR #150, opened 2015-05-19
Add ITagFactory

### LINKED ISSUE #146: Add a RegisterTag method
I use MEF to allow the user of my app to add plugins and for DotLiquid add Tags and Filters.
I also use MEF to do all dependency injection work and avoid the use of static classes.

But since DotLiquid.Template.RegistarTag works only with type and instanciate the tag itself the dependency injection can't work.

Is it possible to add a RegisterTag method which takes directly a variable of type Tag?

If you are ok with the idea I can propose a PR.

laedit:
Fix #146

I am based on the idea of a factory for a particular tag and not a factory for any tags.

Also I have keeped "GetTagType" because it is public but I don't know if it still has any interest.


laedit:
Is there any chance to see the PR merged soon?
Just to know if I must wait a little bit more or use a fork in the meantime :smiley:

src/DotLiquid/Template.cs
```diff
@@ -27,32 +27,52 @@ public class Template
...
+			Tags = new Dictionary<string, Tuple<ITagFactory, Type>>();
             SafeTypeTransformers = new Dictionary<Type, Func<object, object>>();
 			ValueTypeTransformers = new Dictionary<Type, Func<object, object>>();
 		}
 
 		public static void RegisterTag<T>(string name)
 			where T : Tag, new()
 		{
-			Tags[name] = typeof(T);
+            var tagType = typeof(T);
```
haf: It would seem this PR mixes indentation.

laedit: I'm glad to hear from this PR :)
If this is blocking I will gladly fix it.

haf: I'm author of https://suave.io, and since we (and I) depend on dotLiquid I'm trying to steer it towards something less zombie ;).

Would you be interested in being part maintainer of this project?

tgjones: Thanks @laedit - I've added you to the dotliquid organisation. Any time you can put into triaging pull requests and issues would be gratefully received!

haf:
@laedit I think we're starting to move now. Could you please rebase this on master?

laedit:
I will do it ASAP, but probably not before this week end.

haf:
@laedit thank you!

laedit:
I am not familiar with rebase but I think it is done.

haf:
Nice, thank you!

PR #155, opened 2015-08-20
Added FromStrictDictionary method

yorek:
This is useful if you want to be sure that when a template uses a variable that does not exist in the Hash you get an error


yorek:
Just found a little bug. With this approach includes stops working :( I'm trying to figure out why.

PR #156, opened 2015-08-21
Unknown variables are now reported in the Error collection

yorek:
When a variable used in the template cannot be resolved, an new VariableResolutionException is added to the Error list so that it is possible to show to the end user which variables are evaluated to null because they do not exists.


src/DotLiquid.Tests/FilterTests.cs
```diff
@@ -169,7 +169,14 @@ public void TestJoin()
 			Assert.AreEqual("1 2 3 4", new Variable("var | join").Render(_context));
 		}
 
-		[Test]
+        [Test]
```
haf: It would seem this PR mixes indentation.

yorek:
My identation is set to 4...the original source code had it set to 8 maybe?

haf:
Hi @yorek Could you please rebase on master?

haf:

Issue #164, opened 2015-10-15
Issue with 'An item with the same key has already been added.'


avishnyakov:
Hello team, 

We are having that issue for custom registered object sometimes.

```
An item with the same key has already been added

   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector, IEqualityComparer`1 comparer)
   at DotLiquid.TypeResolution..ctor(Type type, Func`2 filterMemberCallback)
   at DotLiquid.DropProxy.CreateTypeResolution(Type type)
   at DotLiquid.DropBase.get_TypeResolution()
   at DotLiquid.DropBase.InvokeDrop(Object name)
   at DotLiquid.Context.LookupAndEvaluate(Object obj, Object key)
   at DotLiquid.Context.Variable(String markup)
   at DotLiquid.Context.Resolve(String key)
   at DotLiquid.Variable.RenderInternal(Context context)
   at DotLiquid.Tags.Assign.Render(Context context, TextWriter result)
   at DotLiquid.Block.<>c__DisplayClass1.<RenderAll>b__0(Object token)
```

What might be a clue to resolve or troubleshoot that?

PR #165, opened 2015-10-27
Fixed Issue with 'An item with the same key has already been added.' #164 

avishnyakov:
Fixed Issue with 'An item with the same key has already been added.' #164 

Both methods and props have to be checked on  mi.DeclaringType == type - meaning, only CURRENT entity props/methods.

If not, then a duplicated of the members / methods can be included into the result dictionary for the following POCOs:
```
  public class Entity 
    {
        public string Name { get; set; }
    } 

 public class MyEntity :Entity
    {
        public new string Name { get; set; }
    }
```


haf:
@avishnyakov Hi there, we've normalised spacing throughout; could you rebase this on master?

PR #170, opened 2016-01-26
Fix issue with RubyNamingConvention when the Culture is set to Turkish

Elph:
When the underlying Thread culture is set to Turkish, the RubyNamingConvention returns unexpected results in some cases. I detected the problem with turkish language, but it seems posible to happen with another. 

Added a test to simulate the issue, and a easy fix that solves, using ToLowerInvariant instead ToLower.


tgjones:
Looks great, thank you!

PR #173, opened 2016-02-12
Optimized Hash.FromAnonymousObject speed by using cached ExpressionTree

Mart-Bogdan:
+ additionally Variable.Render was treading Strings as IEnumerable, and was joining Chars in loop

Benchmark results:
Benchmark                             Mean Mean-Error   Sdev  Unit
Reflection                           0,038      0,014  0,009 ms/op
LINQ Expressions                     0,005      0,004  0,003 ms/op

Benchmakr can be downloaded from https://gist.github.com/Mart-Bogdan/aadc1b2c8743309aeddf


Mart-Bogdan:
I've run all Unit Tests, and lot of tests seems to use Hash, so I think it's safe

Mart-Bogdan:
Measurement was taken by 10 repetitive executions. On long run it will get even better performance.
First call will obliviously slower.

Same fix can be applied to DropBase.InvokeDrop   In this case it can be even implemented on .Net 3.5 using IF-chain, (.Net 3.5 Expressions don't support switches and Method body, only single expression)

Mart-Bogdan:
@tgjones, can you review and accept my PR?

src/DotLiquid/Hash.cs
```diff
@@ -1,15 +1,24 @@
...
 using System.Collections.Generic;
+using System.Linq.Expressions;
 using System.Reflection;
 
 namespace DotLiquid
 {
 	public class Hash : IDictionary<string, object>, IDictionary
-	{
-		#region Fields
+    {
```
haf: It would seem this mixes indentation.

Mart-Bogdan: @haf hi, I've fixed indents in new commit. It was looking good in VS, but GH seems to have 8 spaces per tab.

It's actually Odd, that whole this project uses tabs, it's better practices to use spaces everywhere, changed my commits to Tabs thou.

haf:
Hi, I've normalised all the tabs to spaces. Sorry about the churn it's causing this PR. Could you care to update again?

Mart-Bogdan:
Ok, I've merged my changes.

src/DotLiquid/Variable.cs
```diff
@@ -1,141 +1,144 @@
...
-                Name = name;
-                Arguments = arguments;
-            }
-
-            public string Name { get; set; }
-            public string[] Arguments { get; set; }
-        }
-    }
-}
+using System.Collections;
```
haf: It would seem you've changed all of this file. I suspect you've converted CRLF to LF contrary to .editoconf's wishes ;)

src/DotLiquid/Hash.cs
```diff
@@ -54,6 +54,22 @@ private static void FromAnonymousObject40(object anonymousObject, Hash hash)
             Action<object, Hash> mapper;
             if (!mapperCache.TryGetValue(type, out mapper))
             {
+                /* Bogdan Mart: Note regarding concurrency:
```
haf: Love comments like this!

src/DotLiquid/Variable.cs
Mart-Bogdan: Hmmm, that was default  Git settings on widows.
So all files should be CRLF?

haf: Yes

Mart-Bogdan: how i "love" Bill Gates for makeing line separator two characters, I'll adjust that

Mart-Bogdan:
Updated PR

haf:
Thanks!

PR #175, opened 2016-03-25
Fixes Issue with 'An item with the same key has already been added.' #164

joeburdick:
Fixes Issue with 'An item with the same key has already been added.' #164

This fixes an issue with TypeResolution in Drop.cs. If `Type.GetProperties(Bindingflags flags)` returns a list  of properties with duplicate names, `GetMemberDictionary<T>(IEnumerable<T> members)` will throw a `ArgumentException` with message "An item with the same key has already been added."

This can be reproduced calling `GetMemberDictionary `on the properties for class `Bar` with this class hierarchy:

```
    public class Foo
    {
        public string Name { get; set; }
    }

    public class Bar : Foo
    {
        public new int Name { get; set; }
    }
```

The properties returned by `Type.GetProperties(Bindingflags flags)` will have `PropertyInfo`s with duplicate names in this case because the `Name` property on `Bar` will appear as well as the `Name` property on `Foo`.

So, my solution is to filter properties by name by only selecting the property with the most derived `DeclaredType`.

### DIFF
--- a/src/DotLiquid/Drop.cs
+++ b/src/DotLiquid/Drop.cs
@@ -3,11 +3,12 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 using DotLiquid.NamingConventions;
+using DotLiquid.Util;
 
 namespace DotLiquid
 {
     /// <summary>
-    /// Configurable typing metadata collection
+    ///     Configurable typing metadata collection
     /// </summary>
     internal class TypeResolution
     {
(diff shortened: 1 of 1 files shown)


src/DotLiquid/Drop.cs
```diff
@@ -19,11 +19,11 @@ public TypeResolution(Type type, Func<MemberInfo, bool> filterMemberCallback)
             // Cache all methods and properties of this object, but don't include those 
 			// defined at or above the base Drop class.
             const BindingFlags bindingFlags = BindingFlags.Public | BindingFlags.Instance;
```
haf: Indentation is off

haf:
Could you please rebase this on master?

joeburdick:
Worth noting that #165 attempts to fix the same issue so there shouldn't be any point in merging both. 

I apologize for any formatting issues because I don't know what I'm doing.

haf:
@joeburdick thank you!

PR #176, opened 2016-03-30
Adding url_encode to standard filters

sebastienferry:


src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -29,6 +29,13 @@ public void TestUpcase()
 			Assert.AreEqual(null, StandardFilters.Upcase(null));
 		}
 
+        [Test]
```
haf: Indentation is off, right?

haf:
Could you please rebase on master?

haf:
Thanks!

PR #178, opened 2016-04-11
Fix TestFormat for %W result

grebeshkoff:


src/.vs/config/applicationhost.config
```diff
@@ -0,0 +1,1038 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
haf: Why add this file?

haf:
No response in 25 days. Reopening if response given.

PR #184, opened 2016-06-04
Option to savely use objects that do not implement ILiquidizable/Drop

PeterZbinden:
Added a bool as a optional Parameter(Default Value ensures API stays downward compatible) for all the Render-Methods in 'Template'. This gives developers the option to enable usage of objects in templates that do NOT implement ILiquidizable/Drop if they prefer to do so.

All Tests are passing without any changes.


src/.vs/config/applicationhost.config
```diff
@@ -0,0 +1,1038 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
haf: Please remove this from your PR

PeterZbinden: Removed file in new commit

haf:
Hi, sorry about the merge issues. https://github.com/dotliquid/dotliquid/pull/184/files?w=1 shows a non-whitespace diff. Perhaps you can port the changes to the shared files over from this view into your code and rebase against master?

haf:

PR #186, opened 2016-06-14
Jyeohamilton/lockcontentionfix

jyeohamilton:
Hi!  My team at Microsoft is using DotLiquid in a high scale service where hundreds of threads may be rendering templates at any given moment.  While I was running perf tests, profiling showed severe lock contention occurring in the Regex constructor, causing CPU usage to be pegged at 100% with a ton of request timeouts.  (The cause is explained in the comments of Util\R.cs.)  Here's a pull request with the fix I made.  

I searched through your Issues log and noticed that someone had asked for something similar many years time ago, but a change didn't seem necessary at the time.  I may be able to share some specific perf numbers from my tests with you if you need hard data on why we made this change, but briefly, it has reduced the CPU usage spent in lock contention from nearly 70% to less than 1%.

Let me know if you have any questions or comments!

### DIFF
--- a/src/DotLiquid.Tests/DotLiquid.Tests.csproj
+++ b/src/DotLiquid.Tests/DotLiquid.Tests.csproj
@@ -95,6 +95,9 @@
     <None Include="Formosatek-OpenSource.snk" />
     <None Include="packages.config" />
   </ItemGroup>
+  <ItemGroup>
+    <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
+  </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <!-- To modify your build process, add your task inside one of the targets below and uncomment it. 
        Other similar extension points exist, see Microsoft.Common.targets.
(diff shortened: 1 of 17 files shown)


tgjones:
This is awesome. IIRC, the issue / PR you referred to had some unrelated changes in addition to the Regex optimisations, which I didn't want to merge at that time. This PR looks great.

One thing though - at the moment, the change to `R.Scan` is a breaking change, if anybody has written custom tags that rely on that method. Please could you add back a method with the original signature:

``` csharp
public static List<string> Scan(string input, string pattern)
```

which just calls the new `Scan(string, Regex)`. For bonus points, you could add `[Obsolete]` on it, with an explanation of why people should switch to the faster implementation.

---

As you might have noticed, I'm not doing a great job of maintaining DotLiquid these days, largely because I'm not using it myself, and haven't done so for several years. I am looking for new maintainers - if you're interested, let me know.

tgjones:
By the way, pretty cool that Microsoft is using DotLiquid :) I guess you can't say which team?

jyeohamilton:
Added back the original R.Scan.

tgjones:
Thank you! 👍

jyeohamilton:
Thanks!  Will you be releasing a new version of the package on nuget.org soon?

tgjones:
I believe so. @haf has some ideas about how we can automate that, which we'll probably setup before the next NuGet version is deployed.

PR #187, opened 2016-06-16
Adding functions of one argument

haf:


src/DotLiquid/Strainer.cs
```diff
@@ -31,6 +31,7 @@ public static Strainer Create(Context context)
 
         private readonly Context _context;
         private readonly Dictionary<string, IList<MethodInfo>> _methods = new Dictionary<string, IList<MethodInfo>>();
+        private readonly Dictionary<string, Func<object, object>> _funcs = new Dictionary<string, Func<object, object>>();
```
haf: A bit of a hack; if I had Resharper on OS X I'd extract MethodInfo into being Delegates, thereby unifying MethodInfo with Func<>

PR #188, opened 2016-06-16
Jyeohamilton/listshiftfix

jyeohamilton:
Minor perf fix.  There are a couple of places in the variable resolution inner loop where List methods are used that modify the list (Shift, Reverse). However, the list does not really need to be modified (in these cases), so this is an unnecessary O(N) expense. This PR proposes reading the data in-place instead.


haf:
I think this looks nice, thank you @jyeohamilton for your contribution!

PR #193, opened 2016-07-22
Hopefully enabled automatic nuget publishing

Symphonym:


haf:
Let's see how far this gets us. In the worst case we'll make the nugets private.

Issue #179, opened 2016-04-12
Add `default` filter


miketaber:
Maybe I'm missing something here, but I don't see anywhere in the code that the default filter is supported. I've run tests using both an AnonymousObject and a custom Drop class, but neither will recognize the something like: "{{ testvarname | default: 'defaultData' }}"

I was able to get this working by building a custom Filter as follows:
```
    public static class DefaultFilter
    {
        public static string Default(string input, string @defaultValue)
        {
            return !string.IsNullOrWhiteSpace(input) ? input : defaultValue;
        }
    }

```
It seems like this should be included in the base library but I don't see it anywhere. Am I missing something?

laedit:
Thanks for the report.
It is missing in DotLiquid, can you open a PR with your filter? 😃

miketaber:
Just saw this. Did someone fix this or was it closed because you were waiting on a pull request?

haf:
Hi @miketaber – just because there was a month without a reply.

miketaber:
I'm not terribly familiar with Git, so this took me a while to get it "working". Apologies if I do anything incorrectly here.

I've issued the pull request and added the corresponding unit tests.

I do have one of the existing unit tests which is failing on my own machine related to date formatting. I don't know why that is. I fixed it and after a push, the build server broke, but apparently it's an issue on my local system. I double checked the math for the date and the unit test should return 01 but on my system it's returning 02. I've reverted that change, but it's still broken locally. Seems to be a core issue in .NET, since after doing the math by hand, it should return 01 locally and it doesn't.

laedit:
Thanks for the PR.

PR #203, opened 2016-08-29
Add default filter

miketaber:
Fix for Issue #179 - Add 'default' filter

### DIFF
--- a/src/DotLiquid.Tests/Util/StrFTimeTests.cs
+++ b/src/DotLiquid.Tests/Util/StrFTimeTests.cs
@@ -26,7 +26,7 @@ namespace DotLiquid.Tests.Util
         [TestCase("%p", Result = "PM")]
         [TestCase("%S", Result = "14")]
         [TestCase("%U", Result = "02")]
-        [TestCase("%W", Result = "02")]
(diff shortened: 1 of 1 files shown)


src/DotLiquid.Tests/Util/StrFTimeTests.cs
```diff
@@ -26,7 +26,7 @@ public class StrFTimeTests
         [TestCase("%p", Result = "PM")]
         [TestCase("%S", Result = "14")]
         [TestCase("%U", Result = "02")]
-        [TestCase("%W", Result = "01")]
+        [TestCase("%W", Result = "02")]
```
laedit: I don't think this is necessary, and it's breaking a test.

miketaber: I'm working on reverting that change on my side, but this unit test is failing on my machine, which is why I made the change to begin with. I hadn't realized there was a CI server that was validating this but it's very odd that I'm getting an incorrect result on my system. I double checked the documentation on MSDN and the date itself. The correct result is in fact 01 but I can't explain why my system indicates that it's 02.

laedit:
Thanks! 😄

Issue #167, opened 2015-12-09
allow cache template for tag include and extend


303248153:
I think the return type of  IFileSystem.ReadTemplateFile should be object not string.
In the moment tag extend will detect type is "Template" like
```
Template template = file as Template;
template = template ?? Template.Parse(file == null ? null : file.ToString());
```
but since ReadTemplateFile  return string so first expression always return null.
I will change it in my fork, but because it's a interface break I'll not create the pull request yet,

haf:
This would be great to have as a PR.

PR #209, opened 2016-10-20
allow IFileSystem return Template instance for better performance

303248153:
Previously I open #167 for this, today I refork dotliquid and do the same change.
This is a break change so it should't accept until next major update.

### DIFF
--- a/src/DotLiquid.Tests/Tags/IncludeTagTests.cs
+++ b/src/DotLiquid.Tests/Tags/IncludeTagTests.cs
@@ -9,7 +9,7 @@ namespace DotLiquid.Tests.Tags
     {
         private class TestFileSystem : IFileSystem
         {
-            public string ReadTemplateFile(Context context, string templateName)
+            public object ReadTemplateFile(Context context, string templateName)
             {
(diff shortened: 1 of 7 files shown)


bradphelan:
Can you explain why it is a breaking change? Is the external API changed or do templates generate different results. I'm trying to understand if this is the kind of change that will get noticed by a build failure ( easy to handle ) or a runtime failure with potentially different results ( hard to handle )

bradphelan:
ok. So they will get a compile error and then it's an easy fix. Thanks

sebastienros:
The intent is nice, but the suggested change is not very sound. Using an `object` type just to be able to return some potentially expected other type.

I would suggest to add another service instead, like `ITemplateProvider` which would be called before `IFileSystem`, and after the template is built. This would not break anything and be clear about the intent. The default implementation would just return `null`.

tgjones:
@sebastienros - great idea.

@303248153 I'm going to close this PR, but if you'd like to submit one that implements @sebastienros's idea of an `ITemplateProvider`, I think we'd accept that.

Issue #206, opened 2016-10-12
Is Template.Render thread safe?


hellosnow:
Hi I'm using dotliquid to generate html files, and could I ask whether Template.Render is thread safe or not?

intval:
Not at all. 
Simple templates "Hello {{ name }}" would work, but templates with variables, assignments includes and captures are unpredictable.

	[Test]
	public void TestConcurrentLiquid()
	{
		var tpl = Template.Parse("Hello {{name}}  " +
								 "{% assign val = num %} " +
								 "{%capture out%} " +
								 "   cool" +
								 "   {% assign val = num+1 %}" +
								 "{% endcapture%} " +
								 "{% if x %} {{out}} {%endif%}" +
								 "{{ val }}");

		var input = new []
		{
			new { name = "a", x = false, num = 9},
			new { name = "bla", x = true, num = 99},
		};

		// Single thread execution, for later comparison
		var expected = new[]
		{
			tpl.Render(Hash.FromAnonymousObject(input[0])),
			tpl.Render(Hash.FromAnonymousObject(input[1]))
		};

		// Crashes with different errors or inconsistent output
		Parallel.For(0, 200000, new ParallelOptions {MaxDegreeOfParallelism = 8}, i =>
			{
				Assert.AreEqual(expected[i % 2], tpl.Render(Hash.FromAnonymousObject(input[i%2])));
			}
		);

	}

On a side note, assignment of `val = x + 1` doesn't work either ;/

hellosnow:
Thanks for the sample! Then I'll handle multiple threads in my side.

microalps:
@daviburg I believe this issue is safe to close?

PR #210, opened 2016-10-20
make template instance thread safe, fix #154

### LINKED ISSUE #154: Cached template not working
It seems according to documentation that caching parsed template should work across multiple threads (as in web application) and the same template can be reused many times instead of being parsed. However it doesn't seem to be the case as collection is modified error is generated in the following test:

        [Test]
        public void TestCachedTemplateRender()
        {
            Template template = Template.Parse(@"{% assign foo = 'from instance assigns' %}{{foo}}");

			var parallelOptions = new ParallelOptions
			{
				MaxDegreeOfParallelism = 30
			};

            Parallel.For(0, 10000, parallelOptions, (x) => Assert.AreEqual("from instance assigns", template.Render()));
        }

Anyone else experience the same issue?

303248153:
This commit fix template instance thread safe issue, see #154 and #206.
This is a break change so it should't accept until next major update.

### DIFF
--- a/src/DotLiquid.Tests/ExceptionHandlingTests.cs
+++ b/src/DotLiquid.Tests/ExceptionHandlingTests.cs
@@ -31,9 +31,6 @@ namespace DotLiquid.Tests
             Assert.DoesNotThrow(() => { template = Template.Parse(" {{ errors.syntax_exception }} "); });
             string result = template.Render(Hash.FromAnonymousObject(new { errors = new ExceptionDrop() }));
             Assert.AreEqual(" Liquid syntax error: syntax exception ", result);
-
-            Assert.AreEqual(1, template.Errors.Count);
-            Assert.IsInstanceOf<SyntaxException>(template.Errors[0]);
         }
 
         [Test]
@@ -43,9 +40,6 @@ namespace DotLiquid.Tests
             Assert.DoesNotThrow(() => { template = Template.Parse(" {{ errors.argument_exception }} "); });
             string result = template.Render(Hash.FromAnonymousObject(new { errors = new ExceptionDrop() }));
             Assert.AreEqual(" Liquid error: argument exception ", result);
-
-            Assert.AreEqual(1, template.Errors.Count);
-            Assert.IsInstanceOf<ArgumentException>(template.Errors[0]);
         }
 
         [Test]
@@ -60,9 +54,6 @@ namespace DotLiquid.Tests
             Template template = null;
             Assert.DoesNotThrow(() => { template = Template.Parse(" {% if 1 =! 2 %}ok{% endif %} "); });
             Assert.AreEqual(" Liquid error: Unknown operator =! ", template.Render());
-
-            Assert.AreEqual(1, template.Errors.Count);
-            Assert.IsInstanceOf<ArgumentException>(template.Errors[0]);
         }
 
         [Test]
--- a/src/DotLiquid.Tests/ParallelTest.cs
+++ b/src/DotLiquid.Tests/ParallelTest.cs
@@ -1,0 +1,23 @@
+﻿using NUnit.Framework;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace DotLiquid.Tests
+{
+    [TestFixture]
+    public class ParallelTest
+    {
+        [Test]
+        public void TestCachedTemplateRender()
+        {
+            Template template = Template.Parse(@"{% assign foo = 'from instance assigns' %}{{foo}}");
+
+            var parallelOptions = new ParallelOptions { MaxDegreeOfParallelism = 30 };
+
+            Parallel.For(0, 10000, parallelOptions, (x) => Assert.AreEqual("from instance assigns", template.Render()));
+        }
+    }
+}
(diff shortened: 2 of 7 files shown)


bradphelan:
Can you please explain what was not thread safe about the previous implementation and how the changes improve the thread safety.

303248153:
The previous implementation allow a template object have it's own registers, assigns, instanceAssigns, errors, so use same template object across thread is not safe.
There servals ways to fix it
- Create a new template object every time, it'll slow down the performance very much
- Make access to template owned scope thread safe, it'll slow down the performance a little bit
- Remove template static variables, it'll cause feature lost
I taked the 3rd solution, but you can try to find a better solution

bradphelan:
What feature is lost?

303248153:
Please read #154 and #206 to understand the problem and review the code difference first.
And you're right, this PR shouldn't be merged until you decide what's better.

bradphelan:
I'm assuming that you mean feature that is lost is that caching is now per thread rather than across threads? I've just started helping with the maintenance of this project so I don't know all corners of the code. Just the bits that I've needed to fix.  I'll need to set aside some time to look at the code changes in detail>

bradphelan:
I still don't understand what "feature" you claim will be lost. I've read the two issues and briefly looked at the code. I'm new maintaining here as the original devs have lost interest. If you could write me a paragraph about how a user might discover this "lost feature" I can understand it better. 

Will the user discover the lost feature by

* compilation error
* runtime error
* no error but incorrect data

I agree that making the code thread safe is a good idea and I want to merge but I want to be sure I don't silently break an end users application.

303248153:
Hi,
Before this modification, Template instance can have it's own variables and etc, they just like "static variables" can shared across rendering, and they aren't thread safe if both thread render with same Template instance in the same time

Pick one test case I modified
``` csharp
         [Test]
-        public void TestInstanceAssignsPersistOnSameTemplateParsingBetweenRenders()
+        public void TestInstanceAssignsDoNotPersistOnSameTemplateParsingBetweenRenders()
         {
             // create a template instance
             Template t = Template.Parse("{{ foo }}{% assign foo = 'foo' %}{{ foo }}");
             // first rendering
             Assert.AreEqual("foo", t.Render());
             // second rendering
-            Assert.AreEqual("foofoo", t.Render());
+            Assert.AreEqual("foo", t.Render());
         }
```
You can see before this PR, foo will remain after first rendering, so in the second rendering the first {{ foo }} will give 'foo', but after this PR, the first {{ foo }} will always give null

bradphelan:
Ok now I fully get it.

To me I agree in principle with the idea that it persisting assigns persist between renders is __wierd__ and compromises thread safety.  However somebody thought it was a good idea and those tests were baked in from the very beginning and seem to be intended behavior. Changing that behavior could break a running application in non obvious ways. ie: no compilation errors, no runtime exceptions, just different/incorrect template rendering. They may pick this up in some test case if they robustly write test cases against rendered templates but who does that ??

I have a few ideas. Please comment.

Enforce Single Thread Usage As Default
==========================
The old behavior assumes single threaded usage. Let us enforce that by storing the current thread on first usage/parse and then verifying that we are on the same thread on each render. If on a different thread then throw an exception. This should only effect users who have incorrectly used the library in the first place.

Create A New Template Type That Is Thread Safe
===============================
Implement your new behavior as a new template type

    Template template = Template.Parse("blah blah blah") 
    template = template.MakeThreadSafe()

this now changes to the behaviour that you implements.

I agree this sounds like a design kludge and if starting from scratch would be a dumb idea. I just don't want to break existing applications, especially in a way that is difficult for them to verify.

Another Alternative
============

        [Test]

        public void TestInstanceAssignsPersistOnSameTemplateParsingBetweenRenders()

        {

            Template t = Template.Parse("{{ foo }}{% assign foo = 'foo' %}{{ foo }}");

            Assert.AreEqual("foo", t.Render());

            Assert.AreEqual("foofoo", t.Render());

        }

I don't like that the template allows you to use a variable before it is assigned. IMHO this test should fail because __foo__ is used before being assigned. If we change the library so that rendering of unassigned variables throws an exception rather than rendering whitespace __and__ implement your fix then users relying on the previous behavior will at least get a runtime exception explaining that there is an error.

laedit:
Thanks a lot both of you for your work!

IMHO the last alternative is not a valid option because the use of unassigned variable is permitted by liquid, and it could break a lot of applications which use dotliquid based on liquid templates.

The only case which should raise an exception is precisely when you parse several times a template which use an unassigned variable which have been assigned during the first parse, which is slightly different.

Since this case is very particular, I think that a visible statement in the release notes (maybe show a note on the update of the nuget package?) would be sufficient.

bradphelan:
@laedit That is a good idea. Just to clarify what you mean

    [Test]

    public void TestInstanceAssignsPersistOnSameTemplateParsingBetweenRenders()

    {

        Template t = Template.Parse("{{ foo }}{% assign foo = 'foo' %}{{ foo }}");

        Assert.AreEqual("foo", t.Render());

        // @laedit Should fail here as 'foo' is unassigned but a variable persists
        Assert.AreEqual("foofoo", t.Render());

    }

laedit:
@bradphelan yes, thanks for the clarification :smile:

303248153:
@bradphelan
I think Your solution "Create A New Template Type That Is Thread Safe" is good and practicable, I will create a new branch use your solution later and see if that work

303248153:
Since another PR is merged I'm going to close this PR

anato-s:
@bradphelan 
I would like to hear from you on this [https://github.com/dotliquid/dotliquid/issues/258#issuecomment-542934693](url)
Thanks

PR #211, opened 2016-10-20
add generic interface support for GetSafeTypeTransformer and GetValueTypeTransformer

303248153:
This commit add generic interface support for GetSafeTypeTransformer and GetValueTypeTransformer.
It make code like this work:
Template.RegisterSafeType(typeof(ITreeNode<>), new[] { "Value", "Parent", "Childs" });

Also improve some performance.

### DIFF
--- a/src/DotLiquid.Tests/TemplateTests.cs
+++ b/src/DotLiquid.Tests/TemplateTests.cs
@@ -1,3 +1,4 @@
+using System;
 using System.IO;
 using System.Net;
 using System.Web;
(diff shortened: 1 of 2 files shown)


bradphelan:
Could you please add some specs to demonstrate the new functionality?

303248153:
Assume I have an interface `ITreeNode` defined like
```
public interface ITreeNode<T> {
    T Value { get; set; }
    ITreeNode<T> Parent { get; set; }
    IEnumerable<ITreeNode<T>> Childs { get; set; }
}
```
And I have these implemention type
```
public IntTreeNode : ITreeNode<int> { }
public StringTreeNode : ITreeNode<string> { }
```

Without this PR I have to register every implemention type like
```
Template.RegisterSafeType(typeof(IntTreeNode ), new[] { "Value", "Parent", "Childs" });
Template.RegisterSafeType(typeof(StringTreeNode ), new[] { "Value", "Parent", "Childs" });
```

But with this PR I can just register the generic interface type
```
Template.RegisterSafeType(typeof(ITreeNode<>), new[] { "Value", "Parent", "Childs" });
```

bradphelan:
Ok. So this is an extension and doesn't change any existing functionality. However can you please extend the specs to show that this works. If you have a working spec then I will merge it.

303248153:
Well, I will add it later :)

PR #216, opened 2016-12-04
Add now/today inputs to date filter

cyotek:
Hello,

The documentation for the [date](https://shopify.github.io/liquid/filters/date/) filter states that you can use the `now` and `today` words to substitute for the current date. As Dotliquid doesn't support this, this PR adds support for that. Although the docs don't say either way, [499 for liquid](https://github.com/Shopify/liquid/pull/499) does fix a regression where case insensitivity was broken, so I have made this implementation case insensitive as well.

I also adjusted `StandardFilterTests.TestDate` to include basic checking of the keywords.

I haven't actually used .NET Core yet, and although I was under the expression my 2015 dev environment was up to date (except for NCrunch), neither NCrunch, nor Resharper, not VSTest would actually find or run any tests (except for the DotLiquid.Website.Tests which wasn't very helpfull). So after making my changes, I would compile then run `dotnet-test-nunit.exe DotLiquid.Tests.dll` from the `Debug\net451\win7-x64` folder.

This indicated my adjusted test passed, but always claimed that `DotLiquid.Tests.Util.StrFTimeTests.TestFormat` failed - even if I reverted my original changes. So they don't appear to be related but I thought I'd mention it here regardless.

````
1) Failed : DotLiquid.Tests.Util.StrFTimeTests.TestFormat("%W")
  String lengths are both 2. Strings differ at index 1.
  Expected: "01"
  But was:  "02"
````

As I said, I haven't worked with Core before so it's very possible I may well have missed something - please let me know if this PR is lacking.

Thanks;
Richard Moss

### DIFF
--- a/src/DotLiquid.Tests/StandardFilterTests.cs
+++ b/src/DotLiquid.Tests/StandardFilterTests.cs
@@ -127,6 +127,11 @@ namespace DotLiquid.Tests
 
             Assert.AreEqual("hi", StandardFilters.Date("hi", "MMMM"));
 
+            Assert.AreEqual(DateTime.Now.ToString("MM/dd/yyyy"), StandardFilters.Date("now", "MM/dd/yyyy"));
+            Assert.AreEqual(DateTime.Now.ToString("MM/dd/yyyy"), StandardFilters.Date("today", "MM/dd/yyyy"));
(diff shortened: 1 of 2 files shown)


bradphelan:
Thanks for the pull request. I reviewed it and it looks good. With regards to the failing test at "%W" I think this has something to do with the difference in behaviour between .Net week calculation and the spec the function tries to follow. It doesn't seem to stop the CI build running so i've ignored it for the moment.

(i've just picked up a maintaining role here so I'm not super familiar with the base)

PR #220, opened 2017-01-12
make template instance thread safe (provide extra option)

303248153:
Hello, this PR provide an option that can make Template instance thread safe.
After calling Template.MakeThreadSafe, user can't assign values to template instance anymore.

Example
``` csharp
Template t = Template.Parse("{{ foo }}{% assign foo = 'foo' %}{{ foo }}");
 t.MakeThreadSafe();
Assert.AreEqual("foo", t.Render());
Assert.AreEqual("foo", t.Render());
```

Because it's an extra option, this modification won't affect existing code.
Also I added a global option Template.DefaultIsThreadSafe can make all Template instances thread safe by default.

### DIFF
--- a/src/DotLiquid.Tests/ParallelTest.cs
+++ b/src/DotLiquid.Tests/ParallelTest.cs
@@ -1,0 +1,21 @@
+﻿using NUnit.Framework;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace DotLiquid.Tests {
+    [TestFixture]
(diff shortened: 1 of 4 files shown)


sebastienros:
Even though I would prefer templates to be thread safe and idempotent by default, please accept this PR.

tgjones:
Thank you @303248153. I regret not making `Template` thread-safe to begin with, but this is a good non-API-breaking workaround.

yufeih:
Thank you @303248153. When can we expect a NuGet package that contains this change?

PR #222, opened 2017-01-19
Fix broken CSharpNamingConvention conditions

swiftMessenger:
This PR is to fix a bug with the hasKey and hasValue conditions when used with the CSharpNamingConvention.

To reproduce the bug, simply create any template that checks if a key exists in a dictionary using the syntax {% if my_dictionary hasKey 'keyname' %} and then process that template using the CSharpNamingConvention.
The exact exception will be [DotLiquid.Exceptions.ArgumentException: Unknown operator has]

The nature of the problem comes down to the Regex used to process the {% if %} tag. Specifically - this regex looks for LOWERCASE values, but the hasKey and hasValue conditionals each have a single UPPERCASE letter, which breaks the pattern matching and causes it to look for a conditional operator which matches everything the regex captured - the word 'has'.

There are several potential fixes - the simplest and most future-proof would probably be to rename the conditional to something with all lowercase characters like the rest of the conditionals (eg: haskey and hasvalue, withkey and withvalue, or even has_key and has_value) but that would cause too much of a backwards compatibility issue. To get around the backwards-compatibility, the renamed conditionals could simply be added as aliases of the originals and no existing code would be broken. The main issues there would be agreeing on what to name them and living with the minor deviation from the original Ruby Liquid library.

A second simple fix, and the one I have implemented in this PR, is to simply allow upper-case K and V characters in conditionals. With this change, the hasKey and hasValue conditions will be properly parsed and processed.


bradphelan:
This has been in the product for 7 years. Why is it a problem now? I thought the convention was to force everything to lower case.

https://github.com/dotliquid/dotliquid/blob/130b3fd8c95a135fa595c827d2678802a3cf2cdf/src/DotLiquid.Tests/NamingConventionTests.cs

Perhaps the problem is that when a user enters
```
{% if my_dictionary hasKey 'keyname' %} 
```
it should be forced to
```
{% if my_dictionary haskey 'keyname' %} 
```

and then you can change `hasKey` to `haskey` here https://github.com/dotliquid/dotliquid/blob/130b3fd8c95a135fa595c827d2678802a3cf2cdf/src/DotLiquid/Condition.cs#L34

swiftMessenger:
There are two main reasons this was not caught as a problem before.

The first reason is that the "hasKey" and "hasValue" conditionals are not tested (really - https://github.com/dotliquid/dotliquid/blob/130b3fd8c95a135fa595c827d2678802a3cf2cdf/src/DotLiquid.Tests/ConditionTests.cs).

The second reason is that the CSharpNamingConvention is barely tested (only 1 test in dotliquid/src/DotLiquid.Tests/NamingConventionTests.cs and 2 in dotliquid/src/DotLiquid.Tests/StudlyCapsTests.cs).

Perhaps this indicates that people are not using the CSharpNamingConvention, or that they are not utilizing dictionaries in their locals, or that they have found other ways around the issue and moved on without contributing.

Whatever the cause, forcing the condition to be lowercase would fix it as long as the regex was also updated to allow uppercase letters (which would be similar to my current fix, except going several steps further). Hypothetically this would be backwards compatible with all valid uses of DotLiquid, but more test cases are needed to guarantee present and future compatibility. Also, forcing conditions to be lowercase could become a roadblock to the potential future ability to add custom conditional statements (eg: {% if my_var is_prime_number %} or {% if User.LastVisit IsDateBefore '1/1/2017' %},  which are not currently possible but would be fairly simple to make DotLiquid extendable in this manner).

laedit:
Sorry for the delay, we are few and don't have many time to work on this project.

As stated in the [Drop Wiki](https://github.com/dotliquid/dotliquid/wiki/DotLiquid-Drops#tips), the `CSharpNamingConvention` have only an effect on `Drop` properties, the conditions aren't affected by it.

After some tests on [`hasKey`](https://github.com/dotliquid/dotliquid/blob/130b3fd8c95a135fa595c827d2678802a3cf2cdf/src/DotLiquid/Condition.cs#L34) and [`startswith`](https://github.com/dotliquid/dotliquid/blob/130b3fd8c95a135fa595c827d2678802a3cf2cdf/src/DotLiquid/Condition.cs#L32) it seems that both are available as `haskey` and `HasKey` for the first and `startswith` and `StartsWith` for the second in both Ruby and CSharp naming conventions. 

I think that the all lowercase was intended for the Ruby and the capitalized for CSharp, but we have the flexibility to use both in any case.

So I don't think that `CSharpNamingConvention` is broken, but it is clear that we lack a lot of test for it and maybe a better error message for unknown operator which could suggest the closest or list all operators supported.

laedit:
**Correction:**
The `HasKey` works only by side-effect and if you expect a positive result.

So based on the facts that the other operators specific to DotLiquid (`startswith` and `endswith`) doesn't have capitalized letter, that the wiki explicitly state that the naming convention has an effect on variables only and that `hasKey` and `hasValue` can't be used (by the way if anyone know how to make `hasvalue` works, I haven't find it yet), I think that they are intended to be used in lowercase.

laedit:
**To summarize:**
Right now it seems that only lowercase works for Ruby and CSharp, but CSharp should have aliases in lower camel case.
So I think that the if syntax should allow `a-zA-Z` and not just `a-zKV` and the naming conventions should have another comparator for condition in order to allow only lowercase for ruby and both lowercase and lower camel case.

swiftMessenger:
Thanks for the replies @laedit. I updated the PR with these corrections. Specifically, I:

1. allowed all uppercase letters in if tag operators
2. added a new INamingConvention property for a ConditionalComparer - set to OrdinalIgnoreCase for both Ruby and CSharp naming conventions
3. utilized the new ConditionalComparer property for the Operators dictionary
4. Created unit tests for the hasKey and hasValue conditions
5. Fixed the hasValue operator - which appears to have never worked in the first place.

src/DotLiquid/Tags/If.cs
```diff
@@ -20,7 +20,8 @@ namespace DotLiquid.Tags
     public class If : DotLiquid.Block
     {
         private string SyntaxHelp = Liquid.ResourceManager.GetString("IfTagSyntaxException");
-        private static readonly Regex Syntax = R.B(R.Q(@"({0})\s*([=!<>a-zKV_]+)?\s*({0})?"), Liquid.QuotedFragment);
+        private string TooMuchConditionsHelp = Liquid.ResourceManager.GetString("IfTagTooMuchConditionsException");
```
laedit: It seems that it is not used, could you remove it?

src/DotLiquid/Condition.cs
```diff
@@ -19,7 +19,7 @@ public class Condition
     {
         #region Condition operator delegates
 
-        public static readonly Dictionary<string, ConditionOperatorDelegate> Operators = new Dictionary<string, ConditionOperatorDelegate>(Template.NamingConvention.StringComparer)
+        public static readonly Dictionary<string, ConditionOperatorDelegate> Operators = new Dictionary<string, ConditionOperatorDelegate>(Template.NamingConvention.ConditionComparer)
```
laedit: Since the comparer is used in the creation of a static field, it will not change depending on the selected naming convention.
I think it could be replaced by a custom `IEqualityComparer` based on the current Template.NamingConvention.

src/DotLiquid/Condition.cs
```diff
@@ -32,7 +32,7 @@ public class Condition
             { "startswith", (left, right) => (left is IList) ? EqualVariables(((IList) left).OfType<object>().FirstOrDefault(), right) : ((left is string) ? ((string)left).StartsWith((string) right) : false) },
             { "endswith", (left, right) => (left is IList) ? EqualVariables(((IList) left).OfType<object>().LastOrDefault(), right) : ((left is string) ? ((string)left).EndsWith((string) right) : false) },
             { "hasKey", (left, right) => (left is IDictionary) ? ((IDictionary) left).Contains(right) : false },
-            { "hasValue", (left, right) => (left is IDictionary) ? ((IDictionary) left).OfType<object>().Contains(right) : false }
+            { "hasValue", (left, right) => (left is IDictionary) ? ((IDictionary) left).Values.Cast<object>().Contains(right) : false }
```
laedit: I have not tested, but is the `Cast<object>` really necessary?

src/DotLiquid/NamingConventions/RubyNamingConvention.cs
```diff
@@ -25,6 +25,11 @@ public StringComparer StringComparer
             get { return StringComparer.OrdinalIgnoreCase; }
         }
 
+        public StringComparer ConditionComparer
+        {
+            get { return StringComparer.OrdinalIgnoreCase; }
```
laedit: I would prefer to keep only lowercase for ruby.

src/DotLiquid/NamingConventions/CSharpNamingConvention.cs
```diff
@@ -9,6 +9,11 @@ public StringComparer StringComparer
             get { return StringComparer.Ordinal; }
         }
 
+        public StringComparer ConditionComparer
+        {
+            get { return StringComparer.OrdinalIgnoreCase; }
```
laedit: I would prefer to only authorize lowercase and lower camel case for CSharp.

laedit:
Thanks a lot for your quick reaction! 👍 
There is still some things bothering me, could you fix the comments?

src/DotLiquid/Tags/If.cs
swiftMessenger: The TooMuchConditionsHelp string is from the master. I could remove it if desired, but that would conflict with the master repo: https://github.com/dotliquid/dotliquid/blob/master/src/DotLiquid/Tags/If.cs#L23

src/DotLiquid/Condition.cs
swiftMessenger: The Cast<object>() converts it from a ICollection to an IEnumerable. Unfortunately it seems like it is a necessary step, though I would be happy to be proven wrong.

src/DotLiquid/NamingConventions/RubyNamingConvention.cs
swiftMessenger: I'll switch the Ruby ConditionalComparer to Ordinal then - conditions will have to use the correct capitalization.

swiftMessenger: Or maybe you are saying that you want to throw an error if the condition contains any non-lowercase characters? That would be in line with your comment on the CSharp ConditionComparer as well. I'd have to assume that non-alphanumeric characters would be allowed (at least the same characters that the regex allows: [=!<>a-zA-Z_]). I wouldn't want to break backwards compatibility with anyone who has implemented custom operators: https://github.com/dotliquid/dotliquid/blob/master/src/DotLiquid.Tests/ConditionTests.cs#L171

src/DotLiquid/Tags/If.cs
laedit: I didn't realized, sorry, but the diff shouldn't have showed it.

src/DotLiquid/Condition.cs
laedit: Ok, so let's keep it for now.

swiftMessenger:
I updated the Operators dictionary to use a custom StringComparer which acts as an intermediary to the Template.NamingConvention.StringComparer. The result is that when using the RubyNamingConvention all comparisons will be case insensitive and when using the CSharpNamingConvention they will be case sensitive.

I also added a couple more unit tests to verify the expected behavior with capitalization.

The only piece that I am aware of that this does not address is forcing operators to be lowercase with Ruby. I suppose that could be fairly easily resolved by always using StringComparer.Ordinal; because all of the operators are lowercase, an operator with an uppercase character would throw an ArgumentException. What do you think?

laedit:
Thanks again for your commit 😄 

The way I see this:
With the current version of DotLiquid, only lowercase operators works properly.
There is an issue with the tests, the `AssertEvaluatesTrue` and `AssertEvaluatesFalse` only tests the direct evaluation of the operator without the template parsing phase, which otherwise would have shown that even custom capitalized operators doesn't works properly unless used as lowercase (it fails when you check if the operator is evaluated to false) .

So we should keep the lowercase for compatibility, but for ruby also support snake case (eg. `startswith` and `starts_with`) and for csharp lower camel case (eg. `startswith` and `startsWith`).
I am also wondering if the support of upper camel case for csharp (`StartsWith`) is a good thing or not.

Any other casing should raise an exception.

So it will need customs StringComparer used only for condition operators, used in the `ConditionOperatorComparator` you have added.
And since it is kind of a breaking change I think we should add a [file opened automatically on nuget install](https://docs.microsoft.com/en-us/nuget/create-packages/creating-a-package#adding-a-readme-and-other-files) which explains the issue and possible risks.

What do you think?

swiftMessenger:
I agree that it is an issue that the unit tests are currently unable to test conditionals (and other tags, etc..) in the context of different template naming conventions. They currently CAN test code in the default RubyNamingConvention by using Helper.AssertTemplateResult, which I can update the condition tests to utilize.

In the interest of testability, I would recommend updating Template.NamingConvention to Template.DefaultNamingConvention and making the naming convention configurable per template (which would have to be done via the Context). It may be possible to do this without causing a breaking change (by aliasing Template.DefaultNamingConvention as Template.NamingConvention with a deprecated warning and adding an additional optional arg to the Context constructor) but we would have to be very careful doing it and I think that should be done as part of a separate PR.

swiftMessenger:
Does ruby currently support snake case operators (eg: starts_with)? Looking through the code, it seems like that would not currently work, in which case adding that functionality would potentially be a breaking change if anyone has implemented custom operators using the RubyNamingConvention that use the _ character.

Regarding requiring all lowercase operators with the RubyNamingConvention, I see your point that the current version does not allow uppercase operators, and although I do not see the harm in allowing uppercase characters in the operators it should be fairly easy to prevent them.

Regarding operators with the CSharpNamingConvention, I do not see why we wouldn't allow whatever capitalization people wanted to implement - particularly since the operator lookup is case-sensitive and will throw an error if the operator is not found using the correct capitalization. It would be up to whoever extends DotLiquid to choose the casing that they prefer when adding custom operators.

swiftMessenger:
Upon further reflection and inspection, maybe we should just make all operator comparisons case-sensitive. If someone extends DotLiquid using the RubyNamingConvention and capital conditional operators then that is their decision and doesn't hurt anything. Additionally, if they want to make their operators usable using multiple capitalization options, they could just add an extra item to the dictionary as an alias (ie: Condition.Operators['starts_with'] = Condition.Operators['startswith']).

By default all of the operators are lowercase, so allowing uppercase conditional operators is really something that a developer would have to explicitly allow using a case-sensitive StringComparer.

In the meantime, I added the readme message that you recommended and updated some condition tests to include template parsing.

With the latest commit, the only potential breaking change would be if someone added a custom conditional operator with a capital letter (eg: "divisibleBy") since they would previously have had to access it as eg: "{% if 4 divisibleby 2 %}" and would now have to access it as eg: "{% if 4 divisibleBy 2 %}".

laedit:
I agree with you regarding the tests, right now I use a modification of `Helper.AssertTemplateResult` to check quickly the legacy behavior of operators:

    public static void AssertTemplateResult(string expected, string template, Hash localVariables, INamingConvention namingConvention = null)
    {
        var currentNamingConvention = Template.NamingConvention;
        if (namingConvention == null)
        {
            Template.NamingConvention = new RubyNamingConvention();
        }
        else
        {
            Template.NamingConvention = namingConvention;
        }

        try
        {
            Assert.AreEqual(expected, Template.Parse(template).Render(localVariables));
        }
        catch
        {
            throw;
        }
        finally
        {
            Template.NamingConvention = currentNamingConvention;
        }
    }

And of course I am totally ok with your recommendation of making naming convention configurable by template. If you haven't the time to propose a PR, can you just create an issue in order that we don't forget it? Thanks a lot.

In my previous comment I haven't talk about custom operators but of course they should be used with the exact capitalization which they have been added.

Regarding the snake case, ruby doesn't support as much operators that we do, be since they use [snake case](https://github.com/Shopify/liquid/wiki/Liquid-for-Designers#standard-filters) I think that it is right to support it too.

So in order to minimize the breaking change, I think that lower case must be supported for all naming convention, ruby may support snake case, csharp must support ordinal and custom operators should be available in lower case and in their exact capitalization.
With that, only people actually using capitalized operators will have a breaking change (by either having an operator working without bug or having an exception for unknown operator.

swiftMessenger:
I do not have time right now to propose a PR for the naming convention change, so I created an issue for it: https://github.com/dotliquid/dotliquid/issues/247

Regarding parsing conditions, please let me know if there is anything else you need me to do. Ideally, if it does not currently work the way you expect it to, can you add a unit test to the patch-1 branch (I added you as a collaborator) that tests the desired behavior so that I know what needs to happen in order to satisfy the PR requirements and be merged into the master repository?

Thanks.

laedit:
Sorry if I wasn't clear enough.
I will add some tests.

laedit:
Thanks again for your work!
I must say that I am not totally happy with the use of aliases because it might be forgotten when adding new operators, but that'll do for now.
Can you just remove the useless `using System;` on all `*NamingConvention` classes and change the date in the readme to today and I'll merge it.

PR #224, opened 2017-02-14
Notify unknown variables in the Error collection

albertospelta:
Request to re-open PR #156. When a variable used in the template cannot be resolved, an new VariableNotFoundException is added to the Error list so that it is possible to show to the end user which variables are evaluated to null because they do not exists.

### DIFF
--- a/src/DotLiquid/Properties/Resources.resx
+++ b/src/DotLiquid/Properties/Resources.resx
@@ -219,6 +219,9 @@
   <data name="VariableFilterNotFoundException" xml:space="preserve">
     <value>Error - Filter '{0}' in '{1}' could not be found.</value>
   </data>
+  <data name="VariableNotFoundException" xml:space="preserve">
+    <value>Error - Variabile '{0}' could not be found</value>
+  </data>
   <data name="WeakTableKeyNotFoundException" xml:space="preserve">
     <value>key could not be found</value>
   </data>
--- a/src/DotLiquid/Properties/Resources.resx
+++ b/src/DotLiquid/Properties/Resources.resx
@@ -220,7 +220,7 @@
     <value>Error - Filter '{0}' in '{1}' could not be found.</value>
   </data>
   <data name="VariableNotFoundException" xml:space="preserve">
-    <value>Error - Variabile '{0}' could not be found</value>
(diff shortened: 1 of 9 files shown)


pmccloghrylaing:
This would be great to have added in. In if/unless conditions that are just truthy or falsy values I don't think an error should be logged - I added the small change and some tests in pmccloghrylaing/dotliquid@2db55207787faef46ecbc6af11955aabc62f61b4. I've had issues building it so I'm assuming it passes.

src/DotLiquid/Properties/Resources.resx
```diff
@@ -219,6 +219,9 @@
   <data name="VariableFilterNotFoundException" xml:space="preserve">
     <value>Error - Filter '{0}' in '{1}' could not be found.</value>
   </data>
+  <data name="VariableNotFoundException" xml:space="preserve">
+    <value>Error - Variabile '{0}' could not be found</value>
```
laedit: Little typo, could you fix it?

laedit:
Thanks a lot for this PR!
There is only one little type to fix in the resources before merging.

Regarding @pmccloghrylaing proposition, I think that it is wanted, could you look at it and integrates it?

laedit:
Could you please rebase in order to build/test on AppVeyor?

src/DotLiquid/Properties/Resources.resx
albertospelta: Typo fixed in cde6b906d2188e9bbcca232aace759ad2547db6c

laedit:
It seems that the PR have been "polluted" with other commits, is it possible to clean it? Or to create a new one with only the necessary commits?
And since Liquid  consider [non-existent value as `false`](https://github.com/Shopify/liquid/wiki/Liquid-for-Designers#if--else) can you avoid to add an error in this case as @pmccloghrylaing asked?
Thanks.

albertospelta:
I ran into a few problems with the rebase, now the PR seems to be clean.
I've included @pmccloghrylaing proposition with e5e6cb5.
Thanks!

laedit:
Thanks a lot, you are awesome!
I will review it this weekend.

PR #225, opened 2017-02-14
Fix issue preventing filters with multiple method signatures and injected context argument from being invoked

matthewrk:
When looking for a filter the Strainer.Invoke method was counting all filter params on the Liquid side against all params on the C# side. This was not accounting for the auto injected and optional Context object, which caused the correct filter to be missed.

The reason this wasn't a bigger issue is because the strainer then goes on to find the closest match starting with the one with the most parameters, it then set the context object and all was well.

The issue arrises when you have multiple method signatures on that same method, where it attempts to invoke the wrong method (as the included tests show). So this code fixes that, and should result in the Invoke method finding the correct filter straight off more often rather than working on a fallback scenario.

### DIFF
--- a/src/DotLiquid.Tests/FilterTests.cs
+++ b/src/DotLiquid.Tests/FilterTests.cs
@@ -61,6 +61,32 @@ namespace DotLiquid.Tests
             }
         }
 
+        private static class FiltersWithMulitpleMethodSignatures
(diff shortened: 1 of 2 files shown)


tgjones:
Looks good, with tests too. Thanks!

PR #233, opened 2017-03-22
Remove unnecessary cast to string on IIndexable check

jhominal:
I would like to remove the cast in the call to `IIndexable.ContainsKey`, as I believe that it brings no value, and inhibits at least the following cases:

 * A list-like object that does not desire to implement the whole `IList` interface
 * An object that uses `IIndexable` to support both integer and string indices

### DIFF
--- a/src/DotLiquid.Tests/CustomIndexableTests.cs
+++ b/src/DotLiquid.Tests/CustomIndexableTests.cs
@@ -1,0 +1,100 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using NUnit.Framework;
+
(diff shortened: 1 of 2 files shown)


laedit:
Sorry for the delay, we are few and don't have many time.
I will close and reopen the PR only to relaunch the AppVeyor check.

laedit:
Sounds good to me, can you just add some tests for the two case you mentioning?

jhominal:
@laedit I have added a few tests to this PR. What do you think?

laedit:
Thanks! 😃

Issue #228, opened 2017-03-05
why is Flatten() used in Sort()?


rolandbaer:
In the StandardFilters.cs class the Sort() method calls Flatten() if the given input is IEnumerable.

Why is this? For me it causes some trouble because when you give a List of Hash elements as input (e.g 3 Hash elements with 8 entries each) the list is flattened and the content of the hash elements (e.g. 24 elements) are then tried to be sorted.

Occured for me as i use pretzel and wanted to do a navigation like in jekyll (see http://jekyll.tips/jekyll-casts/navigation/ )

microalps:
This is the behavior of the native ruby implementation. See https://replit.com/@microalps/LiquidWhereTests

PR #234, opened 2017-04-22
Don't flatten list on sort if input is a list of Hash elements and …

rolandbaer:
…property is set.

I had an issue with creating a menu in pretzel as it was discribed for jekyll. The problem was the Flatten() in Sort(), so here it is disabled for this special case. (see also #228).


laedit:
Sorry for the delay, we are only a few maintainers without an extended knowledge of the project.

While adding some tests to `StandardFilters`, I realized that the `hash` case isn't the only issue with the use of `Flatten`: when using a dictionary, the specific handler can't be hit because of the flatten.

@tgjones do you remember why the `Flatten` is used for all `IEnumerable` on `Sort`?

haf:
+1 to merge PR!

laedit:
Since we haven't got a response from @tgjones  I think we can merge.

PR #235, opened 2017-04-23
Adding global.json for AppVeyor

rolandbaer:
Fixes AppVeyor build problems as described in http://help.appveyor.com/discussions/problems/6324-aspnet-core-builds-now-failing


laedit:
Thanks a lot for this!
Is the two files necessaries?
The thread linked indicate that the `global.json` need only to be in the folder where the `dotnet` command is run and in this case it's the root.

rolandbaer:
I'm not sure. In the response 3 is written "should be there as well", therefore I added them to both places. I can try if it will work with only the one in the root and then update this PR.

laedit:
Yes please 😄

laedit:
Wonderful, thanks a lot!

rolandbaer:
Wow, fast!
You're welcome. Was a pleasure to help.

PR #236, opened 2017-04-30
Fix security issues on infinite regex and too much if condition

laedit:
1) Regex denial of service via Replace filter
A 10s default time out is now in place and the user can change it at his convenience.

2) Unbounded recursion leads to uncatchable exception
A limit max of 500 conditions has been added to the if tag.

Sorry for the noise, it seems that the editorconfig have kick in.


laedit:
I don't know why the 'default' build on AppHarbor have failed and I don't have access to the detail, does anyone have more info?

tgjones:
The AppHarbor build is, AFAIK, only for updating the dotliquidmarkup.org website. It shouldn't affect the AppVeyor build.

Looks good to me - thanks for this.

PR #238, opened 2017-05-04
Migration to the new csproj format

laedit:
Compatible with Visual Studio 2017 and the latest .net core SDK.
AppVeyor and TravisCI builds seems fine.

Side note: I have moved the 'master only' of AppVeyor to the deploy section in order to allow the build of any branch, but I think that it still too wide: this PR will push a new version to Nuget, which is not wanted.
So if that is ok I propose to push new version on tag only.


laedit:
@dotliquid/core-team anyone have an opinion on this?

haf:
No objections from me.

tgjones:
@laedit it looks okay to me.

I quite like that all pushes to master also result in a push to NuGet. If this was a busier project, that  probably would be too much, but as it is, it's good that there's no explicit release step.

laedit:
@tgjones dully noted :)

PR #239, opened 2017-05-15
Add ITemplateFileSystem that allow return Template instance for better performance 

303248153:
Refer to #209 
This implementation won't break exists code, it allow user to switch a new interface if they intend to return a cached template instance.
The definition of ITemplateFileSystem is 

``` c#
public interface ITemplateFileSystem : IFileSystem
{
    Template ReadTemplateInstance(Context context, string templateName);
}
```

Then I modified Include and Extend tag, If file system object is also ITemplateFileSystem, it will call ITemplateFileSystem.ReadTemplateInstance before IFileSystem.ReadTemplateFile.

### DIFF
--- a/src/DotLiquid/FileSystems/ITemplateFileSystem.cs
+++ b/src/DotLiquid/FileSystems/ITemplateFileSystem.cs
@@ -1,0 +1,17 @@
+﻿namespace DotLiquid.FileSystems
+{
+    /// <summary>
+    /// This interface allow you return a Template instance,
+    /// it can reduce the template parsing time in some cases.
+    /// Please also provide the implementation of ReadTemplateFile for fallback purpose.
+    /// </summary>
+    public interface ITemplateFileSystem : IFileSystem
+    {
+        /// <summary>
+        /// Called by Liquid to retrieve a template instance
+        /// </summary>
+        /// <param name="templatePath"></param>
+        /// <returns></returns>
+        Template ReadTemplateInstance(Context context, string templateName);
+    }
+}
--- a/src/DotLiquid/FileSystems/ITemplateFileSystem.cs
+++ b/src/DotLiquid/FileSystems/ITemplateFileSystem.cs
@@ -12,6 +12,6 @@
         /// </summary>
         /// <param name="templatePath"></param>
         /// <returns></returns>
-        Template ReadTemplateInstance(Context context, string templateName);
+        Template GetTemplate(Context context, string templateName);
     }
 }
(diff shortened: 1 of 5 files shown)


src/DotLiquid/FileSystems/ITemplateFileSystem.cs
```diff
@@ -0,0 +1,17 @@
...
+    /// Please also provide the implementation of ReadTemplateFile for fallback purpose.
+    /// </summary>
+    public interface ITemplateFileSystem : IFileSystem
+    {
+        /// <summary>
+        /// Called by Liquid to retrieve a template instance
+        /// </summary>
+        /// <param name="templatePath"></param>
+        /// <returns></returns>
+        Template ReadTemplateInstance(Context context, string templateName);
```
laedit: Since the purpose is to get the template, not always reading it, I think that `GetTemplate` is more appropriate.

src/DotLiquid/Tags/Extends.cs
```diff
@@ -103,9 +103,16 @@ public override void Render(Context context, TextWriter result)
         {
             // Get the template or template content and then either copy it (since it will be modified) or parse it
             IFileSystem fileSystem = context.Registers["file_system"] as IFileSystem ?? Template.FileSystem;
-            object file = fileSystem.ReadTemplateFile(context, _templateName);
-            Template template = file as Template;
-            template = template ?? Template.Parse(file == null ? null : file.ToString());
+            Template template = null;
+            if (fileSystem is ITemplateFileSystem)
```
laedit: It is better to avoid the double cast with:

    var tfs = fileSystem as ITemplateFileSystem;
    if (tfs != null)
    {
        template = tfs.ReadTemplateInstance(context, _templateName);
    }

Or shorter with C#7:

    if (fileSystem is ITemplateFileSystem tfs)
    {
        template = tfs.ReadTemplateInstance(context, _templateName);
    }

src/DotLiquid/Tags/Include.cs
```diff
@@ -41,8 +41,16 @@ protected override void Parse(List<string> tokens)
         public override void Render(Context context, TextWriter result)
         {
             IFileSystem fileSystem = context.Registers["file_system"] as IFileSystem ?? Template.FileSystem;
-            string source = fileSystem.ReadTemplateFile(context, _templateName);
-            Template partial = Template.Parse(source);
+            Template partial = null;
+            if (fileSystem is ITemplateFileSystem)
```
laedit: Same comment that for `Extends`.

laedit:
That sounds good!

It only miss a test for `Extends` and some minor fixes commented.

303248153:
Done, please review again!

laedit:
Thanks, all good! :)

PR #244, opened 2017-06-02
+ docfx as an open source project using DotLiquid

yufeih:
Let the world know [docfx](https://github.com/dotnet/docfx) is also using DotLiquid


laedit:
Thanks! :)

PR #245, opened 2017-06-05
Add code coverage

laedit:
And add CodeCover as code coverage analyzer.
@tgjones If you are ok, can you install the [CodeCover integration](https://github.com/integration/codecov) to the repo?
It provide more interaction like PR comments.


laedit:
Looks like the integration is not necessary for comments after all.
The failed check is normal because there is no code coverage for master to compare with.

tgjones:
@laedit I have installed CodeCover integration to dotliquid/dotliquid.

laedit:
Thanks! :)

PR #248, opened 2017-06-14
Add currency filter and tests

kaeedo:
A simple filter to convert an input into a currency string using .Net .ToString() currency format and specified culture info

### DIFF
--- a/src/DotLiquid.Tests/StandardFilterTests.cs
+++ b/src/DotLiquid.Tests/StandardFilterTests.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Globalization;
+using System.Threading;
 using NUnit.Framework;
 
 namespace DotLiquid.Tests
@@ -122,6 +123,11 @@ namespace DotLiquid.Tests
         [TestCase("6.8458", "$6.85")]
         public void TestAmericanCurrencyFromString(string input, string expected)
         {
+#if CORE
+            CultureInfo.DefaultThreadCurrentCulture = new CultureInfo("en-US");
+#else
+            Thread.CurrentThread.CurrentCulture = new CultureInfo("en-US");
+#endif
             Assert.AreEqual(expected, StandardFilters.Currency(input));
         }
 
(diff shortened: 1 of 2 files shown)


src/DotLiquid/StandardFilters.cs
```diff
@@ -227,6 +227,26 @@ public static string StripHtml(string input)
...
+        /// <param name="input"></param>
+        /// <param name="cultureInfo"></param>
+        /// <returns></returns>
+        public static string Currency(object input, string cultureInfo = "en-US")
+        {
+            decimal amount;
+
+            if (decimal.TryParse(input.ToString(), out amount))
+            {
+                var culture = new CultureInfo(cultureInfo);
```
laedit: I am not a globalization expert, but I think in this case it is better to use `new CultureInfo(cultureInfo, false);` in order to avoid any side effect due to the culture of the machine, as stated in [MSDN](https://msdn.microsoft.com/en-us/library/ky2chs3h(v=vs.110).aspx#Anchor_2).

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -115,6 +115,35 @@ public void TestMap()
         }
 
         [Test]
+        public void TestCurrency()
```
laedit: Can you separate this in multiple tests, for example based on the culture?
It is best to have a limited number of asserts in one test.

src/DotLiquid/StandardFilters.cs
kaeedo: .net Standard 1.3 doesn't have that constructor. Therefore, I changed it to read from the `CurrentCulture` static property

src/DotLiquid.Tests/StandardFilterTests.cs
kaeedo: Broken into several tests

laedit: 👍

src/DotLiquid/StandardFilters.cs
laedit: It is effectively best to take the current culture as default, thanks.
It doesn't fix the user override issue but since the specific constructor is not part of .net standard 1.3, that should do it.

PR #253, opened 2017-08-03
Modified the Hash class with the virtualization of GetValue and Conta…

Sbaia:
I've modified the Hash class with GetValue virtualization and Contains methods to create a superclass with the override of the methods and try to access the properties of a particular object just when printing, and then accelerate the creation of the hash.

This is very useful when there are expensive properties, either Lazy or Ghost properties.

I have created a Test class (LazyTestHash) that has an usage example.

Thanks

SR


laedit:
Thanks for your PR, I will review it ASAP!

laedit:
Looks good to me, thanks again!

PR #254, opened 2017-08-03
max iterations error, error output modes

dr0wzee:
This code can be a problem:

```
{% for i in (1..100000) %}
  {{ i }}
{% endfor %}
```

It is also not always convenient to write exceptions to the output. I added an additional exception handling mode.

Thanks

### DIFF
--- a/src/DotLiquid.Tests/ExceptionHandlingTests.cs
+++ b/src/DotLiquid.Tests/ExceptionHandlingTests.cs
@@ -90,6 +90,19 @@ namespace DotLiquid.Tests
         }
 
         [Test]
+        public void TestTimeoutError()
+        {
+            var template = Template.Parse(" {% for i in (1..1000000) %} {{ i }} {% endfor %} ");
+            Assert.Throws<System.TimeoutException>(() =>
+            {
+                template.Render(new RenderParameters
+                {
+                    Timeout = 100 //ms
+                });
+            });
+        }
+
+        [Test]
         public void TestErrorsOutputModeRethrow()
         {
             var template = Template.Parse("{{test}}");
--- a/src/DotLiquid.Tests/ExceptionHandlingTests.cs
+++ b/src/DotLiquid.Tests/ExceptionHandlingTests.cs
@@ -1,4 +1,4 @@
-using DotLiquid.Exceptions;
+﻿using DotLiquid.Exceptions;
 using NUnit.Framework;
 
 namespace DotLiquid.Tests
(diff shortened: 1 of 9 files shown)


bradphelan:
Can you please add a little more detail on exactly what this pull request is fixing. I am guessing that it throws an exception if there are too many iterations or timeout. Could you elaborate on what the behavior before and after the change was. Does any documentation need updating? Will this cause existing deployments of dotliquid to throw exceptions where they didn't beforehand?

dr0wzee:
Hi,

I'm trying to fix 2 things:

1. In the current version, DotLiquid can either throw out exceptions or write them into the result of generation. I believe that another "quiet" mode can be useful - when exception messages are not displayed in the output. I have added property **ErrorsOutputMode** (Rethrow, Display and Suppress) in class  [RenderParameters:](https://github.com/dotliquid/dotliquid/pull/254/commits/f79df8839c36dfecdf67bf432f291a05088e0fc1#diff-91493848de600edec6ae89320d02961c). Property **RethrowErrors** is marked as obsolete, but works the same way as before.

Unit tests [TestErrorsOutputModeRethrow, TestErrorsOutputModeSuppress are TestErrorsOutputModeDisplay](https://github.com/dotliquid/dotliquid/pull/254/commits/f79df8839c36dfecdf67bf432f291a05088e0fc1#diff-b999650e03f5ed466d3a44beeb28b68e) illustrate this change

2. If the application uses a template code written by the end user, in some cases this can lead to significant loads. For example, the following code:
```
{% for i in (1..100000) %}
  {{ i }}
{% endfor %}
```
To provide control in such situations, I added two properties in class [RenderParameters:](https://github.com/dr0wzee/dotliquid/blob/e1e0ec231d473d5c399748b5e108b3e60d233c48/src/DotLiquid/RenderParameters.cs)

** MaxIterations ** - maximum number of iterations for the For tag. If it is exceeded, then the [MaximumIterationsExceededException](https://github.com/dotliquid/dotliquid/pull/254/commits/f79df8839c36dfecdf67bf432f291a05088e0fc1#diff-24e3a4f50ec0df4a40f75ab2f68aa0de) will be thrown. A value MaxIterations <= 0 indicates that there are no restrictions.

** Timeout ** - rendering timeout in ms. If the template rendering takes longer than the time specified in the Timeout property, then an TimeoutException will be thrown. A value Timeout <= 0 indicates that there is no limit.

Unit tests: [TestMaximumIterationsExceededError](https://github.com/dotliquid/dotliquid/pull/254/commits/f79df8839c36dfecdf67bf432f291a05088e0fc1#diff-b999650e03f5ed466d3a44beeb28b68e), [TestTimeoutError](https://github.com/dotliquid/dotliquid/pull/254/commits/e1e0ec231d473d5c399748b5e108b3e60d233c48#diff-b999650e03f5ed466d3a44beeb28b68e)

I hope this makes sense. 
Thanks

bradphelan:
Thanks for that info. Sounds good. Could you update wiki here

https://github.com/dotliquid/dotliquid/wiki/DotLiquid-for-Developers

with the info you typed. This is useful integration stuff.

PR #255, opened 2017-08-04
Fix formatting of code example in README

ErikSchierboom:


laedit:
Thanks a lot !
That was bothering me since a while but haven't got the time to do the change 😄

ErikSchierboom:
You're welcome!

Issue #221, opened 2017-01-12
Inherited property values: Hash.FromAnonymousObject


pk-picturepark:
Hi, 

Up until dotliquid version 2.0.30 the method Hash.FromAnonymousObject could process property values of inherited classes.

Sample:

```
public class BaseTestClass
{
  public string Name { get; set; }
}

public class TestClass : BaseTestClass
{}
```

When passing an instance of TestClass into Hash.FromAnonymousObject, the property "Name" is now omitted.

I couldn't find any changelog info since version 1.8.0. So is this a new restriction or a bug to be resolved?

Thanks

albertospelta:
You should find this message in your template `Errors` collection:

> Object 'TestClass' is invalid because it is neither a built-in type nor implements ILiquidizable

As you can read [here](https://github.com/dotliquid/dotliquid/wiki/DotLiquid-Drops ), types of parameters are now limited to .NET primitive types and some collection types.
So, in order to support arbitrary types, you should inherit from `Drop` class.

``` c#
public class BaseTestClass : Drop
{
    public string Name { get; set; }
}

public class TestClass : BaseTestClass
{
}

class Program
{
    static void Main(string[] args)
    {
        var template = Template.Parse(@"Hello {{ user.Name }} !");
	var variable = Hash.FromAnonymousObject(new
	{
		user = new TestClass()
		{
			Name = "Alberto"
		}
	});

	var result = template.Render(variable);
	template.Errors.ForEach(Console.WriteLine);
	Console.WriteLine(result);

	Console.ReadKey();
    }
}
```

TigerPoint:
If you have multi-level inheritance then I suggest converting your object to a `Dictionary` then using `Hash.FromDictionary` instead

microalps:
I am unable to reproduce any of the above in a current version. @TigerPoint do you have a test case? This might have been fixed in the latest build as @robin-parker changed the behavior of nested dictionaries.

microalps:
Closing as we haven't heard back from the user.

PR #259, opened 2017-08-14
Hash.FromAnonymousObject also gets properties from base class.

farukterzioglu:
Fixes issue  #221


laedit:
Thanks for the update 😄 

Can you just add some tests?

farukterzioglu:
I wanted to add some test but couldn't find a proper place for testing this. (and delayed tests for later :( )
Is there a test class for Hash or do  I need to create a new test class for Hash behaviors? 

Can you lead me to the right way?

laedit:
Sorry for not providing enough context.
Each source class should have the equivalent test class, in this case it seems that you have to create a new `HashTests` class.

src/DotLiquid/Hash.cs
```diff
@@ -48,7 +48,9 @@ private static void FromAnonymousObject(object anonymousObject, Hash hash, bool
 
         private static Action<object, Hash> GetObjToDictionaryMapper(Type type, bool includeBaseClassProperties)
         {
-            if (!mapperCache.TryGetValue(type, out Action<object, Hash> mapper))
+            var cacheKey = type.FullName + (includeBaseClassProperties ? "WithBaseProperties" : "WithoutBaseProperties");
```
farukterzioglu: to use same class with/without base properties at the same time, caching should consider including base properties.

src/DotLiquid/Hash.cs
```diff
@@ -24,25 +24,33 @@ public class Hash : IDictionary<string, object>, IDictionary
...
-            Action<object, Hash> mapper = GetObjToDictionaryMapper(anonymousObject.GetType());
+            Action<object, Hash> mapper = GetObjToDictionaryMapper(anonymousObject.GetType(), includeBaseClassProperties);
             mapper.Invoke(anonymousObject, hash);                
         }
 
-        private static Action<object, Hash> GetObjToDictionaryMapper(Type type)
+        private static Action<object, Hash> GetObjToDictionaryMapper(Type type, bool includeBaseClassProperties)
         {
-            if (!mapperCache.TryGetValue(type, out Action<object, Hash> mapper))
+            var cacheKey = type.FullName + (includeBaseClassProperties ? "WithBaseProperties" : "WithoutBaseProperties");
```
laedit: Can you add a separator between the type name and the suffix?

farukterzioglu: Done

laedit:
Thanks!

PR #262, opened 2017-08-21
Fix nested dictionary check

### LINKED ISSUE #252: Working with Dynamic/ExpandoObjects
Hi all,

I'm trying to use dotliquid to transform xml, in particular, soap envelopes. Moreover, I'm dynamically invoking the webservices so I don't have the proxies. All of this works just fine. I then was trying to use dotliquid to transform the soap response envelope, converting the xml to a dynamic object, but I cannot get the Render to work properly.

Any thoughts on this?
Thanks.

```csharp
[TestMethod()]
        public void Test2()
        {
            var soap = @"<soap:Envelope xmlns:soap='http://schemas.xmlsoap.org/soap/envelope/' xmlns:xsi='http://www.w3.org/2001/XMLSchema-instance' xmlns:xsd='http://www.w3.org/2001/XMLSchema'> 
   <soap:Body> 
      <GetArrayResponse xmlns='http://smartbear.com'>
         <GetArrayResult> 
            <int>0</int> 
            <int>1</int> 
            <int>2</int> 
            <int>3</int> 
            <int>4</int> 
            <int>5</int> 
            <int>6</int> 
            <int>7</int> 
            <int>8</int> 
            <int>9</int> 
         </GetArrayResult> 
      </GetArrayResponse> 
   </soap:Body> 
</soap:Envelope>";

            XDocument xDoc = XDocument.Parse(soap);
            var root = stripNS(xDoc.Root);
            XmlDocument xD = new XmlDocument();
            xD.LoadXml(root.ToString());
            var json = JsonConvert.SerializeXmlNode(xD.FirstChild);
            dynamic obj = JsonConvert.DeserializeObject<dynamic>(json);

            Template.RegisterSafeType(typeof(MyGenericInterface<>), new[] { "Value" });
            Template template = Template.Parse("{{context.Value.Envelope}}");
            var hash = Hash.FromAnonymousObject(new { context = new MyGenericImpl<dynamic> { Value = obj } });
            var output = template.Render(hash);
        }

        public interface MyGenericInterface<T>
        {
            T Value { get; set; }
        }

        public class MyGenericImpl<T> : MyGenericInterface<T>
        {
            public T Value { get; set; }
        }

        static XElement stripNS(XElement root)
        {
            return new XElement(
                root.Name.LocalName,
                root.HasElements ?
                    root.Elements().Select(el => stripNS(el)) :
                    (object)root.Value
            );
        }
```

ErikSchierboom:
The `Hash.FromDictionary` method takes an `IDictionary<string, object>` parameter. Within the method, there is a nested dictionary check, but that check then requires a `Dictionary<string, object>` (which is the concrete implementation, not the interface). This PR fixes that and also adds two unit tests for the dictionary as variable use case and two tests for using dynamic (`ExpandoObject`).

This should also fix #252

### DIFF
--- a/src/DotLiquid.Tests/ContextTests.cs
+++ b/src/DotLiquid.Tests/ContextTests.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Dynamic;
 using System.Linq;
 using DotLiquid.Exceptions;
 using NUnit.Framework;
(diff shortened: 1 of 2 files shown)


laedit:
Excellent, thanks a lot! 😃

ErikSchierboom:
You're welcome!

PR #263, opened 2017-08-21
Fix culture-specific TestTimes test

ErikSchierboom:
The `TestTimes` test failed on my machine (Dutch culture). This PR makes the culture explicit to have the test succeed.


laedit:
Thanks!

PR #266, opened 2017-09-13
Ensure that it is possible to use an explicit culture everywhere

bradphelan:
We found that the following test would fail on the users computer if they had their culture set different to what we expect. Our code should run independant of the culture of the current computer.

        private class ActionDisposable : IDisposable
        {
            private readonly Action _Action;

            public ActionDisposable(Action action) => _Action = action;

            public void Dispose() => _Action();
        }
        IDisposable SetCulture(CultureInfo ci)
        {
            var old = CultureInfo.CurrentCulture;
            CultureInfo.CurrentCulture = ci;
            return new ActionDisposable( ()=>CultureInfo.CurrentCulture = old );
        }

        [Test]
        public void ParsingWithCommaDecimalSeparatorShouldWorkWhenPassedCultureIsDifferentToCurrentCulture()
        {
            var ci = new CultureInfo(CultureInfo.CurrentCulture.Name)
            {
                NumberFormat =
                      {
                          NumberDecimalSeparator = ","
                          , NumberGroupSeparator = "."
                      }
            };
            using (SetCulture( ci ))
            {
                var t = Template.Parse( "{{2.5}}" );
                var result = t.Render( new Hash(), CultureInfo.InvariantCulture );

                Assert.AreEqual( result, "2.5" );
            }
        }

Before fixes the result would be 25 even though the invariantculture was being passed to render. It turns out that there were other places in the code that pulled the default culture. This fix changes that. The majority of the changes are just passing the culture from the outerscope through to where calls like

    float.TryParse

are actually used. Messy but necessary.

I have commit rights but I'd appreciate if somebody could review and give feedback if this is acceptable.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -32,7 +32,7 @@ after_build:
   - cmd: nuget pack src/DotLiquid/DotLiquid.nuspec -Symbols -Version "%APPVEYOR_BUILD_VERSION%" -OutputDirectory build\pkg
 
 test_script:
-  - opencover.console -target:"C:\Tools\NUnit3\nunit3-console.exe" -targetargs:"src\DotLiquid.Tests\bin\Release\net451\DotLiquid.Tests.dll --where=cat!=windows --result=TestsResults.xml;format=AppVeyor" -output:TestsCoverage.xml -filter:"+[DotLiquid]*" -register:user -returntargetcode
+  - opencover.console -target:"C:\Tools\NUnit3\nunit3-console.exe" -targetargs:"src\DotLiquid.Tests\bin\Release\net461\DotLiquid.Tests.dll --where=cat!=windows --result=TestsResults.xml;format=AppVeyor" -output:TestsCoverage.xml -filter:"+[DotLiquid]*" -register:user -returntargetcode
   - codecov -f TestsCoverage.xml
 
 artifacts:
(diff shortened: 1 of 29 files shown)


laedit:
@bradphelan Here's my review even though the branch have already been merged.
I am good for most part, but there are some blockers for me:

 - change public APIs without a good reason: why not propose an overload which takes a `CultureInfo` and the old will just call the new APIs with a default culture?
 - change the .net target framework to .net 4.6 without any reason (or have I missed it?)
 - adding references to a package without any reason (or have I missed it?)

This version will be a no-go for a lot of consumer of DotLiquid which are not and/or can't go to .net 4.6.

bradphelan:
* I only updated to 4.6 for the unit test because I couldn't set the culture in the thread in 4.5. The main library is still 4.5
* With regards to changing the API. That was hard. Because the behavior was already broken. If the user set the Culture in the Render call it would not work. Getting the culture into all the right places required adding it to the RenderParameters as well. I could have made it optional in the RenderParameters but then it was too easy to forget to add it and end up with a different culture as to the passed into Render. It might be possible to get away with just the culture being passed to Render but I think that would have been a bigger refactoring than I would have liked to do.

laedit:
I am really sorry, I reviewed it too quickly.

So if I am understanding well, the standard usage (as indicated in the wiki, Template.Parse/Render) will continue to work without any modification needed, only advanced usage will?

bradphelan:
Yes that is correct.

laedit:
Ok, so that PR is good for me, thanks a lot! 😃

bradphelan:
Thanks for the review. I appreciate it 👍

PR #273, opened 2017-10-18
Added support for strip, lstrip and rstrip filters

darrelmiller:
The [Liquid docs](https://github.com/Shopify/liquid/wiki/Liquid-for-Designers#standard-filters) includes strip,  lstrip and rstrip.  This PR adds those filters to the StandardFilters class.


laedit:
Thanks a lot! I will just close and reopen the PR to relaunch the Travis build but otherwise all seems fine 😄

Issue #195, opened 2016-07-22
Rebase PRs


haf:
Here's a list of stale PRs https://github.com/dotliquid/dotliquid/pulls?utf8=%E2%9C%93&q=is%3Apr%20is%3Aclosed%20label%3A%22rebase-me%22

Anyone can pick these up, should they want to.

PR #274, opened 2017-10-19
Adds PermissiveNamingConvention to allow snake_case, PascalCase and camelCase variables and filters.

gerfen:


src/DotLiquid/Strainer.cs
```diff
@@ -103,11 +103,15 @@ public void AddMethodInfo(string rawName, object target, MethodInfo method)
 
         public bool RespondTo(string method)
         {
+            method = Template.NamingConvention.GetMemberName(method);
```
haf: Why this change?

haf:
plx provide a description of your PR and rebase it on master.

haf:
You should also test your code, please.

src/DotLiquid/Strainer.cs
microalps: When registering methods, we use this function to build a static list of methods. PR wants to allow use of aliases, so calling GetMemberName normalizes the name to PascalCase and thus matches the defined list of methods even if used as pascalCase. This change obviously affects all naming conventions, for better or for worse.

daviburg: Well ouch then, that (affects all naming conventions) is a problem right?

microalps: I'm not sure of the impact, it needs more study. In C# we return name as-is anyways, so it would do nothing (although I would suggest normalizing there too). In Ruby, it would allow replace_first (current behavior) and ReplaceFirst and replaceFirst but NOT replacefirst or rePlaceFirst. I don't see a negative impact (although some negative tests might become positive)

Another option is to extend the interface to have an explicit way to get all alternatives for a given name and register them in the first place. If it's not implemented, it is strict exact match. Maybe an IAliasNamingConvention to maintain backward compatibility?

daviburg: I *think* it's ok to accept more name formats for filters. What happens if the input had content which names clashed with the newly allowed filter name format?

microalps: If you register class named test and Test, Strainer.Extend calls GetMemberName and will map to the same in ruby (and thus override). C# passes as is and can live side by side currently. This one line won't change that, but changing c# to normalize names will make it behave it like Ruby.

daviburg (approved):

microalps:
@gerfen Would you be willing to take a stab at rebasing this or are you OK with us cherry picking it and modifying? I know we are 3 years late to the party....

gerfen:
Hi @microalps,

I'm pretty slammed at the moment.  Feel free to cherry pick.

Thanks!

microalps:
Found a fundamental conversion flaw that has me stumped. If the primary naming convention is PascalCase (as StringCompararer.ToPascalCase does) and secondary is camelCase and snake_case, abbreviations cannot be properly converted. For example, is product_id supposed to be ProductId or ProductID in Pascal? What about PONumber?

daviburg:
Id is an abbreviation which is most often incorrectly capitalized in Pascal. Many think it is an acronym hence case it as `ID` (for two-letter acronyms the special case is to capitalize both letters). But the letters I and D are not the first letters of two words in identification, identifier or identity. Instead they are the two first letters of the word hence an abbreviation. So the proper capitalization is `Id` in Pascal. This is an named example of the Microsoft Framework design guidelines. The proper casing is also then `ProductId`.
As a side note, you will find cases where we Microsoft also got it wrong. But once a given API is released, it would be a breaking change to fix it so some of these are gran-fathered in.

https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions

`PONumber` is the casing in Pascal. Product order number is the expanded fragment for that abbreviation composed of a two letter acronym and the fully spelled number word. It follows the same rules as the example `IOStream`.

microalps:
Thanks for the clarity. But what is the snake case expectation in this case? po_number according to code in this task, will map to PoNumber, which doesn't exist either.

daviburg:
I have no experience with snake_case. As a naming scheme popular with Ruby, a fellow Ruby user should provide that input.

microalps:
Adding to the pile referenced by #195

Issue #275, opened 2017-10-19
map and sort filters violate safe-type allowed members (Drops concept breach)


nytmo:
### Dotliquid version
2.0.183

### Expected behavior
**map** and **sort** filters should obey safe-type allowed members collections.

### Actual behavior
**map** and **sort** filters access directly object members through reflection ignoring allowed members collection of the accessed type

### Steps to reproduce the Problem (you can add files)

	public class Employee
	{
		public string Name { get; set; }
		public decimal Salary { get; set; }
	}

	var template = DotLiquid.Template.Parse("{{employees | sort:'Salary' | map:'Name' | join:', '}} | {{employees | map:'Salary' | join:', '}}");
	DotLiquid.Template.RegisterSafeType(typeof(Employee), new[] { nameof(Employee.Name) });
	var result = template.Render(DotLiquid.Hash.FromDictionary(new Dictionary<string, object> {
		{ "employees", new [] { new Employee { Name = "A", Salary = 456 }, new Employee { Name = "B", Salary = 123 } } }
	}));

This results in:
`B, A | 456, 123`

But using the `Salary` property is prohibited as it's not declared as allowed member via RegisterSafeType method, so an empty string or null should be returned instead of its true value.

This is heavy security issue and it violates the Drops concept.

nytmo:
I propose this solution, which solves not only the security breach, but also naming conventions of property used as a map filter's parameter (#272) and makes map filter resilient to null input (#269)

	public static IEnumerable Map(IEnumerable input, string property)
	{
		if (input == null)
			return null;

		List<object> ary = input.Cast<object>().ToList();
		if (!ary.Any())
			return ary;

		if ((ary.All(o => o is IDictionary)) && ((IDictionary)ary.First()).Contains(property))
			return ary.Select(e => ((IDictionary)e)[property]);

		return ary.Select(e => {
			if (e == null)
				return null;

			var drop = e as DropBase;
			if (drop == null)
			{
				var type = e.GetType();
				var safeTypeTransformer = Template.GetSafeTypeTransformer(type);
				if (safeTypeTransformer != null)
					drop = safeTypeTransformer(e) as DropBase;
				else
				{
					var attr = type.GetTypeInfo().GetCustomAttributes(typeof(LiquidTypeAttribute), 	false).FirstOrDefault() as LiquidTypeAttribute;
					if (attr != null)
						drop = new DropProxy(e, attr.AllowedMembers);
				}
			}
[...]

Issue #269, opened 2017-09-14
Sort and map filters aren't resilient to nulls


tpluscode:
### Dotliquid version

2.0.183

### Expected behavior

When `sort` and `map` filter is passed null it should silently ignore it.

### Actual behavior

Depending on the setup `ArgumentNullException` or `NullReferenceException` is thrown

### Steps to reproduce the Problem (you can add files)

``` c#
var tmpl = DotLiquid.Template.Parse("x: '{{ test.deep | sort }}'");
var model = Hash.FromDictionary(new Dictionary<string, object>
{
   { "test", null }
});

tmpl.Render(model);
```

### Workaround

Here's a simple fix we applied in our project

``` c#
public static class ResilientStandardFilters
{
    public static IEnumerable Sort(object input, string property = null)
    {
        if (input == null)
        {
            return new object[0];
        }

        return StandardFilters.Sort(input, property);
    }

    public static IEnumerable Map(IEnumerable input, string property)
    {
        if (input == null)
        {
            return new object[0];
        }

        return StandardFilters.Map(input, property);
    }
}
```

nytmo:
Consider changing the Map's filter first row from:
`List<object> ary = input.Cast<object>().ToList();`
to:
`List<object> ary = input.Cast<object>().Where(o => o != null).ToList();`

tpluscode:
That would fix null elements of the `input` collection, which I don't think is broken by the way. The problem is when `input` itself in null.

nytmo:
Tested null elements and it's a problem, as later in Map method there is this:
`if (ary.All(o => o.RespondTo(property)))`
which leads to `NullReferenceException` when `o is null`

tpluscode:
I'd have to try again but from what you're saying there are in fact two possibilities for improvement?

nytmo:
Yes, there are two bugs concerning null handling. And I have reported another issue concerning naming conventions of the property name used in map filter. But it's kind a different story.

nytmo:
Resolved by PR #294 . The issue may be closed, I guess.

PR #276, opened 2017-10-19
Update Map filter to use SafeTypeTransformer to get object's member

nytmo:
partially solving issues #275 , #269
fully solving issue #272

### DIFF
--- a/src/DotLiquid.Tests/Helper.cs
+++ b/src/DotLiquid.Tests/Helper.cs
@@ -42,5 +42,10 @@ namespace DotLiquid.Tests
             public string PropAllowed { get; set; }
             public string PropDisallowed { get; set; }
         }
+
+        public class DataObjectDrop : Drop
+        {
+            public string Prop { get; set; }
+        }
     }
 }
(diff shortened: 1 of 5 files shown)


nytmo:
What is the pourpose of this test?

	CollectionAssert.AreEqual(new[] { new { a = 1 }, new { a = 2 }, new { a = 3 }, new { a = 4 } },
		StandardFilters.Map(new[] { new { a = 1 }, new { a = 2 }, new { a = 3 }, new { a = 4 } }, "b"));

It doesn't make sence to return the same collection, when there is no "b" property! The result should be collection of four empty objects or four nulls, but not the same collection.

laedit:
Thanks for this PR, I will take a look at it as soon as I can.

Regarding your question about the test, I have absolutely no idea. Maybe it is the original behavior from liquid, otherwise I must say that I think the same way as you.

nytmo:
OK, I've written unit test cases for every new use case I might have created, but the coverage isn't changing. Strange enough, that it shows no coverage/no hits on lines where I can prove hit by debugging the test... Really strange...

laedit:
Sounds good to me, thanks a lot! 😄 
About the unit test, I thinks it's ok, maybe the use of `?.` disturbs the code coverage utility.
If you are done with the modifications I will merge the PR.

nytmo:
I've finished the modifications around the map filter. You surely can merge this PR.
The next step will be the sort filter, that suffers from kind of similar issues, but I don't have enough time for this right now. Hopefully, I won't forget about it.

PR #277, opened 2017-10-24
Update README.markdown

watkinsmatthewp:
Fix broken link to old docs


laedit:
Thanks a lot for your interest, the link is ok but there is an error during the website build on AppHarbor which prevent the new version to be deployed.
I will keep your PR open for now and will merge it if the problem resolution takes too much time.

Issue #278, opened 2017-10-25
Support wildcard member filters


watkinsmatthewp:
I'm decorating my drop classes with `[LiquidType]` and passing in the allowed members as directed:

```cs
[LiquidType("my_value")]
public class MyPOCO
{
  public MyObject my_value { get; set; }
}
```

However, on large classes, this list can get lengthy, and cumbersome to modify. If a developer adds or renames a new member he/she wants to expose several lines down, it's easy to forget to update the list on the class attribute. 

I _could_ make my models inherit from `Drop` (sometimes) or write an implementation of `ILiquidizable`, but the inheritance approach is overkill (and sometimes not possible), and the `ILiquidizable` route has a lot of the same problems. Ultimately I love the attribute decorator approach and want to build on that.

I have two suggestions of what I would like to see as helpful features:

## Member-level attribute decoration

Use member-level attribute decorators to specify which members are allowed. This removes the need for the potentially long list of magic strings at the top of your class file and makes it refactor-proof

```cs
[LiquidType]
public class MyPOCO
{
  [LiquidType]
  public MyObject my_value { get; set; }
}
```

## Wildcard / member-type specifiers

Alternatively, if member-level attributes are not possible, in the cases where I would like to be able to specify that all current and future members of the class should be accessible to the renderer, a wildcard match would be nice:

```cs
[LiquidType("*")]
public class MyPOCO
{
  public MyObject my_value { get; set; }
}
```

Bonus points if you could specify an enum value in the attribute constructor indicating which member types to match on  (Property, Method, etc):

```cs
[LiquidType("*", MemberType.Property)]
public class MyPOCO
{
  public MyObject my_value { get; set; }
}
```

In the example above, properties would be allowed, but not methods.

laedit:
@watkinsmatthewp sorry for the delay, I haven't got the time to study your issue until now.

I think the the `[LiquidType("*")]` can do the trick, it is safe in the way that the dev still need to indicate it explicitly.

@dotliquid/core-team any though on this?

watkinsmatthewp:
I would really prefer member-level attribute decorators and base-class member allowance inheritance over the asterisk if at all possible.

PR #280, opened 2017-11-03
#278: Wildcard global member allowance

watkinsmatthewp:
#278 (Wildcard global member allowance):

- Modified Drop to allow all members if the filter is ["*"]
- Added unit tests to verify correct behavior


haf:
Looks good! Thank you!

PR #282, opened 2017-11-10
Fixed Issue with For Loop Not Working in Sandbox

mike9182:
The For loop functionality was throwing an Exception when it was ran within a Sandbox. I started creating a unit test to reproduce this, however I started running into issues with the System.Runtime.Remoting library not being available for .NETCore, which seemed to prevent the DotLiquid.Tests project from building.  I was attempting to follow the link below to setup the test.

https://docs.microsoft.com/en-us/dotnet/framework/misc/how-to-run-partially-trusted-code-in-a-sandbox

### DIFF
--- a/src/DotLiquid/Tags/For.cs
+++ b/src/DotLiquid/Tags/For.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections;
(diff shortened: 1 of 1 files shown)


laedit:
Thanks! 😃

PR #287, opened 2017-12-12
Added Enum type to be allowed by default (similar to integral types) (#118)

pfeigl:
This allows using Enums very similar to using integral types by allowing Enum in Liquidize method.

Code has been blindly "copied" from the old pull request in #118, but as the forked repository was no longer available, I could rebase the changes.

Added all required unit tests and everything seems to work fine.
Might be good to re-open #118 until this is merged. 

Thanks,
Philipp

### DIFF
--- a/src/DotLiquid.Tests/ConditionTests.cs
+++ b/src/DotLiquid.Tests/ConditionTests.cs
@@ -1,8 +1,10 @@
+using System;
+using System.Collections.Generic;
 using System.Globalization;
 using System.Text.RegularExpressions;
-using DotLiquid.Exceptions;
 using DotLiquid.NamingConventions;
 using NUnit.Framework;
+using ArgumentException = DotLiquid.Exceptions.ArgumentException;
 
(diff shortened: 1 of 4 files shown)


nytmo:
If this is allowed, how do you attach safe type transformer to enum type?
Let's have an enum Visibility { Hidden, Collapsed, Visible } And I want my transformer to produce _Off_ when Collapsed or Hidden and _On_ when Visible. This PR totally disqualifies this mechanism and therefore is backward incompatible. Am I wrong?

TerraVenil:
@pfeigl do you need any help to resolve conflicts ? I also interested in current functionality.

pfeigl:
@TerraVenil tbh we moved away from using DotLiquid quite some time ago and use alternatives.

Feel free to take over the comimt and send an new PR. I do currently not have the time to update this commit (also with the added danger of it still not beeing merged).

microalps:
Replaced by #449

Issue #290, opened 2017-12-17
Date filter makes dates lose milliseconds during formatting


nytmo:
### Dotliquid version
2.0.232

### Expected behavior
Let's have a date 2017-12-17 18:48:31.256, once formatted using filter **date:'ffffff'** it should return **256000**

### Actual behavior
Now the filter **date:'ffffff'** results to **000000**

###
The problem is on the line:
**value = input.ToString();**
of the date filter.
because, if _input_ is of DateTime type, ToString() method on DateTime type results in a string without milliseconds. The algorithm is first converting a datetime to string and then parsing it back, which is OK unless the _object_ is already a DateTime type, at that point the logic of converting DateTime to string and then parsing it back to DateTime leads to incorrect behaviour.

nytmo:
Resolved by PR #291 .

PR #291, opened 2017-12-17
Date filter modification - now allowing milliseconds formatting

nytmo:
This PR resolves issue #290. If the input object is DateTime type, then it is formatted straight away. Otherwise the old way is used.

### DIFF
--- a/src/DotLiquid.Tests/StandardFilterTests.cs
+++ b/src/DotLiquid.Tests/StandardFilterTests.cs
@@ -277,8 +277,8 @@ namespace DotLiquid.Tests
 
             Assert.AreEqual("05/07/2006 10:00:00", StandardFilters.Date("05/07/2006 10:00:00", string.Empty));
             Assert.AreEqual("05/07/2006 10:00:00", StandardFilters.Date("05/07/2006 10:00:00", null));
-            Assert.AreEqual(new DateTime(2006, 7, 5, 10, 0, 0), StandardFilters.Date("5/7/2006 10:00:00", string.Empty));
-            Assert.AreEqual(new DateTime(2006, 7, 5, 10, 0, 0), StandardFilters.Date("5/7/2006 10:00:00", null));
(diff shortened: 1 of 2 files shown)


nytmo:
Something must have gone wrong with Travis CI:
E: Unable to locate package dotnet-dev-1.0.3
E: Couldn't find any package by glob 'dotnet-dev-1.0.3'
E: Couldn't find any package by regex 'dotnet-dev-1.0.3'
Your build has been stopped.

laedit:
I have also fixed the Travis build in master, a merge should fix your PR.

nytmo:
I guess I need some help with clarifying, what's the desired behaviour when format is null/empty.
Is this really correct?
Assert.AreEqual("08/01/2006 10:00:00", StandardFilters.Date("08/01/2006 10:00:00", string.Empty));
Because .ToString() method run on datetime object results in _8/1/2006 10:00:00 AM_ (according to current culture of the build server).
But the Date filter (before my modifications) has ignored the date parsing when format was null/empty and returned the input string. BUT, if you pass "now" keyword to date filter with empty/null filter, it returns "now" string again and not the current time... Because of the null/empty condition on filter that comes first. So what's the correct/expected behaviour? Without your answer I'm not able to write those tests as I would just make them up so that it would work, but it wouldn't have to be the correct way.

laedit:
Since the "now" and "today" aren't the target of this PR they should be ignored: the behavior was flawed before et it is not the goal of this PR to fix it.

But for the DateTime object as input, I think that without format (null or string.empty) it should return `.ToString()`, as before.

Is that ok?

nytmo:
There you go. Everything should be ready now. I hope my brain won't explode, as I let the Date filter flawed as far as "now"/"today" is concerned. I hate to let things unrepaired...

laedit:
Yeah, sorry about that. But if you are motivated to do another PR about that "now"/"today" thing, please be my guest! 😄

nytmo:
Motivated? It's not an option to me. I just have to do that to establish a ballance in the nature again.

PR #293, opened 2017-12-18
Date filter - "now" reserved string should return correct date time even without format string

nytmo:


nytmo:
Here it is. I may go (shopping) in peace...

laedit:
Thanks a lot!

nytmo:
You're welcome. Looking forward to find these merges on nuget :)

laedit:
The good news is that the nuget publication is done after each merge in master, so the new version has already been published (2.0.251), even if the nuget website should display it only tomorrow after his indexation.

Issue #295, opened 2017-12-21
Context.AddFilter() does nothing when rendering a Template


mjashanks:
### Dotliquid version
Using commit: baaf2a6f4ff0d1cad8ccbdbcb234d49574bb85dd
### Expected behavior
Adding a filter function, via Context.AddFilter, should work when using a Template and a Drop
### Actual behavior
The filter function is not registered, and therefor not used
### Steps to reproduce the Problem (you can add files)

PR #296 demonstrates the issue with a unit test.

microalps:
@mjashanks have you ever resolved this? Please update task and PR if this is all good. It seems to me that the issue with the unit test is that you are setting localVariables to the renderParameters, but context is already created so it doesn't override the context. The below snippet works for me:

```
var context = new Context(CultureInfo.InvariantCulture);
var renderParams = RenderParameters.FromContext(context, context.FormatProvider);
context.Environments.Add( Hash.FromAnonymousObject(data) );
```

mjashanks:
Hey, I haven't revisited this, so am happy for you to close this. Thanks for getting back :)

PR #296, opened 2017-12-21
filter functions do not work with templates & drops

mjashanks:
Hello.

I'm not expecting this PR to be accepted as, honestly, I haven't really taken the time to properly understand the code base.

Hopefully you can see from the Test what my issue is - that I cannot get filter functions to work when using a Template with a Drop (All the filter function tests use a context and variable directly).

I have made some changes to make the test pass, and all other tests pass. However, I am not sure if what I have done is the correct thing to do.

see issue #295

### DIFF
--- a/src/DotLiquid.Tests/FunctionFilterTests.cs
+++ b/src/DotLiquid.Tests/FunctionFilterTests.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Globalization;
 using NUnit.Framework;
+using DotLiquid.NamingConventions;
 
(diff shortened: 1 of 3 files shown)


haf:
@mjashanks Hi there! Appreciate the honesty! What would it take you to be more sure? Could you spend that time, and we can get this merged?

It seems to be a pretty broad change from cursory inspection.

microalps:
Underlying issue was closed by author. Correct implementation was offered. Closing.

PR #297, opened 2017-12-29
Slice filter resilient to negative start index out of bounds

nytmo:
Example: If the input string is of length 2, start is -4 and len = 3, then the result should be just the first letter of the input string. Now an exception "Index out of range" is thrown.


laedit:
Thanks!
Could you add some tests?

nytmo:
Some tests have been added. Enjoy :)

haf:
Thanks @nytmo !

PR #302, opened 2018-01-25
Changed Dictionary Handling, So that an interator is available

vitalragaz:
With that dictionary can now accessed by their handle as also by a for loop.


haf:
Thanks!

PR #306, opened 2018-04-17
Update to support decimal in "times" filter

jinhong-:
Extended to support decimal


laedit:
Thanks for this PR! :smile: 
Can you add some tests for it please?

jinhong-:
Added!

laedit:
Thanks a lot! :+1:

PR #310, opened 2018-07-22
Sorting when only some elements contain the sorting criteria

rolandbaer:
On sorting, it could be that a collection to sort has some elements that contain the sorting property but some other elements do not contain this property. This should not prevent the sorting to take place.
This PR changes the behaviour of the sorting to take place it the sorting property is at least available in one element of the collection.


laedit:
Thanks a lot, I will review it as soon as possible 😄

laedit:
Can you add a test which demonstrate the new ability?
Thanks.

rolandbaer:
@laedit Test is added.

laedit:
Thanks!

PR #312, opened 2018-08-07
Added default liquidize support for enum types

Matti-Koopa:
Enum types now be liquidized normally without registering them all as safe types.
This works fine "out of the box" as the enum value will get printed using ToString().


laedit:
Thanks for your PR!
Could you please add some tests of the feature?

microalps:
Replace by #449

PR #313, opened 2018-08-14
Support to %s dateformat

mz1991:
Added support to %s Ruby dateformat.
%s - Number of seconds since 1970-01-01 00:00:00 UTC.


laedit:
Thanks for your PR!
It seems that the AppVeyor had a temporary fail, since I don't have the rights to launch a new build, could you touch a file and push a new commit?

laedit:
Thanks! :+1:

PR #315, opened 2018-08-17
Standard filters

mz1991:
Added two standard filters.
Uniq: Removes any duplicate elements in an array.
UrlDecode: convert a input string to URLDECODE.


src/DotLiquid.Tests/Util/StrFTimeTests.cs
```diff
@@ -23,6 +23,7 @@ public class StrFTimeTests
         [TestCase("%m", ExpectedResult = "01")]
         [TestCase("%M", ExpectedResult = "32")]
         [TestCase("%p", ExpectedResult = "PM")]
+		[TestCase("%s", ExpectedResult = "1326033134")]
```
laedit: Could you rebase from master to avoid the already merged commit?

src/DotLiquid/StandardFilters.cs
```diff
@@ -95,6 +95,18 @@ public static string UrlEncode(string input)
...
+		/// <summary>
+        /// convert a input string to URLDECODE
+        /// </summary>
+        /// <param name="input"></param>
+        /// <returns></returns>
+		public static string UrlDecode(string input)
+        {
+            return input == null
+                ? input
+                : Uri.UnescapeDataString(input);
```
laedit: [`UnescapeDataString`](https://docs.microsoft.com/en-us/dotnet/api/system.uri.unescapedatastring?view=netframework-4.7.2) doesn't handle the `+` to space conversion and I think that Liquid does since the [`url_encode` filter](http://shopify.github.io/liquid/filters/url_encode/) converts the space to `+`. If it is true, a `.Replace('+',' ')` should be added.

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -579,6 +588,28 @@ public void TestCapitalize()
...
+			Assert.AreEqual(new List<object> {5}, StandardFilters.Uniq(5));
+        }
+		
+		[Test]
+        public void TestAbs()
+        {
+			Assert.AreEqual(0, StandardFilters.Abs("notNumber"));
+            Assert.AreEqual(10, StandardFilters.Abs(10));
+            Assert.AreEqual(5, StandardFilters.Abs(-5));
+            Assert.AreEqual(19, StandardFilters.Abs(19.86));
```
laedit: `Abs` shouldn't round a number as I understand from the [liquid doc](https://shopify.github.io/liquid/filters/abs/).

mz1991: yes, my bad. Fixing soon.

src/DotLiquid/StandardFilters.cs
mz1991: Yes, that's true. I replace with HttpUtility.UrlEncode / UrlDecode

src/DotLiquid.Tests/Util/StrFTimeTests.cs
mz1991: Done

src/DotLiquid/DotLiquid.csproj
```diff
@@ -35,6 +35,7 @@
   <ItemGroup>
     <PackageReference Include="SonarAnalyzer.CSharp" Version="5.11.0.1761" />
     <PackageReference Include="System.ValueTuple" Version="4.3.0" />
+	<Reference Include="System.Web" />
```
laedit: The reference to `System.Web` bothers me, is it possible to do without it?

mz1991: Yes, I agree. I used `System.Net`

laedit:
Thanks!

Issue #320, opened 2018-09-16
Iterating over Nested Dictionary in For Loop


danielcrabtree:
It is not possible to loop over the elements of a nested dictionary using a for loop.

For example, this does not work:
```
var dictionary = new Dictionary<string, object> {{
    "People", new Dictionary<string, object> {
        { "ID1", "Jane" },
        { "ID2", "Mike" }
    }
}};

// This currently outputs empty string "" instead of "JaneMike"
Helper.AssertTemplateResult("JaneMike", "{% for item in People %}{{ item.Value }}{%endfor%}", Hash.FromDictionary(dictionary));
```

The issue appears to have been caused by the PR for #157.

In my example above, we have a dictionary containing a dictionary containing string values - 2 levels of dictionary.

In #157, there is a dictionary containing a dictionary containing a dictionary containing string values - 3 levels of dictionary.

The PR for #157 added code that intercepts values where the 2nd level is a dictionary, but then only outputs values when the 3rd level is also a dictionary. When the 3rd level is not a dictionary, as in my example where it is a string, it doesn't set anything in the context (since value is Hash is false in BuildContext).

It is possible that nested dictionaries were not supported properly at the time of #157, but simply removing the code added for #157 now resolves all these issues.

There is also the issue that #157 had a bug in it's test case:
```
Helper.AssertTemplateResult("JaneMike", "{% for item in People %}{{ item.First }}{%endfor%}", Hash.FromDictionary(dictionary));
```

This test case just ignores the 2nd level entirely and dots into the 3rd level. You wouldn't expect that when iterating over the 2nd level dictionary.

Instead, the test for #157 should have been:
```
Helper.AssertTemplateResult("JaneMike", "{% for item in People %}{{ item.Value.First }}{%endfor%}", Hash.FromDictionary(dictionary));
```

Here you are explicitly stepping through the 2nd level. It is possible that #157 was never an issue and the original writer was just unaware of .Value or perhaps .Value was not supported at that time.

I have submitted a PR which fixes this issue. It also adds 2 new tests and updates the third. 

One new test is for my example, the 2nd level dictionary. 
The other new test shows how you can substitute a list for the dictionary at the 2nd level to keep the direct item.First access desired in #157.
The updated test fixes the original test for #157 to step through the 2nd level dictionary using .Value.

microalps:
@daviburg What are your thoughts here? Is this still an issue (I didnt test) and are there any reservations to fixing this?

PR #321, opened 2018-09-16
Fix Iterating over Nested Dictionary in For Loop AND Fix Capture Variable Support

danielcrabtree:
This PR fixes the issues discussed in #320.

### DIFF
--- a/src/DotLiquid.Tests/Tags/StandardTagTests.cs
+++ b/src/DotLiquid.Tests/Tags/StandardTagTests.cs
@@ -81,6 +81,36 @@ namespace DotLiquid.Tests.Tags
                 }
             } };
 
+            Helper.AssertTemplateResult("JaneMike", "{% for item in People %}{{ item.Value.First }}{%endfor%}",
+                Hash.FromDictionary(dictionary));
+        }
+
+        [Test]
+        public void TestForWithSingleLevelNestedDictionary()
+        {
+            var dictionary = new Dictionary<string, object> { {
+            "People",
+            new Dictionary<string, object> {
+                    { "ID1", "Jane" },
+                    { "ID2", "Mike" }
+                }
+            } };
+
+            Helper.AssertTemplateResult("JaneMike", "{% for item in People %}{{ item.Value }}{%endfor%}",
+                Hash.FromDictionary(dictionary));
+        }
+
+        [Test]
+        public void TestForWithListNestedDictionary()
+        {
+            var dictionary = new Dictionary<string, object> { {
+            "People",
+            new List<object> {
+                    new Dictionary<string, object>{ { "First", "Jane" }, { "Last", "Green" } },
+                    new Dictionary<string, object>{ { "First", "Mike" }, { "Last", "Doe" } }
+                }
+            } };
+
             Helper.AssertTemplateResult("JaneMike", "{% for item in People %}{{ item.First }}{%endfor%}",
                 Hash.FromDictionary(dictionary));
         }
--- a/src/DotLiquid.Tests/Tags/StandardTagTests.cs
+++ b/src/DotLiquid.Tests/Tags/StandardTagTests.cs
@@ -365,6 +365,14 @@ namespace DotLiquid.Tests.Tags
             Helper.AssertTemplateResult("content foo content foo ",
                 "{{ var2 }}{% capture var2 %}{{ var }} foo {% endcapture %}{{ var2 }}{{ var2 }}", assigns);
         }
+        
+        [Test]
+        public void TestCaptureWithDashes()
+        {
+            Hash assigns = Hash.FromAnonymousObject(new { var = "content" });
+            Helper.AssertTemplateResult("content foo content foo ",
+                "{{ var-2 }}{% capture var-2 %}{{ var }} foo {% endcapture %}{{ var-2 }}{{ var-2 }}", assigns);
+        }
 
         [Test]
         public void TestCaptureDetectsBadSyntax()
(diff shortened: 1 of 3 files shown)


danielcrabtree:
This also now fixes #331

microalps:
@daviburg This change is breaking, but it is the correct behavior IMHO. The test is changed to reference using item**.Value.**First

src/DotLiquid/Tags/Capture.cs
```diff
@@ -22,7 +22,7 @@ namespace DotLiquid.Tags
     /// </summary>
     public class Capture : DotLiquid.Block
     {
-        private static readonly Regex Syntax = R.C(@"(\w+)");
+        private static readonly Regex Syntax = R.B(R.Q(@"({0}+)"), Liquid.VariableSignature);
```
microalps, line 25: Variable signature includes dotted and bracketed notations. This should be a simple variable name imho. Other regex just include hyphen.

src/DotLiquid.Tests/Tags/StandardTagTests.cs
```diff
@@ -81,6 +81,36 @@ public void TestForWithNestedDictionary()
                 }
             } };
 
+            Helper.AssertTemplateResult("JaneMike", "{% for item in People %}{{ item.Value.First }}{%endfor%}",
```
daviburg, line 84: Formatting should be cleaned-up.
If we move parameters to multiple lines, each parameter should start on a new line.

daviburg: ... and use named parameters as we have at least two strings.

src/DotLiquid.Tests/Tags/StandardTagTests.cs
```diff
@@ -81,6 +81,36 @@ public void TestForWithNestedDictionary()
...
             } };
 
+            Helper.AssertTemplateResult("JaneMike", "{% for item in People %}{{ item.Value.First }}{%endfor%}",
+                Hash.FromDictionary(dictionary));
+        }
+
+        [Test]
+        public void TestForWithSingleLevelNestedDictionary()
+        {
+            var dictionary = new Dictionary<string, object> { {
```
daviburg, line 91: Fix the formatting here also. I assume that there are other use of nested bracketed collection initializers to refer to for consistency.

daviburg (changes requested):

daviburg:
> @daviburg This change is breaking, but it is the correct behavior IMHO. The test is changed to reference using item**.Value.**First

Sorry I don't see an existing test been changed in this PR. Could you help me understand the breaking change? What worked before that won't work after this?

microalps:
> > @daviburg This change is breaking, but it is the correct behavior IMHO. The test is changed to reference using item**.Value.**First
> 
> Sorry I don't see an existing test been changed in this PR. Could you help me understand the breaking change? What worked before that won't work after this?

Yes its hard to see, but if you look carefully at the before and after, the old test for TestForWithNestedDictionary is on line 114 and new one is line 84. The old test remains, but now is applied to a different data set. The difference between these two lines is a ".Value". I missed it several times until I ran it locally.

daviburg:
> > > @daviburg This change is breaking, but it is the correct behavior IMHO. The test is changed to reference using item**.Value.**First
> > 
> > 
> > Sorry I don't see an existing test been changed in this PR. Could you help me understand the breaking change? What worked before that won't work after this?
> 
> Yes its hard to see, but if you look carefully at the before and after, the old test for TestForWithNestedDictionary is on line 114 and new one is line 84. The old test remains, but now is applied to a different data set. The difference between these two lines is a ".Value". I missed it several times until I ran it locally.

I'm really dense today...
image

image

So before

```
        [Test]
        public void TestForWithNestedDictionary()
        {
            var dictionary = new Dictionary<string, object> { {
            "People", 
            new Dictionary<string, object> {
                    { "ID1", new Dictionary<string, object>{ { "First", "Jane" }, { "Last", "Green" } } },
                    { "ID2", new Dictionary<string, object>{ { "First", "Mike" }, { "Last", "Doe" } } }
                }
            } };

            Helper.AssertTemplateResult("JaneMike", "{% for item in People %}{{ item.First }}{%endfor%}",
                Hash.FromDictionary(dictionary));
        }
```

And after

```
        [Test]
        public void TestForWithNestedDictionary()
        {
            var dictionary = new Dictionary<string, object> { {
            "People", 
            new Dictionary<string, object> {
                    { "ID1", new Dictionary<string, object>{ { "First", "Jane" }, { "Last", "Green" } } },
                    { "ID2", new Dictionary<string, object>{ { "First", "Mike" }, { "Last", "Doe" } } }
                }
            } };

            Helper.AssertTemplateResult("JaneMike", "{% for item in People %}{{ item.Value.First }}{%endfor%}",
                Hash.FromDictionary(dictionary));
        }
```

Ah, ok, going from `{{ item.First }}` to `{{ item.Value.First }}`

Yeah we can't do that as `{{ item.First }}` was explicitly working before.

I didn't understand before that the 'star' characters were to highlight the keyword.

You are not using .First however in the new tests. Can as backward compatibility .First be a shortcut for .Value.First? I'm not really thinking this through, just tossing ideas.

microalps:
From a performance perspective, I'm not a fan of it and just lets more people continue the same mistake without knowing. Maybe we can cut this code as part of a opt-in flag.

microalps:
Closing this pull request as the bigger issue was resolved in another PR and the minor issue was incorrectly fixed as mentioned in code review.

Issue #336, opened 2019-03-01
Hash.FromAnonymousObject doens't work on class hierarchies


rwb196884:
### Dotliquid version
2.0.298

### Expected behavior
`Hash.FromAnonymousObject(thing, true)` has a key for every property of every class in the class heirarchy of `thing`.

### Actual behavior
`Hash.FromAnonymousObject(thing, true)` only has a key for every property of the type of  `thing` and its immediate base class; properties of the base class of the base class (etc.) of `thing` are omitted.

### Steps to reproduce the Problem (you can add files)
```
    public class Bottom
    {
        public string BottomValue { get; set; }

        public object ToLiquid() => Hash.FromAnonymousObject(this, true);
    }

    public class Middle : Bottom
    {
        public string MiddleValue { get; set; }
    }

    public class Top : Middle
    {
        public string TopValue { get; set; }
    }

    [TestFixture]
    public class TestLiquidInheritance : UKPlc.UnitTest.SetUpFixture
    {
        [TestCase]
        public void TestRenderInheritance()
        {
            Top t = new Top()
            {
                BottomValue = "Bottom",
                MiddleValue = "Middle",
                TopValue = "Top"
            };

            Hash h = (Hash)t.ToLiquid();
            Assert.AreEqual(h.Keys.Count, 3); // Fails because Bottom is omitted.            
        }
    }
```

rwb196884:
Hello?
Anyone give a fuck?

microalps:
@rwb196884 Care to clean up your two PRs into a single PR for review? Also, there is no test for child class overriding base class object with same name (although code seems to have that)

rwb196884:
I don't know how so you'll have to take them or leave them.

PR #337, opened 2019-03-01
Update Hash.cs

rwb196884:
Fix for #336 
Recurse down the class hierarchy to get properties -- not just the first base class.

### DIFF
--- a/src/DotLiquid/Hash.cs
+++ b/src/DotLiquid/Hash.cs
@@ -75,9 +75,26 @@ namespace DotLiquid
             return mapper;
         }
 
-        private static void AddBaseClassProperties(Type type, List<PropertyInfo> propertyList) {
-            propertyList.AddRange(type.GetTypeInfo().BaseType.GetTypeInfo().DeclaredProperties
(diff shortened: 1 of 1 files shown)


src/DotLiquid/Hash.cs
```diff
@@ -75,9 +75,26 @@ private static void FromAnonymousObject(object anonymousObject, Hash hash, bool
             return mapper;
         }
 
-        private static void AddBaseClassProperties(Type type, List<PropertyInfo> propertyList) {
-            propertyList.AddRange(type.GetTypeInfo().BaseType.GetTypeInfo().DeclaredProperties
-                .Where(p => p.CanRead && p.GetMethod.IsPublic && !p.GetMethod.IsStatic).ToList());
+        private static void AddBaseClassProperties(Type type, List<PropertyInfo> propertyList)
+        {
+            //If the type is null we have no BaseType most like never happen as you should run in to System.Object 
+            if (type == null || string.Equals(type.ToString(), "System.Object"))
```
daviburg: type.Equals(typeof(System.Object))

PR #346, opened 2019-05-29
Wrap rethrown exception to preserve stack trace

RicoSuter:
Throwing a catched exception with `throw ex` here will rewrite the stack trace and I no longer see the stack trace where the exception actually happened. Wrapping it in another exception with inner exception will preserve the original stack trace in the inner exception. 

I have a strange issue (seems to be a concurrency problem) and the only way to debug is to check the stack trace - but at the moment there is no way to get the original stack trace.

### DIFF
--- a/src/DotLiquid/Context.cs
+++ b/src/DotLiquid/Context.cs
@@ -169,7 +169,7 @@ namespace DotLiquid
                 return string.Empty;
 
             if (_errorsOutputMode == ErrorsOutputMode.Rethrow)
-                throw ex;
+                throw new Exception("Could not render the Liquid template.", ex);
 
             if (ex is SyntaxException)
             {


RicoSuter:
Wont work because we are not in the catch clause but another method (HandleError)

RicoSuter:
Ref: https://rsuter.visualstudio.com/NJsonSchema/_build/results?buildId=1818&view=ms.vss-test-web.build-test-results-tab&runId=1001780&resultId=100080&paneView=debug

Current stack trace ends at HandleError

mylemans:
Not 'throw ex;' but 'throw;' ...

RicoSuter:
This does only work in a catch clause...

mylemans:
True, I missed that, my bad. Maybe move up the check? As in, inside Block.RenderAll check if rethrow is needed and do it there?

RicoSuter:
We'd need to `throw;` it here: 
https://github.com/dotliquid/dotliquid/blob/24491a6844014f98376b100ebbd276e139de4c98/src/DotLiquid/Block.cs#L192

But in this class we have no access (yet?) to the rethrow settings...

mylemans:
Maybe try/catch HandleError for a specific exception that indicates a rethrow is required, and then just rethrow (outside the catch, set a bool or something)

RicoSuter:
> Maybe try/catch HandleError for a specific exception that indicates a rethrow is required, and then just rethrow (outside the catch, set a bool or something)

?

Outside the catch clause you cannot rethrow with stack trace preserving.

src/DotLiquid/Context.cs
```diff
@@ -169,7 +169,7 @@ public string HandleError(Exception ex)
                 return string.Empty;
 
             if (_errorsOutputMode == ErrorsOutputMode.Rethrow)
-                throw ex;
+                throw new Exception("Could not render the Liquid template.", ex);
```
daviburg, line 172: This is not the right fix - to preserve the stack the proper way is to just do a

throw;

in the original catch block.

This helper is only called there (Block::RenderAll). That code need to change to be told to rethrow in these circumstances.

daviburg:
> Just rethrow? (throw;)

Yes that's what needs happening, although it will take work as at the moment the rethrow decision point is in a helper - helper itself won't have access to rethrow.

daviburg:
> > Maybe try/catch HandleError for a specific exception that indicates a rethrow is required, and then just rethrow (outside the catch, set a bool or something)
> 
> ?
> 
> Outside the catch clause you cannot rethrow with stack trace preserving.

Hence you need the helper to tell the caller it needs to rethrow.

BoyaWu10:
It seems we can use `ExceptionDispatchInfo` to preserve the stack trace ([reference](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.exceptionservices.exceptiondispatchinfo?view=net-5.0)). PR #405 is created with this change.

daviburg:
As https://github.com/dotliquid/dotliquid/pull/405 has addressed this issue, closing this PR for the same issue.

PR #348, opened 2019-08-28
Fix object extension methods to take into account Drops...

jmalarcon:
… that overwrite the indexer to get their values.

With the current implementation of the `RespondTo` and `Send` methods in `DotLiquid.Util.ObjectExtensionMethods`, if you have a `Drop` derived class that overrides the indexer in `DropBase` so that it can return arbitrary properties (such as those defined in the Front-Matter of a file), the current methods aren't able to deal with them. **This prevents being able to sort using those arbitrary properties with the sort filter**.

Fortunately, the solution is pretty straight forward: just 2 and 6 lines of code respectively, as you can see in the PR.

In the following block, below, I've pasted a simple console program that you can use to test how this works with the original code and with the fixed one. It tries to sort bu name or length the Markdown files inside a folder. In a real situation the fields would be taken from the Front-Matter.

Hope you can accept the PR, generate a new version and update the NuGet package. 

Thanks for your great work on this library!!

-----
**Sample code below**:

```csharp
using System;
using System.Linq;
using System.IO;
using DotLiquid;

namespace TestSort
{
    class Program
    {
        static void Main(string[] args)
        {
            DirectoryInfo di = new DirectoryInfo(@"C:\TestFiles\");
            var allFiles = (from file in di.EnumerateFiles("*.md", SearchOption.TopDirectoryOnly)
                            select new MarkdownFile(file.Name, file.Length)
                           );
            Template parser = Template.Parse(@"
{% assign sorted = allFiles | sort: 'length' %}
{% for file in sorted %}
- {{file.Name}}
{% endfor %}");
            Hash context = new Hash(allFiles);
            string res = parser.Render(context);
            Console.WriteLine(res);
            Console.ReadLine();
        }
    }

    //A sample Drop overriding the indexer from the base class
    public class MarkdownFile : Drop
    {
        private string fullName;
        private long length;

        public MarkdownFile(string fullName, long length)
        {
            this.fullName = fullName;
            this.length = length;
        }

        public override object this[object fieldName]
        {
            get
            {
                //Tries to return a real property form the object if present
                object res = base[fieldName];
                if (res == null)
                {
                    switch (fieldName.ToString().ToLowerInvariant())
                    {
                        case "name":
                            res = fullName.ToLowerInvariant();
                            break;
                        case "length":
                            res = length;
                            break;
                        default:
                            res = "";
                            break;
                    }
                }

                return res;
            }
        }
    }
}
```

### DIFF
--- a/src/DotLiquid/MIISdotliquid.csproj
+++ b/src/DotLiquid/MIISdotliquid.csproj
@@ -6,7 +6,7 @@
     <NeutralLanguage>en-US</NeutralLanguage>
     <Authors>Tim Jones;Alessandro Petrelli</Authors>
     <TargetFrameworks>netstandard1.3;net45</TargetFrameworks>
-    <AssemblyName>DotLiquid</AssemblyName>
+    <AssemblyName>MIISdotliquid</AssemblyName>
     <AssemblyOriginatorKeyFile>../Formosatek-OpenSource.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
     <PublicSign Condition=" '$(OS)' != 'Windows_NT' ">true</PublicSign>
@@ -30,6 +30,7 @@
     <DebugType>full</DebugType>
     <DebugSymbols>True</DebugSymbols>
       <DocumentationFile>bin\Debug\$(TargetFramework)\DotLiquid.xml</DocumentationFile>
+      <RootNamespace>DotLiquid</RootNamespace>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/DotLiquid/Properties/AssemblyInfo.cs
+++ b/src/DotLiquid/Properties/AssemblyInfo.cs
@@ -1,4 +1,4 @@
-﻿using System.Reflection;
+using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 
@@ -6,8 +6,8 @@ using System.Runtime.InteropServices;
 // set of attributes. Change these attribute values to modify the information
 // associated with an assembly.
 
-[assembly: AssemblyTitle("DotLiquid")]
-[assembly: AssemblyDescription("DotLiquid is a templating system ported to the .NET framework from Ruby’s Liquid Markup.")]
+[assembly: AssemblyTitle("MIISDotLiquid")]
+[assembly: AssemblyDescription("Fixed version for MIIS - DotLiquid is a templating system ported to the .NET framework from Ruby’s Liquid Markup.")]
 [assembly: AssemblyConfiguration("")]
 [assembly: AssemblyCompany("Formosatek Ltd")]
 [assembly: AssemblyProduct("DotLiquid")]
@@ -25,8 +25,8 @@ using System.Runtime.InteropServices;
 
 [assembly: Guid("75d1dc49-0097-4f60-b5fd-0e43d213b38b")]
 
-[assembly: AssemblyVersion("1.0.0.0")]
-[assembly: AssemblyFileVersion("1.0.0.0")]
-[assembly: AssemblyInformationalVersion("1.0.0.0")]
+[assembly: AssemblyVersion("2.0.315.0")]
+[assembly: AssemblyFileVersion("2.0.315.0")]
+[assembly: AssemblyInformationalVersion("2.0.315.0")]
 
 [assembly: InternalsVisibleTo("DotLiquid.Tests, PublicKey=002400000480000094000000060200000024000052534131000400000100010093ae26e2c87851b659e9847a0a9c6088a4ad1988df9b176d56c3996e33458273df5c2138b5bf13b2352a99152f10ef1bc2564069179d5344ba723a875ea048b80fcb34c1c5ff7e3d131cb208140265e5144183570d1e0433c1a37959720e0d8d83a7ee870d5e0dd904afc62663103eb2e2105e1eddeadfe876c9ccc90a31cfbf")]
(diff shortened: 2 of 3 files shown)


src/DotLiquid/MIISdotliquid.csproj
```diff
@@ -6,7 +6,7 @@
     <NeutralLanguage>en-US</NeutralLanguage>
     <Authors>Tim Jones;Alessandro Petrelli</Authors>
     <TargetFrameworks>netstandard1.3;net45</TargetFrameworks>
-    <AssemblyName>DotLiquid</AssemblyName>
+    <AssemblyName>MIISdotliquid</AssemblyName>
```
daviburg: Assembly rename merge to master... Why?

src/DotLiquid/Properties/AssemblyInfo.cs
```diff
@@ -1,13 +1,13 @@
...
 using System.Runtime.InteropServices;
 
 // General Information about an assembly is controlled through the following 
 // set of attributes. Change these attribute values to modify the information
 // associated with an assembly.
 
-[assembly: AssemblyTitle("DotLiquid")]
-[assembly: AssemblyDescription("DotLiquid is a templating system ported to the .NET framework from Ruby’s Liquid Markup.")]
+[assembly: AssemblyTitle("MIISDotLiquid")]
+[assembly: AssemblyDescription("Fixed version for MIIS - DotLiquid is a templating system ported to the .NET framework from Ruby’s Liquid Markup.")]
```
daviburg: Single customer fix with change name for everyone can't be merged back to master

src/DotLiquid/Properties/AssemblyInfo.cs
```diff
@@ -25,8 +25,8 @@
 
 [assembly: Guid("75d1dc49-0097-4f60-b5fd-0e43d213b38b")]
 
-[assembly: AssemblyVersion("1.0.0.0")]
-[assembly: AssemblyFileVersion("1.0.0.0")]
-[assembly: AssemblyInformationalVersion("1.0.0.0")]
+[assembly: AssemblyVersion("2.0.315.0")]
```
daviburg: Probably the 1.0.0.0 version is a placeholder for build to overwrite with actual build value. Eitherway, the assembly build version should not be baked in the code

src/DotLiquid/Util/ObjectExtensionMethods.cs
```diff
@@ -11,6 +11,9 @@ public static bool RespondTo(this object value, string member, bool ensureNoPara
             if (value == null)
                 throw new ArgumentNullException("value");
 
+            if (value is DropBase)
+                return (value as DropBase).ContainsKey(member);
```
daviburg: to avoid repeat type casting C# allows in current version to use the shortcut
if (value is DropBase dropBase)
 return dropBase.FooBar();

daviburg:
You need to add test cases matching to your proposed new functionality.

daviburg:
Please create a feature suggestion in the issue backlog and link it here to separate discussion on the merit of the feature vs the review of the implementation.

src/DotLiquid/MIISdotliquid.csproj
jmalarcon: This is a later commit I did and wasn't in the original pull request 😞 

Since nobody seemed to answer the original pull request and I needed this fixed to use it in my project, I decided to use my fork as a new project, independent of your original DotNetLiquid. To use a different name and being able to distribute it with my other project (MIIS), I changed the Assembly name. I'd prefer to use just DotLiquid and not my own version, but since nobody seemed to be in the project anymore...

This can be ignored if you decide to incorporate the fix (and the first commit) and I happily will use DotLiquid instead.

src/DotLiquid/Properties/AssemblyInfo.cs
jmalarcon: Yes, you're right. Sorry for that. I wouldn't have changed it if I had expected your response, but after a lot of time without it, I decided to go solo. See previos file comment. Thanks.

src/DotLiquid/Properties/AssemblyInfo.cs
jmalarcon: Yes, same reason as the first comment. I was flying solo, but I gladly will move to DotLiquid again when my original fix (or a similar one) is accepted. I can go back and make a new PR if you're in the project and willing to accept it. Thanks.

jmalarcon:
> Please create a feature suggestion in the issue backlog and link it here to separate discussion on the merit of the feature vs the review of the implementation.

I've just created [a new issue for this](https://github.com/dotliquid/dotliquid/issues/364). If you want me to create a new PR with just the fix and removing the Assembly name and version changes, I'll be more than happy to do it. Thanks!

jmalarcon:
BTW, in a couple of weeks I'll be in Redmond for a whole week for the MVP summit. We can discuss anything in person if you want 😊

src/DotLiquid/Util/ObjectExtensionMethods.cs
```diff
@@ -11,6 +11,9 @@ public static bool RespondTo(this object value, string member, bool ensureNoPara
             if (value == null)
                 throw new ArgumentNullException("value");
 
+            if (value is DropBase)
```
daviburg, line 14: Late comment inspired by @microalps comment in the open feature request: should test for is IIndexable contract instead, as that's where the ContainsKey method is defined. Same below, as the array accessor is also defined in IIndexable.

microalps:
We are closing this issue as we have not gotten a response to the related issue. Please reopen this issue when you update with more info.

PR #354, opened 2019-11-11
Updated to make Rendering and FileSystem operations async by default.

mwadams:
This is a WIP PR that is **not really intended to be merged**, but I am creating it for discussion. (Hence it is created as a draft.)

It is a significant breaking change that changes the file system and rendering operations to be async-by-default.

I would like to create some side-by-side benchmarking to understand the perf impact of these changes in various scenarios.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -12,3 +12,29 @@ trim_trailing_whitespace = true
 end_of_line = lf
 indent_style = space
 indent_size = 2
+
+[*.cs]
+
+# IDE0065: Misplaced using directive
+csharp_using_directive_placement = inside_namespace:none
+
+# IDE0008: Use explicit type
+csharp_style_var_when_type_is_apparent = true:none
+
+# IDE0008: Use explicit type
+csharp_style_var_for_built_in_types = false:none
+
+# IDE0008: Use explicit type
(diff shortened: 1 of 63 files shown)


daviburg:
Having async as an option, great. Forcing everything to async as a breaking change... maybe not great.

mwadams:
I entirely agree!

microalps:
As this seems to have forked to another library and no feedback on how we can make this backwards compatible, can we close this PR?

OpenSpacesAndPlaces (commented):
Is there any reason that adding async/await is being held up? Seems like a good upgrade - especially for heavily modularized rendering.

daviburg:
> Is there any reason that adding async/await is being held up? Seems like a good upgrade - especially for heavily modularized rendering.

This is held up because there is no contributor addressing the acknowledge issues with the draft PR. The draft was done in a major breaking way. If a contributor wanted to step up, the change to asynchronous would need to be introduced as a set of parallel async public APIs, so the existing sync public APIs are not altered, and no existing client is broken.

microalps:
This PR is being abandoned for the following reasons:
* This PR would result in breaking changes as is
* There are unresolved merge conflicts
* Author has not responded in more than a year

PR #359, opened 2020-02-20
Update

rpnitdgp:
Adding parameter to do string to number change validation


rpnitdgp:
Creating variable to test string to number translation

daviburg:
Why would we need a same typed second parameter in the demo class in the master branch?

microalps:
DotLiquidMarkup site is not intended for unit tests. Feel free to run more complex tests in LinqPad or .NET fiddle.

Issue #360, opened 2020-02-25
Integer support limited to 32 bit is below RFC-8259 recommended integer range of ~64 bit long


daviburg:
### Dotliquid version
Any

### Expected behavior
Per compliance with RFC-8259 "[...]numbers that are integers and
   are in the range [-(2**53)+1, (2**53)-1] are interoperable[...]"
https://tools.ietf.org/html/rfc8259#section-6

### Actual behavior
Liquid error: Value was either too large or too small for an Int32.

### Steps to reproduce the Problem (you can add files)
https://stackoverflow.microsoft.com/questions/187325
https://stackoverflow.microsoft.com/questions/185882

Input

{   
    "stringasint": "10038449706009",
    "intasint":10038449706009   
}
Map

{
    "addnumbertostringasint":"{{content.stringasint |Plus: 2}}",
    "addnumbertointasint":"{{content.intasint |Plus: 2}}"
}
Result

{
  "addnumbertostringasint": "100384497060092",
  "addnumbertointasint": "Liquid error: Value was either too large or too small for an Int32."
}

PR #362, opened 2020-02-25
Support integer values for numbers beyond the range of 32 bit integer…

daviburg:
…s but within the RFC allowed 64 bit integers.
#360

### DIFF
--- a/src/DotLiquid.Tests/ConditionTests.cs
+++ b/src/DotLiquid.Tests/ConditionTests.cs
@@ -64,7 +64,7 @@ namespace DotLiquid.Tests
         public void TestContainsWorksOnArrays()
         {
             _context = new Context(CultureInfo.InvariantCulture);
-            _context["array"] = new[] { 1, 2, 3, 4, 5 };
+            _context["array"] = new long[] { 1, 2, 3, 4, 5 };
 
             AssertEvaluatesFalse("array", "contains", "0");
             AssertEvaluatesTrue("array", "contains", "1");
@@ -223,10 +223,12 @@ namespace DotLiquid.Tests
             try
             {
                 Condition.Operators["IsMultipleOf"] =
-                    (left, right) => (int)left % (int)right == 0;
+                    (left, right) => (long)left % (long)right == 0;
 
                 // exact match
                 AssertEvaluatesTrue("16", "IsMultipleOf", "4");
+                AssertEvaluatesTrue("2147483646", "IsMultipleOf", "2");
+                AssertEvaluatesTrue("2147483648", "IsMultipleOf", "2");
                 AssertEvaluatesFalse("16", "IsMultipleOf", "5");
 
                 // lower case: compatibility
@@ -266,10 +268,12 @@ namespace DotLiquid.Tests
                 try
                 {
                     Condition.Operators["DivisibleBy"] =
-                        (left, right) => (int)left % (int)right == 0;
+                        (left, right) => (long)left % (long)right == 0;
 
                     // exact match
                     AssertEvaluatesTrue("16", "DivisibleBy", "4");
+                    AssertEvaluatesTrue("2147483646", "DivisibleBy", "2");
+                    AssertEvaluatesTrue("2147483648", "DivisibleBy", "2");
                     AssertEvaluatesFalse("16", "DivisibleBy", "5");
 
                     // lower case: compatibility
--- a/src/DotLiquid.Tests/CustomIndexableTests.cs
+++ b/src/DotLiquid.Tests/CustomIndexableTests.cs
@@ -22,7 +22,7 @@ namespace DotLiquid.Tests
                 get
                 {
                     long? index = key as long?;
-                    if (!index.HasValue || index.Value < 0 || index.Value >= items.Length) {
+                    if (!index.HasValue || index.Value < 0L || index.Value >= items.Length) {
                         throw new KeyNotFoundException();
                     }
                     return items[index.Value];
@@ -32,7 +32,7 @@ namespace DotLiquid.Tests
             public bool ContainsKey(object key)
             {
                 long? index = key as long?;
-                return index.HasValue && index.Value >= 0 && index.Value < items.Length;
+                return index.HasValue && index.Value >= 0L && index.Value < items.Length;
             }
 
             public IEnumerator GetEnumerator()
@@ -70,7 +70,7 @@ namespace DotLiquid.Tests
         public void TestVirtualListLoop()
         {
             string output = Template.Parse("{%for item in list%}{{ item }} {%endfor%}")
-                .Render(Hash.FromAnonymousObject(new {list = new VirtualList((long)1, "Second", (long)3)}));
+                .Render(Hash.FromAnonymousObject(new {list = new VirtualList(1L, "Second", 3L)}));
             Assert.AreEqual("1 Second 3 ", output);
         }
 
@@ -78,7 +78,7 @@ namespace DotLiquid.Tests
         public void TestVirtualListIndex()
         {
             string output = Template.Parse("1: {{ list[0] }}, 2: {{ list[1] }}, 3: {{ list[2] }}")
-                .Render(Hash.FromAnonymousObject(new {list = new VirtualList((long)1, "Second", (long)3)}));
+                .Render(Hash.FromAnonymousObject(new {list = new VirtualList(1L, "Second", 3L)}));
             Assert.AreEqual("1: 1, 2: Second, 3: 3", output);
         }
 
--- a/src/DotLiquid.Tests/CustomIndexableTests.cs
+++ b/src/DotLiquid.Tests/CustomIndexableTests.cs
@@ -1,4 +1,4 @@
-﻿using System.Collections;
+using System.Collections;
 using System.Collections.Generic;
 using NUnit.Framework;
 
@@ -21,7 +21,7 @@ namespace DotLiquid.Tests
             {
                 get
                 {
-                    int? index = key as int?;
+                    long? index = key as long?;
                     if (!index.HasValue || index.Value < 0 || index.Value >= items.Length) {
                         throw new KeyNotFoundException();
                     }
@@ -31,7 +31,7 @@ namespace DotLiquid.Tests
 
             public bool ContainsKey(object key)
             {
-                int? index = key as int?;
+                long? index = key as long?;
                 return index.HasValue && index.Value >= 0 && index.Value < items.Length;
             }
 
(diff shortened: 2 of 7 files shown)


Paul-Dempsey:
Couple of observations:  (mostly in tests)  why `(long)100` (cast) versus `100L` typed long constant? In a modern compiler, they should be equivalent as an optimization, but this is reliant on a particular implementation.

Also, in context.cs when parsing a range, should the numeric elements of the range also be parsed as int64 (long) instead of int?

daviburg:
> Couple of observations: (mostly in tests) why `(long)100` (cast) versus `100L` typed long constant? In a modern compiler, they should be equivalent as an optimization, but this is reliant on a particular implementation.

No particular reason, I will switch to the cleaner nnnL annotation.

> Also, in context.cs when parsing a range, should the numeric elements of the range also be parsed as int64 (long) instead of int?

I hesitated about that because we are talking such large numbers that a range with such a high index will not work. It gets a little bizarre in the limit operation where the parsed upper limit goes through the context integer code (hence int 64) but the other numerics of the operation do not and stay int 32. So at some places you see me convert the int 64 back to int 32 for ~range and the like (enumerable, collection, ...). Those would through if the number doesn't fit, which i felt was right as such large collection would break anyway.

dsparkplug:
This pull request breaks the `truncate` filter with the error `System.ArgumentException : Object of type 'System.Int64' cannot be converted to type 'System.Int32'`. Add the following test to the `FilterTests` class to reproduce:

```csharp
		[Test]
		public void Truncate()
		{
			_context["var"] = "this is a longish string";
			Assert.AreEqual("this is a ", new Variable("var | truncate: 10").Render(_context));
		}
```

dsparkplug:
See pull request #365

src/DotLiquid.Tests/ConditionTests.cs
```diff
@@ -64,7 +64,7 @@ public void TestContainsWorksOnStrings()
         public void TestContainsWorksOnArrays()
         {
             _context = new Context(CultureInfo.InvariantCulture);
-            _context["array"] = new[] { 1, 2, 3, 4, 5 };
+            _context["array"] = new long[] { 1, 2, 3, 4, 5 };
```
daviburg, line 67: It was wrong for me to change the test input. Instead the library needs to continue working with int and work with long seamlessly.

src/DotLiquid/Context.cs
```diff
@@ -336,7 +336,9 @@ private object Resolve(string key, bool notifyNotFound = true)
             // Integer.
             match = IntegerRegex.Match(key);
             if (match.Success)
-                return Convert.ToInt32(match.Groups[1].Value);
+            {
+                return Convert.ToInt64(match.Groups[1].Value);
```
daviburg, line 340: Here we should have tried backward compatible Int32 first and only on overflow expand to Int64.

src/DotLiquid/Context.cs
```diff
@@ -437,14 +439,14 @@ private object Variable(string markup, bool notifyNotFound)
 
                 // If object is a KeyValuePair, we treat it a bit differently - we might be rendering
                 // an included template.
-                if (IsKeyValuePair(@object) && (part.Equals(0) || part.Equals("Key")))
+                if (IsKeyValuePair(@object) && (part.Equals(0L) || part.Equals("Key")))
```
daviburg, line 442: We need indexes to work on any integer type. Condition class already had type-insensitive equal test capability.

src/DotLiquid/Context.cs
```diff
@@ -511,7 +513,7 @@ private static bool IsHashOrArrayLikeObject(object obj, object part)
             if ((obj is IDictionary && ((IDictionary)obj).Contains(part)))
                 return true;
 
-            if ((obj is IList) && (part is int))
+            if ((obj is IList) && (part is long))
```
daviburg, line 516: We still need to allow int index as well.

src/DotLiquid/StandardFilters.cs
```diff
@@ -41,7 +41,7 @@ public static int Size(object input)
         /// <param name="start"></param>
         /// <param name="len"></param>
         /// <returns></returns>
-        public static string Slice(string input, int start, int len = 1)
+        public static string Slice(string input, long start, long len = 1)
```
daviburg, line 44: Revert to int and let caller do the narrowing conversion when safe.

src/DotLiquid.Tests/ConditionTests.cs
```diff
@@ -223,10 +223,12 @@ public void TestCapitalInCustomOperator()
             try
             {
                 Condition.Operators["IsMultipleOf"] =
-                    (left, right) => (int)left % (int)right == 0;
+                    (left, right) => (long)left % (long)right == 0;
```
daviburg, line 226: We need to preserve original support for int while allowing new custom operator to handle long values when required by the value exceeding the int range.

src/DotLiquid.Tests/CustomIndexableTests.cs
```diff
@@ -87,7 +87,7 @@ public void TestCustomIndexableIntKeys()
         {
             string output = Template.Parse("1: {{container[0]}}, 2: {{container[1]}}")
                 .Render(Hash.FromAnonymousObject(new {container = new CustomIndexable()}));
-            Assert.AreEqual("1: System.Int32 0, 2: System.Int32 1", output);
+            Assert.AreEqual("1: System.Int64 0, 2: System.Int64 1", output);
```
daviburg, line 90: Keep index keys within int range just 32 bit ints.

src/DotLiquid.Tests/CustomIndexableTests.cs
```diff
@@ -70,15 +70,15 @@ public object ToLiquid()
...
+                .Render(Hash.FromAnonymousObject(new {list = new VirtualList(1L, "Second", 3L)}));
             Assert.AreEqual("1 Second 3 ", output);
         }
 
         [Test]
         public void TestVirtualListIndex()
         {
             string output = Template.Parse("1: {{ list[0] }}, 2: {{ list[1] }}, 3: {{ list[2] }}")
-                .Render(Hash.FromAnonymousObject(new {list = new VirtualList(1, "Second", 3)}));
+                .Render(Hash.FromAnonymousObject(new {list = new VirtualList(1L, "Second", 3L)}));
```
daviburg, line 81: list index should at minimum keep supporting int, TBD if long makes any sense here.

src/DotLiquid.Tests/CustomIndexableTests.cs
```diff
@@ -21,8 +21,8 @@ internal VirtualList(params object[] items)
             {
                 get
                 {
-                    int? index = key as int?;
-                    if (!index.HasValue || index.Value < 0 || index.Value >= items.Length) {
+                    long? index = key as long?;
```
daviburg, line 24: While long are implemented array / index accessors in .NET, it works in conjunction with more common int index keys. So we need at minimum to keep supporting int index and optional support long indexes as well.

src/DotLiquid/StandardFilters.cs
```diff
@@ -602,8 +602,8 @@ public static object Minus(object input, object operand)
         /// <returns></returns>
         public static object Times(object input, object operand)
         {
-            return input is string && operand is int
-                ? Enumerable.Repeat((string)input, (int)operand)
+            return input is string && operand is long
```
daviburg, line 605: int and long both to be supported.

src/DotLiquid/Tags/For.cs
```diff
@@ -107,8 +107,8 @@ public override void Render(Context context, TextWriter result)
                     : Convert.ToInt32(context[_attributes["offset"]])
                 : 0;
 
-            int? limit = _attributes.ContainsKey("limit") ? context[_attributes["limit"]] as int? : null;
-            int? to = (limit != null) ? (int?) (limit.Value + from) : null;
+            long? limit = _attributes.ContainsKey("limit") ? context[_attributes["limit"]] as long? : null;
```
daviburg, line 110: here also, support either integer type

src/DotLiquid/Tags/For.cs
```diff
@@ -166,7 +166,7 @@ public override void Render(Context context, TextWriter result)
             });
         }
 
-        private static List<object> SliceCollectionUsingEach(Context context, IEnumerable collection, int from, int? to)
+        private static List<object> SliceCollectionUsingEach(Context context, IEnumerable collection, int from, long? to)
```
daviburg, line 169: Iteration upper limit needed 64 bit range would exceed by nature the configured iteration limit defined as 32 bit int and make little sense. While collection, say an array, can extend to long index, an iteration action in the 64 bit range of steps would take a prohibitive execution time.

src/DotLiquid/Tags/For.cs
daviburg: Actually here convert to 32 bit int only as a for loop with a limit outside the 32 bit range is unreasonable execution cost.

src/DotLiquid.Tests/ConditionTests.cs
```diff
@@ -266,10 +268,12 @@ public void TestCapitalInCustomCSharpOperator()
                 try
                 {
                     Condition.Operators["DivisibleBy"] =
-                        (left, right) => (int)left % (int)right == 0;
+                        (left, right) => (long)left % (long)right == 0;
```
daviburg, line 271: Keep the int test for backward compatibility and add the new test as a system convert to int64 to support both inputs.

src/DotLiquid.Tests/CustomIndexableTests.cs
daviburg: Long to be used when the index exceeds 32 bit integer range, which is ok for arrays (.NET allows very large arrays which accessor index is long)

src/DotLiquid.Tests/FilterTests.cs
```diff
@@ -33,12 +33,12 @@ public static string Money(object input)
 
         private static class FiltersWithArguments
```
daviburg, line 34: Ensure arguments of either bit size work.

src/DotLiquid.Tests/Tags/StandardTagTests.cs
```diff
@@ -182,7 +182,7 @@ public void TestLimiting()
         public void TestDynamicVariableLimiting()
         {
             Hash assigns = Hash.FromAnonymousObject(new { array = new[] { 1, 2, 3, 4, 5, 6, 7, 8, 9, 0 } });
-            assigns["limit"] = 2;
+            assigns["limit"] = 2L;
```
daviburg, line 185: 32 bit limit makes a lot more sense here due to compute cost of a 64 bit boundary.

daviburg:
Fixing the mess I created -> #378

PR #365, opened 2020-03-03
Fix long to integer conversion issue with truncate filters

dsparkplug:
Fixes `System.ArgumentException : Object of type 'System.Int64' cannot be converted to type 'System.Int32'` error when using `truncate` or `truncate-word` filters

### DIFF
--- a/src/DotLiquid.Tests/FilterTests.cs
+++ b/src/DotLiquid.Tests/FilterTests.cs
@@ -262,19 +262,51 @@ namespace DotLiquid.Tests
             Assert.AreEqual(" King Kong has 1000$ ", new Variable("var | bank_statement").Render(_context));
         }
 
-
         [Test]
         public void Truncate()
         {
-            _context["var"] = "this is a longish string";
-            Assert.AreEqual("this is...", new Variable("var | truncate: 10").Render(_context));
+            var variable = new Variable("\"Ground control to Major Tom.\" | truncate: 20");
+            Assert.AreEqual("Ground control to...", variable.Render(_context));
+
+            variable = new Variable("\"Ground control to Major Tom.\" | truncate: 25, \", and so on\"");
+            Assert.AreEqual("Ground control, and so on", variable.Render(_context));
+
+            variable = new Variable("\"Ground control to Major Tom.\" | truncate: 20, \"\"");
+            Assert.AreEqual("Ground control to Ma", variable.Render(_context));
+
+            variable = new Variable("\"Ground control to Major Tom.\" | truncate: 0");
+            Assert.AreEqual("...", variable.Render(_context));
+
+            variable = new Variable("\"Ground control to Major Tom.\" | truncate: -1");
+            Assert.That(() => variable.Render(_context), Throws.Exception.TypeOf<ArgumentException>());
+
+            variable = new Variable($"\"Ground control to Major Tom.\" | truncate: {((long)int.MaxValue) + 1}");
+            Assert.That(() => variable.Render(_context), Throws.Exception.TypeOf<ArgumentException>());
+
         }
 
         [Test]
         public void TruncateWords()
         {
-            _context["var"] = "this is a longish string";
-            Assert.AreEqual("this is...", new Variable("var | truncate_words: 2").Render(_context));
+
+            var variable = new Variable("\"Ground control to Major Tom.\" | truncate_words: 3");
+            Assert.AreEqual("Ground control to...", variable.Render(_context));
+
+            variable = new Variable("\"Ground control to Major Tom.\" | truncate_words: 3, \"--\"");
+            Assert.AreEqual("Ground control to--", variable.Render(_context));
+
+            variable = new Variable("\"Ground control to Major Tom.\" | truncate_words: 3, \"\"");
+            Assert.AreEqual("Ground control to", variable.Render(_context));
+
+            variable = new Variable("\"Ground control to Major Tom.\" | truncate_words: 0");
+            Assert.AreEqual("...", variable.Render(_context));
+
+            variable = new Variable("\"Ground control to Major Tom.\" | truncate_words: -1");
+            Assert.That(() => variable.Render(_context), Throws.Exception.TypeOf<ArgumentException>());
+
+            variable = new Variable($"\"Ground control to Major Tom.\" | truncate_words: {((long)int.MaxValue) + 1}");
+            Assert.That(() => variable.Render(_context), Throws.Exception.TypeOf<ArgumentException>());
+
         }
 
     }
--- a/src/DotLiquid.Tests/FilterTests.cs
+++ b/src/DotLiquid.Tests/FilterTests.cs
@@ -261,5 +261,21 @@ namespace DotLiquid.Tests
             _context.AddFilters(typeof(ContextFilters));
             Assert.AreEqual(" King Kong has 1000$ ", new Variable("var | bank_statement").Render(_context));
         }
+
+
+        [Test]
+        public void Truncate()
+        {
+            _context["var"] = "this is a longish string";
+            Assert.AreEqual("this is...", new Variable("var | truncate: 10").Render(_context));
+        }
+
+        [Test]
+        public void TruncateWords()
+        {
+            _context["var"] = "this is a longish string";
+            Assert.AreEqual("this is...", new Variable("var | truncate_words: 2").Render(_context));
+        }
+
     }
 }
--- a/src/DotLiquid.Tests/FilterTests.cs
+++ b/src/DotLiquid.Tests/FilterTests.cs
@@ -262,52 +262,5 @@ namespace DotLiquid.Tests
             Assert.AreEqual(" King Kong has 1000$ ", new Variable("var | bank_statement").Render(_context));
         }
 
-        [Test]
-        public void Truncate()
-        {
-            var variable = new Variable("\"Ground control to Major Tom.\" | truncate: 20");
-            Assert.AreEqual("Ground control to...", variable.Render(_context));
-
-            variable = new Variable("\"Ground control to Major Tom.\" | truncate: 25, \", and so on\"");
-            Assert.AreEqual("Ground control, and so on", variable.Render(_context));
-
-            variable = new Variable("\"Ground control to Major Tom.\" | truncate: 20, \"\"");
-            Assert.AreEqual("Ground control to Ma", variable.Render(_context));
-
-            variable = new Variable("\"Ground control to Major Tom.\" | truncate: 0");
-            Assert.AreEqual("...", variable.Render(_context));
-
-            variable = new Variable("\"Ground control to Major Tom.\" | truncate: -1");
-            Assert.That(() => variable.Render(_context), Throws.Exception.TypeOf<ArgumentException>());
-
-            variable = new Variable($"\"Ground control to Major Tom.\" | truncate: {((long)int.MaxValue) + 1}");
-            Assert.That(() => variable.Render(_context), Throws.Exception.TypeOf<ArgumentException>());
-
-        }
-
-        [Test]
-        public void TruncateWords()
-        {
-
-            var variable = new Variable("\"Ground control to Major Tom.\" | truncate_words: 3");
-            Assert.AreEqual("Ground control to...", variable.Render(_context));
-
-            variable = new Variable("\"Ground control to Major Tom.\" | truncate_words: 3, \"--\"");
-            Assert.AreEqual("Ground control to--", variable.Render(_context));
-
-            variable = new Variable("\"Ground control to Major Tom.\" | truncate_words: 3, \"\"");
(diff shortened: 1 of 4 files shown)


src/DotLiquid.Tests/FilterTests.cs
```diff
@@ -261,5 +261,21 @@ public void TestContextFilter()
             _context.AddFilters(typeof(ContextFilters));
             Assert.AreEqual(" King Kong has 1000$ ", new Variable("var | bank_statement").Render(_context));
         }
+
+
+        [Test]
+        public void Truncate()
```
daviburg, line 267: Is context required to expose the issue with integer conversion? Otherwise there are two existing unit tests under standard filters - TestTruncate and TestTruncateWords

src/DotLiquid.Tests/FilterTests.cs
```diff
@@ -261,5 +261,21 @@ public void TestContextFilter()
             _context.AddFilters(typeof(ContextFilters));
             Assert.AreEqual(" King Kong has 1000$ ", new Variable("var | bank_statement").Render(_context));
         }
+
+
+        [Test]
+        public void Truncate()
+        {
+            _context["var"] = "this is a longish string";
```
daviburg, line 269: Ideally cover the sample from the specification at https://shopify.github.io/liquid/filters/truncate/

{{ "Ground control to Major Tom." | truncate: 20 }}
{{ "Ground control to Major Tom." | truncate: 25, ", and so on" }}
{{ "Ground control to Major Tom." | truncate: 20, "" }}

src/DotLiquid/StandardFilters.cs
```diff
@@ -166,12 +166,14 @@ public static string H(string input)
         /// <param name="length"></param>
         /// <param name="truncateString"></param>
         /// <returns></returns>
-        public static string Truncate(string input, int length = 50, string truncateString = "...")
+        public static string Truncate(string input, long length = 50, string truncateString = "...")
```
daviburg, line 169: Is the right fix to change the type here to long? String Length in .NET is int. There is no secondary 'LongLength' of type long. Memory constrains make that the limit of int max count of character are already not reachable. A string should never reach a length requiring a 64 bit integer to describe.

src/DotLiquid/StandardFilters.cs
```diff
@@ -166,12 +166,14 @@ public static string H(string input)
...
         /// <param name="truncateString"></param>
         /// <returns></returns>
-        public static string Truncate(string input, int length = 50, string truncateString = "...")
+        public static string Truncate(string input, long length = 50, string truncateString = "...")
         {
             if (string.IsNullOrEmpty(input))
                 return input;
 
-            int l = length - truncateString.Length;
+            int lengthAsInt = length < int.MinValue ? int.MinValue : (length > int.MaxValue ? int.MaxValue : (int)length);
```
daviburg, line 174: A length input beyond int min or max appears to be an error. Hence the right behavior is to error out, rather than do a best effort to the nearest value in the range of int.

Also... what is the meaning of a negative length (int.MinValue) in this context? Any negative value may be an error condition already.

src/DotLiquid/StandardFilters.cs
```diff
@@ -166,12 +166,14 @@ public static string H(string input)
...
+
+            int l = lengthAsInt - truncateString.Length;
```
daviburg, line 176: for code readability, this is the right opportunity to rename the variable to some descriptive word or compound words. 'l' and other single letter variable names are poor choices giving little to no information to engineers trying to maintain the code. As the change requires to understand the existing purpose, do rename to something meaningful.

src/DotLiquid/StandardFilters.cs
dsparkplug: Yes, fair enough. Will change to throw an ArgumentException.

src/DotLiquid/StandardFilters.cs
dsparkplug: This is really a byproduct of pull request #362. 

The [change](https://github.com/dotliquid/dotliquid/pull/362/files#diff-15f31b73418c0b6c972c740104f051d3R336-R341) to convert any parameters that match a integer regex to a long, means that all functions need to accept long parameters - otherwise we get an ArgumentException even if the value passed in is a valid integer.

I'm not sure if there is any other way to fix this exception other than to revert pull request #362.

src/DotLiquid.Tests/FilterTests.cs
dsparkplug: OK - will do

src/DotLiquid.Tests/FilterTests.cs
dsparkplug: The two existing unit tests under standard filters did not fail before the changes in this commit, but this one did. They require the _context and I'm not sure where else they would fit, so I put them in here. Do you have any other suggestions?

src/DotLiquid/StandardFilters.cs
dsparkplug: Can do. I usually prefer a more descriptive variable names, but didn't think it my place to change existing naming here.

dsparkplug:
I don't understand why this is not more of a priority. Both the `truncate` and `truncate-word` filters are broken and cannot be used at all in the current release. The pull request #362 that broke them was approved and merged within a single day.

daviburg:
> I don't understand why this is not more of a priority. Both the `truncate` and `truncate-word` filters are broken and cannot be used at all in the current release. The pull request #362 that broke them was approved and merged within a single day.

Sorry I didn't see my notification from this follow-up. Paying attention now. I agree this regression is urgent to fix. Thank you for making it clear.

src/DotLiquid/StandardFilters.cs
daviburg: I reproduced this locally and looking at the call stack I am thinking that Strainer::Invoke or something ahead of it could do safe conversion when the available input argument exceeds the range of required input argument type, with boundary check, so that we don't modify every method that should and can only handle an int32 to handle an int64. I'm wondering if other keywords may have been broken besides Truncate. (And will search for that to answer it too.)

src/DotLiquid.Tests/FilterTests.cs
daviburg: I see the existing standard filter tests did not fail because they inline the native type input parameter rather than taking it from parsed input / liquid template. This can be done there with lines such as:

Helper.AssertTemplateResult(expected: "Hello", template: "{{ Hello | truncate:4 }}");

(That's what I used just now to reproduce this problem.)

src/DotLiquid/StandardFilters.cs
daviburg: This may be a generic approach to handle this for any liquid filter in Strainer::Invoke:

            // Attempt narrowing converstions where required by type mismatch and possible by value range.
            for (var argumentIndex = 0; argumentIndex < parameterInfos.Length; argumentIndex++)
            {
                if (args[argumentIndex] != null && args[argumentIndex].GetType() != parameterInfos[argumentIndex].ParameterType)
                {
                    args[argumentIndex] = Convert.ChangeType(args[argumentIndex], parameterInfos[argumentIndex].ParameterType);
                }
            }

src/DotLiquid.Tests/FilterTests.cs
daviburg: Corrected sample unit test:

Helper.AssertTemplateResult(expected: "H...", template: "{{ 'Hello' | truncate:4 }}");

validated with that sample that the generic approach in Strainer could work.

src/DotLiquid/StandardFilters.cs
daviburg: We also need to test for IConvertible to avoid trouble with Join filter

            // Attempt conversions where required by type mismatch and possible by value range.
            // These may be narrowing conversions (e.g. Int64 to Int32) when the actual range doesn't cause an overflow.
            for (var argumentIndex = 0; argumentIndex < parameterInfos.Length; argumentIndex++)
            {
                if (args[argumentIndex] != null
                    && args[argumentIndex].GetType() != parameterInfos[argumentIndex].ParameterType
                    && args[argumentIndex] is IConvertible)
                {
                    args[argumentIndex] = Convert.ChangeType(args[argumentIndex], parameterInfos[argumentIndex].ParameterType);
                }
            }

src/DotLiquid.Tests/FilterTests.cs
dsparkplug: I've now moved these unit tests to the existing standard filters unit test methods

src/DotLiquid/StandardFilters.cs
dsparkplug: Now implemented as recommended

dsparkplug:
Thanks for looking into this. I've now made the recommend changes and the new unit tests pass. Please test and review.

daviburg:
I am working on an extensive fix for this and other regressions caused by the earlier change. I am going through each changed line and preparing a backward compatible fix so all int(32) access are restored while long remain possible where the value exceeds the 32 bit range.

src/DotLiquid/StandardFilters.cs
```diff
@@ -166,15 +166,18 @@ public static string H(string input)
...
         /// <param name="truncateString"></param>
         /// <returns></returns>
-        public static string Truncate(string input, int length = 50, string truncateString = "...")
+        public static string Truncate(string input, long length = 50, string truncateString = "...")
         {
             if (string.IsNullOrEmpty(input))
                 return input;
 
-            int l = length - truncateString.Length;
+            if (length < 0 || length > int.MaxValue)
```
daviburg: I see how check for negative values can be an improvement. However it is also a breaking change as calls which here and below for truncate words returned just the ellipsis would now throw an argument exception or liquid error.

dsparkplug: OK, I'm happy to remove this check if its already handled elsewhere - if you want me to continue? What's your plan? Are you going to update and merge your pull request, and then add this one on top?

daviburg: Glad you are here. I've tried to take all your new test cases and some of your code improvement in my PR. There may not be a need for your PR on top, apologies to the 'take over'. Let me know if I missed sometime.

daviburg:
> Thanks for looking into this. I've now made the recommended changes and the new unit tests pass. Please test and review.

I am terrible... While heads-on reviewing my poor earlier PR I didn't see you pushed a new iteration. I will review that for anything / any good idea / any fix I missed.

src/DotLiquid/StandardFilters.cs
dsparkplug: That's fine - you can take over. I'll just close this if that's OK?

src/DotLiquid/Strainer.cs
```diff
@@ -109,8 +109,8 @@ public bool RespondTo(string method)
         public object Invoke(string method, List<object> args)
         {
             // First, try to find a method with the same number of arguments minus context which we set automatically further down.
-            var methodInfo = _methods[method].FirstOrDefault(m => 
-                m.Item2.GetParameters().Where(p => p.ParameterType != typeof(Context)).Count() == args.Count);
+            var methodInfo = _methods[method].FirstOrDefault(m =>
+                m.Item2.GetParameters().Count(p => p.ParameterType != typeof(Context)) == args.Count);
```
daviburg, line 113: That's a good change. I'll take this and add to my PR.

src/DotLiquid/Strainer.cs
```diff
@@ -133,6 +133,16 @@ public object Invoke(string method, List<object> args)
                     args.Add(parameterInfos[i].DefaultValue);
                 }
 
+            // Attempt conversions where required by type mismatch and possible by value range.
+            // These may be narrowing conversions (e.g. Int64 to Int32) when the actual range doesn't cause an overflow.
+            for (int argumentIndex = 0; argumentIndex < parameterInfos.Length; argumentIndex++)
+            {
+                var convertibleArg = args[argumentIndex] as IConvertible;
```
daviburg, line 140: That's also good, access array index only once.

src/DotLiquid/StandardFilters.cs
daviburg: Picked two good coding ideas and adopted them, plus adding pattern matching on top.

daviburg: Negative values for truncate and truncate words currently lead to just the truncate string to be returned. As it was not always obvious from reading the code, I've done small refactoring in my PR to make this explicit (and preserve current behavior).
Beside truncate and truncate-words my February PR merge broke a number of other scenarios so I restored, fixed and added systematically all the unit tests that were touched back then.
I hope that my PR now is including all of the corrections that you need, as well as the good ideas you added.

dsparkplug:
Closing as all changes now incorporated in #378

PR #370, opened 2020-04-17
Try to use decimal first for floating point operations to avoid precision error in results for financial arithmetic

### LINKED ISSUE #368: Decimal modulo has rounding errors
### Dotliquid version
Any version

### Expected behavior
modulo filter preserves decimals as-is

### Actual behavior
modulo filter introduces decimal rounding errors

{{ 148387.77 | modulo:10 }} should be 7.77 but actual response is 7.76999999998952

As this line to existing TestModulo unit test to reproduce:

            Helper.AssertTemplateResult("7.77", "{{ 148387.77 | modulo:10 }}");

### LINKED ISSUE #369: Combining plus and minus filters lead to round/precision error in (some) decimals
### Dotliquid version
Any

### Expected behavior
No rounding/precision error - decimals are preserved unchanged

### Actual behavior
Precision introduces an error to some results

Use this new unit test case for repro:

        [Test]
        public void TestPlusCombinedWithMinus()
        {
            using (CultureHelper.SetCulture("en-GB"))
            {
                Helper.AssertTemplateResult("0.1", "{{ 0.1 | plus: 10 | minus: 10 }}");
            }
        }

Fails with 

Message:   Expected string length 3 but was 18. Strings differ at index 2.
  Expected: "0.1"
  But was:  "0.0999999999999996"
  -------------^

daviburg:
Closes #368 
Closes #369


Paul-Dempsey:
I didn't spot obvious issues with these changes. (Caveat: I'm not expert in this codebase, and I haven't done any testing - just a visual inspection).

src/DotLiquid/StandardFilters.cs
```diff
@@ -667,10 +667,29 @@ private static object DoMathsOperation(object input, object operand, Func<Expres
...
+                    operand = Convert.ToDecimal(operand);
+
+                    return ExpressionUtility
+                        .CreateExpression(
+                            body: operation,
+                            leftType: input.GetType(),
+                            rightType: operand.GetType())
+                        .DynamicInvoke(input, operand);
+                }
+                catch (Exception ex) when (ex is OverflowException || (ex is TargetInvocationException && ex?.InnerException is OverflowException))
```
tgjones, line 688: You’re catching OverflowExceptions caused by using Expression.*Checked here, but presumably double operations can now overflow below? I don’t think we should change that behaviour.

daviburg: Fair point. I'll see if I can both get the checked' overflow to be aware when Decimal range is exceeded while not changing the no-overflow current behavior for operations which go even beyond the range of float.

daviburg: Fair point. I'll see if I can both get the checked' overflow to be aware when Decimal range is exceeded while not changing the no-overflow current behavior for operations which go even beyond the range of float.

daviburg: Turns out double is not throwing overflow exception for checked operation, instead it gives infinity and negative infinity. Still researching and testing.

daviburg: That's right, no exception for double arithmetic anyway, as designed and documented: https://docs.microsoft.com/en-us/dotnet/api/system.double?view=netframework-4.8#Exceptions
I will be pushing unit tests demonstrating this anyway.

daviburg: There is a regression however for float and double division by zero. Decimal will throw for those while float and double will return either infinite values or NaN (not a number). I need to catch also the division by zero catching to fallback to double type just like the overflow so the existing float and double behavior is preserved.

maksimkim:
lgtm

rarayudu:
looks good.

Issue #390, opened 2020-06-13
Capitalize does not follow Shopify spec - capitalizes each word


daviburg:
### Dotliquid version
Current as of 6-13-2020

### Expected behavior
{{ "my great title" | capitalize }} outputs My great title
as per https://shopify.github.io/liquid/filters/capitalize/

### Actual behavior
{{ "my great title" | capitalize }} outputs My Great Title

### Steps to reproduce the Problem (you can add files)
            // Following test disabled due to out-of-spec bug
            //// Helper.AssertTemplateResult(
            ////     expected: "My great title",
            ////     template: "{{ 'my great title' | capitalize }}");

microalps:
CSS has a text transform named identically that behaves like DotLiquid. The spec is out of line with what users expect. Maybe we should add CapitalizeFirst which is explicit and in line with other filters like ReplaceFirst

daviburg:
Agree we need an approach that is non-breaking.
At the same time to facilitate re-use of Liquid templates from one implementation to the other, we need to have one implementation that is consistent with Shopify's (maybe with a flag 'StrictShopify').

microalps:
Does PR #420 fully resolve this issue?

daviburg:
> Does PR #420 fully resolve this issue?

yes

microalps:
Shopify is working on updating their docs (see https://github.com/Shopify/liquid/pull/1403 and the future docs at https://ericfromcanada.github.io/liquid/filters/capitalize/) that notes that capitalize actually lower cases everything. So mcDonald becomes Mcdonald! LiquidJS does not implement this way and is aligned with how we implemented in 2.1. However, as this does not match, this issue is being reopened.

I tested Ruby liquid on my server - 
```
{
  "template": "{{ name | capitalize }}",
  "data": {
    "name": "John G. Chalmers-Smith"
  }
}
```
Result: _John g. chalmers-smith_
if I add a space before the name, it will be all lowercase.

microalps:
I just found this (previously) undocumented behavior was discussed in https://github.com/dotliquid/dotliquid/pull/398#discussion_r496221817 shame I missed that when I wrote my patch.

PR #398, opened 2020-08-21
Updated "Capitalize" method to follow Shopify spec

cathalnoonan:
## Describe in your message, what you're fixing. Reference the original issue, if possible.
"Capitalize does not follow Shopify spec - capitalizes each word #390"

Rather than using the title-case approach, this code uses a regex match to find the first letter after full stops

The regex pattern used is `^\D|\.( ){1,}\D`
- `^\D` will match with the starting letter of the string
- `|` represents `OR`
- `\.( ){1,}\D` will match with the remaining letters that follow a full stop and some whitespace

For a visual representation of the regex pattern, open [this regexr link](https://regexr.com/5alh7)

## Describe how if any API-breaking changes your PR may introduce.

This code changes how dotliquid renders content; the current implementation of the 'Capitalize' tag may now be 'a feature rather than a bug'

For backwards compatibility, it might make more sense to leave the current implementation as is

There was a comment above the unit test of the updated code that indicates this bug is out of spec
> ``` csharp
> // Following test disabled due to out-of-spec bug see https://github.com/dotliquid/dotliquid/issues/390
> ```

## State whether it's ready-to-review/-merge or if it's a WIP PR.

***ready-to-review** - please read the API-breaking changes section before accepting*

The test method included in the pull request has been updated to reflect the expected behavior of Shopify (as outlined by #390)

The test passes using the test runner inside Visual Studio

image

### DIFF
--- a/src/DotLiquid.Tests/StandardFilterTests.cs
+++ b/src/DotLiquid.Tests/StandardFilterTests.cs
@@ -881,16 +881,17 @@ namespace DotLiquid.Tests
             Assert.AreEqual(null, StandardFilters.Capitalize(null));
             Assert.AreEqual("", StandardFilters.Capitalize(""));
             Assert.AreEqual(" ", StandardFilters.Capitalize(" "));
-            Assert.AreEqual("That Is One Sentence.", StandardFilters.Capitalize("That is one sentence."));
+            Assert.AreEqual("That is one sentence.", StandardFilters.Capitalize("That Is One Sentence."));
+            Assert.AreEqual("That is one sentence. This is another.", StandardFilters.Capitalize("That Is One Sentence. this Is Another."));
 
             Helper.AssertTemplateResult(
                 expected: "Title",
                 template: "{{ 'title' | capitalize }}");
 
             // Following test disabled due to out-of-spec bug see https://github.com/dotliquid/dotliquid/issues/390
-            //// Helper.AssertTemplateResult(
-            ////     expected: "My great title",
-            ////     template: "{{ 'my great title' | capitalize }}");
+            Helper.AssertTemplateResult(
+                expected: "My great title",
+                template: "{{ 'my great title' | capitalize }}");
         }
 
         [Test]
--- a/src/DotLiquid/StandardFilters.cs
+++ b/src/DotLiquid/StandardFilters.cs
@@ -109,22 +109,15 @@ namespace DotLiquid
         }
 
         /// <summary>
-        /// capitalize words in the input sentence
+        /// Capitalizes the first word in each sentence of <paramref name="input"/>
         /// </summary>
         /// <param name="input"></param>
         /// <returns></returns>
         public static string Capitalize(string input)
         {
-            if (input.IsNullOrWhiteSpace())
-                return input;
-
-            return string.IsNullOrEmpty(input)
+            return input.IsNullOrWhiteSpace()
                 ? input
-#if CORE
-                : Regex.Replace(input, @"\b(\w)", m => m.Value.ToUpper(), RegexOptions.None, Template.RegexTimeOut);
-#else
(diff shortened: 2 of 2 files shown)


src/DotLiquid/StandardFilters.cs
```diff
@@ -109,22 +109,15 @@ public static string UrlDecode(string input)
...
-
-            return string.IsNullOrEmpty(input)
+            return input.IsNullOrWhiteSpace()
                 ? input
-#if CORE
-                : Regex.Replace(input, @"\b(\w)", m => m.Value.ToUpper(), RegexOptions.None, Template.RegexTimeOut);
-#else
-                : CultureInfo.CurrentCulture.TextInfo.ToTitleCase(input);
-#endif
+                : Regex.Replace(input.ToLower(), @"^\D|\.( ){1,}\D", m => m.Value.ToUpper(), RegexOptions.None, Template.RegexTimeOut);
```
daviburg, line 120: I don't know that ToLower, i.e. everything else forced to lower case is the intent of the spec. The spec says "so later words are not affected"

The Shopify spec example does not include a capital letter in the input so it is difficult to say for sure.

My guess is that the intent is to leave other characters unchanged. If the input had a proper noun, setting all words in the sentence to lower case may not be desirable.

Either avoid ToLower or get a clarification from Shopify by asking their engineers what is the intended behavior (e.g. via their forum or their presence on twitter).

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -881,16 +881,17 @@ public void TestCapitalize()
             Assert.AreEqual(null, StandardFilters.Capitalize(null));
             Assert.AreEqual("", StandardFilters.Capitalize(""));
             Assert.AreEqual(" ", StandardFilters.Capitalize(" "));
-            Assert.AreEqual("That Is One Sentence.", StandardFilters.Capitalize("That is one sentence."));
+            Assert.AreEqual("That is one sentence.", StandardFilters.Capitalize("That Is One Sentence."));
```
daviburg, line 884: Best put such behavior change behind an opt-in flag so existing users are not surprised by the change. I filed the bug myself because the current behavior is out of spec. It has been so for a long time so it is probable some users now rely on it. Furthermore on Shopify's own forum there are customers wanting / expecting the behavior dot liquid has actually implemented, making it likely some would take a dependency on it.
Even Shopify's own wiki has contradictory statement in their wiki:
https://github.com/Shopify/liquid/wiki/Liquid-for-Designers#standard-filters

> capitalize - capitalize words in the input sentence
Note the unfortunate plural.

daviburg (changes requested):

src/DotLiquid/StandardFilters.cs
sebastienros: Looks like the shopify implementation is using String.capitalize, which does lower everything else. And also doesn't transform chars after a dot or an hyphen, only spaces.

https://github.com/Shopify/liquid/blob/a03f02789b14b0728e68c800c6699292a4c76672/lib/liquid/standardfilters.rb#L40

`"HellO. wOrld"` => `"Hello. world"`

https://repl.it/repls/MeanHumbleVolcano

daviburg: The code speaks, I guess. Shopify might not have documented by we can run their code to see what they implemented. Funny that the very line in their code is preceded by "capitalize words in the input centence" [sic]

microalps:
This still doesn't match the spec. The documentation does not say it supports sentence case,  just the first letter of a string (not sentence). I would create a separate filter, but it will not handle all the issues mentioned in https://stackoverflow.com/questions/3141426/net-method-to-convert-a-string-to-sentence-case#3141467

microalps:
@daviburg since #420 includes this as confirmed in #390 please close this PR.

Issue #330, opened 2018-11-29
Missing Standard filters: Ceil, Concat, EscapeOnce, Floor, SortNatural


mj-fox:
### Dotliquid version
2.0.180.0
### Expected behavior
The Liquid filters Ceil, Concat, EscapeOnce, Floor, SortNatural to be working in DotLiquid.
### Actual behavior
Filters not working because not implemented yet.

The Original Liquid's set of standard filters also include the above mentioned. Although most people probably don't need filters like Concat or SortNatural, Ceil and Floor are basic functions available in many template languages.

cdwiegand:
Submitted pull request to add ceil and floor filters: https://github.com/dotliquid/dotliquid/pull/399

microalps:
@daviburg  PR #399 was merged, please close

daviburg:
PR 399 is not enough. This issue also notes additional missing filters besides Ceil and Floor.

microalps:
Liquid 5.0 introduces echo and liquid tags. These are also a gap in DotLiquid now.

microalps:
As the only remaining filter has its own issue report, this can be closed.

PR #399, opened 2020-09-14
add Ceil and Floor filters and their tests

cdwiegand:
Added ceil and floor liquid filters, could have used them on a recent project. Issue is https://github.com/dotliquid/dotliquid/issues/330

### DIFF
--- a/src/DotLiquid.Tests/StandardFilterTests.cs
+++ b/src/DotLiquid.Tests/StandardFilterTests.cs
@@ -765,6 +765,32 @@ namespace DotLiquid.Tests
         }
 
         [Test]
+        public void TestCeil()
+        {
+            using (CultureHelper.SetCulture("en-GB"))
+            {
+                Helper.AssertTemplateResult("2", "{{ 1.234678 | ceil }}");
+                Helper.AssertTemplateResult("1", "{{ 1 | ceil }}");
+
+                Assert.Null(StandardFilters.Ceil(""));
+                Assert.Null(StandardFilters.Ceil("two"));
+            }
+        }
+
+        [Test]
+        public void TestFloor()
+        {
+            using (CultureHelper.SetCulture("en-GB"))
+            {
+                Helper.AssertTemplateResult("1", "{{ 1.234678 | floor }}");
+                Helper.AssertTemplateResult("1", "{{ 1 | floor }}");
+
+                Assert.Null(StandardFilters.Floor(""));
+                Assert.Null(StandardFilters.Floor("two"));
+            }
+        }
+
+        [Test]
         public void TestTimes()
         {
             using (CultureHelper.SetCulture("en-GB"))
--- a/src/DotLiquid.Tests/StandardFilterTests.cs
+++ b/src/DotLiquid.Tests/StandardFilterTests.cs
@@ -769,8 +769,10 @@ namespace DotLiquid.Tests
         {
             using (CultureHelper.SetCulture("en-GB"))
             {
-                Helper.AssertTemplateResult("2", "{{ 1.234678 | ceil }}");
-                Helper.AssertTemplateResult("1", "{{ 1 | ceil }}");
+                Helper.AssertTemplateResult("2", "{{ 1.2 | ceil }}");
+                Helper.AssertTemplateResult("2", "{{ 2.0 | ceil }}");
+                Helper.AssertTemplateResult("184", "{{ 183.357 | ceil }}");
+                Helper.AssertTemplateResult("4", "{{ \"3.5\" | ceil }}");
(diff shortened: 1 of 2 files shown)


src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -764,6 +764,32 @@ public void TestRound()
             }
         }
 
+        [Test]
+        public void TestCeil()
+        {
+            using (CultureHelper.SetCulture("en-GB"))
+            {
+                Helper.AssertTemplateResult("2", "{{ 1.234678 | ceil }}");
```
daviburg, line 772: Please cover the 4 spec examples at

https://shopify.github.io/liquid/filters/ceil/

Importantly do test the different input type i.e. number vs string containing a number.

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -764,6 +764,32 @@ public void TestRound()
...
+                Assert.Null(StandardFilters.Ceil("two"));
+            }
+        }
+
+        [Test]
+        public void TestFloor()
+        {
+            using (CultureHelper.SetCulture("en-GB"))
+            {
+                Helper.AssertTemplateResult("1", "{{ 1.234678 | floor }}");
```
daviburg, line 785: Same here, cover all spec normative examples:
https://shopify.github.io/liquid/filters/floor/

src/DotLiquid/StandardFilters.cs
```diff
@@ -650,6 +650,42 @@ public static object Round(object input, object places = null)
...
+        /// <summary>
+        /// Rounds a decimal value up to the next integer, unless already the integer value, removing all decimal places 
+        /// </summary>
+        /// <param name="input"></param>
+        /// <returns>The rounded value; null if an exception have occured</returns>
+        public static object Ceil(object input)
+        {
+            try
+            {
+                var i = Convert.ToDecimal(input);
```
daviburg, line 662: please inline single-use variable here and below

src/DotLiquid/StandardFilters.cs
```diff
@@ -650,6 +650,42 @@ public static object Round(object input, object places = null)
...
+                return Math.Ceiling(i);
+            }
+            catch (Exception)
```
daviburg, line 665: do not swallow all exceptions. target specific exception you need handling of, such as conversion exception, TBD if any math out of range exception would apply here at the edge. You don't want to swallow a memory exception or the like by accident here.

daviburg (changes requested):

src/DotLiquid/StandardFilters.cs
cdwiegand: Ok - I copied this from Round() and modified from there, so I will remove the try/catch as there is no math exception I would catch that should not bubble up to the calling program (i.e. you try to Ceil(Int64.MaxValue) or something really weird should probably fire an exception).

cdwiegand:
Updated code to reflect recommendations, tests pass with standard values from the original liquid website's own examples. Cleaned up code re: variables, also using trycatch to handle odd values (like the word "two" not being valid).

src/DotLiquid/StandardFilters.cs
```diff
@@ -650,6 +650,32 @@ public static object Round(object input, object places = null)
             }
         }
 
+        /// <summary>
+        /// Rounds a decimal value up to the next integer, unless already the integer value, removing all decimal places 
+        /// </summary>
+        /// <param name="input"></param>
+        /// <returns>The rounded value; null if an exception have occured</returns>
```
daviburg, line 657: Nit nit: occurred (two 'r')

daviburg: oh and, an exception has occurred (not have)

daviburg (approved):

src/DotLiquid/StandardFilters.cs
daviburg: Yes for those specific case you want to catch the specific exception
Console.WriteLine(Math.Ceiling(Decimal.MaxValue)); however does not throw
https://dotnetfiddle.net/MsJ4Kq

daviburg: Round(...)'s precedence... not great :-) should fix it if we have the time, it will be tricky to figure out all exceptions it used to catch to weed the good from the bad.

src/DotLiquid/StandardFilters.cs
daviburg: Merging as-is as no response. Consider a supplemental PR for the tweaks.

PR #401, opened 2020-09-28
Fix of some cast problems between epoch date, plus method and back to date

sstraus:
The main porpoise of this fix is to correctly handle a common pattern in Liquid date managements

The following code will explain what I'm trying to accomplish, add a day to a date.

```
Template template = Template.Parse(@"Add one day to current date: {% assign seconds = 14 | times: 24 | times: 60 | times: 60 %}{{mydate | date: ""%s"" | plus: seconds | date: ""%a, %b %d, %y"" }}"); 
Console.WriteLine(template.Render(Hash.FromAnonymousObject(new { mydate = DateTime.Now })));
```

### DIFF
--- a/src/DotLiquid.Tests/StandardFilterTests.cs
+++ b/src/DotLiquid.Tests/StandardFilterTests.cs
@@ -715,6 +715,9 @@ namespace DotLiquid.Tests
                 Helper.AssertTemplateResult("5.5", "{{ 2  | plus:3.5 }}");
                 Helper.AssertTemplateResult("5.5", "{{ 3.5 | plus:2 }}");
                 Helper.AssertTemplateResult("11", "{{ '1' | plus:'1' }}");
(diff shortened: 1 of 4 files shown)


daviburg:
This is a breaking change not supported by Shopify specification of behavior one way or the other. The date filter in Shopify spec is about formatting, it does not say anything about math operations on the output, neither does it explicitly declare the output type.

sstraus:
> The date filter in Shopify spec is about formatting, it does not say anything about math operations on the output

@daviburg I agree with you but partially, Liquid is born on ruby, that isn’t a typed language. Using the date as epoc to add or remove days from dates is common, as you can see in many samples provided to the community in their forum. https://community.shopify.com/c/Shopify-Design/Simple-Date-Math-Manipulation-Help-Needed-for-Liquid/td-p/204393.

PR #403, opened 2020-10-23
Created IRenderableFactory and ability to pass it in to the Template.Parse

aes2devgroup:
The company I work with uses this package heavily, and we recently needed the functionality to modify (at our discretion) the output of a variable in a custom way, that wouldn't require editing the source template files.

This leaves default functionality intact without any discernible performance impact. The only thing that the default behavior does now that it didn't before is the `if (tag is Block block)` check in `Block.cs`, so that it can pass the IRendererFactory to sub blocks.

I added in a couple unit tests to show a simple example of what we needed to accomplish.

This is ready to review and merge if accepted. 

There are no breaking changes, the parameter for the factory on the `Template.Parse` method is optional, and if not provided will work exactly as before.

Due to our reliance on this package, if you'd like, my company has shown interest in becoming a maintainer for DotLiquid, and could help in upgrading it to .netStandard2.0 and/or core.

### DIFF
--- a/src/DotLiquid/Template.cs
+++ b/src/DotLiquid/Template.cs
@@ -212,10 +212,10 @@ namespace DotLiquid
         /// </summary>
         /// <param name="source"></param>
         /// <returns></returns>
-        public static Template Parse(string source)
+        public static Template Parse(string source, IRenderableFactory renderableFactory = null)
         {
             Template template = new Template();
-            template.ParseInternal(source);
+            template.ParseInternal(source, renderableFactory);
             return template;
         }
 
@@ -275,12 +275,12 @@ namespace DotLiquid
         /// </summary>
         /// <param name="source">The source code.</param>
         /// <returns>The template.</returns>
-        internal Template ParseInternal(string source)
+        internal Template ParseInternal(string source, IRenderableFactory renderableFactory = null)
         {
             source = DotLiquid.Tags.Literal.FromShortHand(source);
             source = DotLiquid.Tags.Comment.FromShortHand(source);
 
-            this.Root = new Document();
+            this.Root = new Document(renderableFactory);
             this.Root.Initialize(tagName: null, markup: null, tokens: Template.Tokenize(source));
             return this;
         }
--- a/src/DotLiquid.Tests/RenderableFactoryTests.cs
+++ b/src/DotLiquid.Tests/RenderableFactoryTests.cs
@@ -1,0 +1,89 @@
+using NUnit.Framework;
+using System.IO;
+
+namespace DotLiquid.Tests
+{
+    [TestFixture]
+    public class RenderableFactoryTests
+    {
+
+        [Test]
+        public void TestDefaultVariableFactory()
+        {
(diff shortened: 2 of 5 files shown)


src/DotLiquid/Template.cs
```diff
@@ -212,10 +212,10 @@ public static void RegisterValueTypeTransformer(Type type, Func<object, object>
         /// </summary>
         /// <param name="source"></param>
         /// <returns></returns>
-        public static Template Parse(string source)
+        public static Template Parse(string source, IRenderableFactory renderableFactory = null)
```
daviburg, line 215: This runs contrary to the standing pattern for this class. Configuration / customization is not done via additional parameters but why global registration. We should remains consistent with the existing pattern unless we have specific need not to.

daviburg (changes requested):

daviburg:
Probably this feature and how to approach it should be discuss via gitter for dotliquid first to get a sense how other current users would like to see added to the current pattern.
https://gitter.im/dotliquid/dotliquid?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge

microalps:
We'd appreciate additional maintainers but I agree with @daviburg on all his points. Please create an issue to discuss and/or hop on gitter. A better description of the use case would garner more attention.

microalps:
WHEREAS 
(a) there has been no developer input, and 
(b) in seeing that Liquid reference library has now implemented an 'echo' tag (https://ericfromcanada.github.io/liquid/tags/template/), and 
(c) this PR has no similar functionality in the reference library, and
(d) the entire concept of this tag is to change the behavior of a standard syntax from what's documented and from what designers expect
(e) the same thing can be achieved via Tags (below)
I don't think this is in the best interest of the platform to implement.

Here is a sample tag that will do the "More Complex" example
```c#
public class MoreComplexTestVariable : Tag
{
	private Variable variable;

	public override void Initialize(string tagName, string markup, List<string> tokens)
	{
		variable = new Variable(markup);
		base.Initialize(tagName, markup, tokens);
	}

	public override void Render(Context context, TextWriter result)
	{
		var variableNumber = 1;
		if (context.Environments[0].ContainsKey("__variableIndex"))
			variableNumber = (int)context.Environments[0]["__variableIndex"];
		context.Environments[0]["__variableIndex"] = variableNumber + 1;
		
		result.Write($"<span data-render-num=\"{variableNumber}\" data-original-variable-name=\"{variable.Name}\">");
		base.Render(context, result);
		result.Write("</span>");
	}
}
```

And the template would be (after registering with `Template.RegisterTag<MoreComplexTestVariable>("intvar");`)
```liquid
{% intvar test %}{% intvar test2 %}{%intvar test3%}
```

You can even replace {{ }} in your template before parsing if you so chose, but it would NOT be expected behavior by a designer and thus not something the library should do natively.

@daviburg Is it OK to close this PR?

daviburg:
Agreed, closing this PR. Please consider the feedback above for future new take on this ask.

PR #406, opened 2020-12-09
Add an interface to create Context with CancellationToken

BoyaWu10:
Add an interface to create Context with CancellationToken

### DIFF
--- a/src/DotLiquid.Tests/ExceptionHandlingTests.cs
+++ b/src/DotLiquid.Tests/ExceptionHandlingTests.cs
@@ -1,4 +1,6 @@
+using System.Collections.Generic;
 using System.Globalization;
+using System.Threading;
 using DotLiquid.Exceptions;
 using NUnit.Framework;
 
@@ -100,6 +102,19 @@ namespace DotLiquid.Tests
                 {
                     Timeout = 100 //ms
                 });
+            });
+        }
+
+        [Test]
+        public void TestOperationCancelledError()
+        {
+            var template = Template.Parse(" {% for i in (1..1000000) %} {{ i }} {% endfor %} ");
+            var source = new CancellationTokenSource(100);
+            var context = new Context(new List<Hash>(), new Hash(), new Hash(), ErrorsOutputMode.Rethrow, 0, source.Token, CultureInfo.InvariantCulture);
+
+            Assert.Throws<System.OperationCanceledException>(() =>
+            {
+                template.Render(RenderParameters.FromContext(context, CultureInfo.InvariantCulture));
             });
         }
 
--- a/src/DotLiquid.Tests/ExceptionHandlingTests.cs
+++ b/src/DotLiquid.Tests/ExceptionHandlingTests.cs
@@ -110,7 +110,14 @@ namespace DotLiquid.Tests
         {
             var template = Template.Parse(" {% for i in (1..1000000) %} {{ i }} {% endfor %} ");
             var source = new CancellationTokenSource(100);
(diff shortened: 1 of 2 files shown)


BoyaWu10:
@daviburg Could you please help to review this PR?

src/DotLiquid.Tests/ExceptionHandlingTests.cs
```diff
@@ -103,6 +105,19 @@ public void TestTimeoutError()
             });
         }
 
+        [Test]
+        public void TestOperationCancelledError()
+        {
+            var template = Template.Parse(" {% for i in (1..1000000) %} {{ i }} {% endfor %} ");
+            var source = new CancellationTokenSource(100);
+            var context = new Context(new List<Hash>(), new Hash(), new Hash(), ErrorsOutputMode.Rethrow, 0, source.Token, CultureInfo.InvariantCulture);
```
daviburg, line 113: I recommend to use named parameters when using constant input values such as 0 or null so it is easier to understand for human readers.

src/DotLiquid/Context.cs
```diff
@@ -91,6 +93,39 @@ public Context
...
+        /// <param name="outerScope"></param>
+        /// <param name="registers"></param>
+        /// <param name="errorsOutputMode"></param>
+        public Context
+            (List<Hash> environments
+             , Hash outerScope
+             , Hash registers
+             , ErrorsOutputMode errorsOutputMode
+             , int maxIterations
+             , CancellationToken cancellationToken
```
daviburg, line 109: Why not add at the end?

src/DotLiquid/Context.cs
```diff
@@ -91,6 +93,39 @@ public Context
             SquashInstanceAssignsWithEnvironments();
         }
 
+        /// <summary>
+        /// Creates a new rendering context
+        /// </summary>
+        /// <param name="environments"></param>
+        /// <param name="outerScope"></param>
+        /// <param name="registers"></param>
+        /// <param name="errorsOutputMode"></param>
```
daviburg, line 102: Add the XML comment for the new parameter and take the opportunity to fill in the missing comments content for the existing parameters.

src/DotLiquid/Context.cs
```diff
@@ -91,6 +93,39 @@ public Context
             SquashInstanceAssignsWithEnvironments();
```
daviburg, line 93: Avoid code duplication by having one of the two methods call the other one instead.

daviburg (changes requested):

daviburg:
Please add a feature work item to the backlog of this and link it to this PR

BoyaWu10:
Thanks for the comments. They are fixed in the update. Do we track feature work items in GitHub issues? If so, I've created an issue #407 for this PR.

daviburg:
> Thanks for the comments. They are fixed in the update. Do we track feature work items in GitHub issues? If so, I've created an issue #407 for this PR.

As we have multiple companies wanting to add various features to DotLiquid lately, we should be a bit more formal with the changes, so yes, we should have a item for each change. I'm not a GitHub expert, but it may well be that all items are 'issues' even feature requests.

src/DotLiquid/Context.cs
daviburg: The XML param declaration has been added but the comment is still empty. Each comment should be something like:

<param name="errorsOutputMode">Errors output mode</param>

(Feel free to expand the descriptions based on your knowledge of what the parameter does).

daviburg (approved):

Issue #409, opened 2021-01-11
Contains not working


TheTrigger:
### Dotliquid version
2.0.366

### Expected behavior
Returns `OK`

### Actual behavior
Returns `aaabbbccc`

### Steps to reproduce the Problem
```csharp
[Fact]
public void TestContains()
{
    var html = "{% if array contains 'ccc' %}OK{% else %}{{ array }}{% endif %}";
    var array = new string[] { "aaa", "bbb", "ccc" };
    
    var template = DotLiquid.Template.Parse(html);
    var result = template.Render(Hash.FromAnonymousObject(new
    {
        array,
    }));
    
    Assert.Equal("OK", result);
}
```

microalps:
@daviburg This seems to be related to PR #378 . Can you comment? BackCompatSafeTypeInsensitiveEqual prohibits comparing string to string. There are no test cases for this (and I didn't go back in time to see if this worked pre- #362

daviburg:
What are the types passed to compare for this case? It appears that the types should be two strings, but evidently aren't as the comparison fails. Why aren't these types strings in the first place?

microalps:
Re: #409  Here is some debug output I generated for int and string test cases.
```
Testing array contains 6 == false
Value: 1 (System.Int32) String: False !Char: True Other: 6 (System.Int32) String: False !Char: True
Value: 2 (System.Int32) String: False !Char: True Other: 6 (System.Int32) String: False !Char: True
Value: 3 (System.Int32) String: False !Char: True Other: 6 (System.Int32) String: False !Char: True
Value: 4 (System.Int32) String: False !Char: True Other: 6 (System.Int32) String: False !Char: True
Value: 5 (System.Int32) String: False !Char: True Other: 6 (System.Int32) String: False !Char: True
Testing array contains '1' == false
Value: 1 (System.Int32) String: False !Char: True Other: 1 (System.String) String: True !Char: True
Value: 2 (System.Int32) String: False !Char: True Other: 1 (System.String) String: True !Char: True
Value: 3 (System.Int32) String: False !Char: True Other: 1 (System.String) String: True !Char: True
Value: 4 (System.Int32) String: False !Char: True Other: 1 (System.String) String: True !Char: True
Value: 5 (System.Int32) String: False !Char: True Other: 1 (System.String) String: True !Char: True
```
vs strings
```
Testing array contains 'Banana' == true
[...]

PR #416, opened 2021-02-02
Fix contains condition support for string arrays

microalps:
Resolves issue #409

### DIFF
--- a/src/DotLiquid.Tests/ConditionTests.cs
+++ b/src/DotLiquid.Tests/ConditionTests.cs
@@ -176,7 +176,10 @@ namespace DotLiquid.Tests
 
             AssertEvaluatesTrue(left: "array", op: "contains", right: "'Apple'");
             AssertEvaluatesTrue(left: "array", op: "startsWith", right: "first");
+            AssertEvaluatesTrue(left: "array.first", op: "==", right: "first");
             AssertEvaluatesFalse(left: "array", op: "contains", right: "'apple'");
+            AssertEvaluatesFalse(left: "array", op: "startsWith", right: "'apple'");
+            AssertEvaluatesFalse(left: "array.first", op: "==", right: "'apple'");
             AssertEvaluatesFalse(left: "array", op: "contains", right: "'Mango'");
             AssertEvaluatesTrue(left: "array", op: "contains", right: "'Orange'");
             AssertEvaluatesTrue(left: "array", op: "contains", right: "'Banana'");
@@ -197,6 +200,7 @@ namespace DotLiquid.Tests
 
             AssertEvaluatesTrue(left: "array", op: "contains", right: "first");
             AssertEvaluatesTrue(left: "array", op: "startsWith", right: "first");
+            AssertEvaluatesTrue(left: "array.first", op: "==", right: "first");
             AssertEvaluatesTrue(left: "array", op: "contains", right: "clone");
             AssertEvaluatesTrue(left: "array", op: "startsWith", right: "clone");
             AssertEvaluatesTrue(left: "array", op: "endsWith", right: "last");
@@ -213,6 +217,7 @@ namespace DotLiquid.Tests
 
             AssertEvaluatesTrue(left: "array", op: "contains", right: "first");
             AssertEvaluatesTrue(left: "array", op: "startsWith", right: "first");
+            AssertEvaluatesTrue(left: "array.first", op: "==", right: "'true'");
             AssertEvaluatesTrue(left: "array", op: "startsWith", right: "'true'");
 
             AssertEvaluatesFalse(left: "array", op: "contains", right: "'true'"); // to be re-evaluated in #362
@@ -230,6 +235,7 @@ namespace DotLiquid.Tests
             AssertEvaluatesTrue(left: "array", op: "contains", right: "'A'");
             AssertEvaluatesTrue(left: "array", op: "contains", right: "first");
             AssertEvaluatesTrue(left: "array", op: "startsWith", right: "first");
+            AssertEvaluatesTrue(left: "array.first", op: "==", right: "first");
             AssertEvaluatesFalse(left: "array", op: "contains", right: "'a'");
             AssertEvaluatesFalse(left: "array", op: "contains", right: "'X'");
             AssertEvaluatesTrue(left: "array", op: "contains", right: "'B'");
@@ -249,6 +255,7 @@ namespace DotLiquid.Tests
             AssertEvaluatesFalse(left: "array", op: "contains", right: "0");
             AssertEvaluatesFalse(left: "array", op: "contains", right: "'0'");
             AssertEvaluatesTrue(left: "array", op: "startsWith", right: "first");
+            AssertEvaluatesTrue(left: "array.first", op: "==", right: "first");
             AssertEvaluatesTrue(left: "array", op: "contains", right: "first");
             AssertEvaluatesFalse(left: "array", op: "contains", right: "1");
             AssertEvaluatesTrue(left: "array", op: "endsWith", right: "last");
--- a/src/DotLiquid.Tests/ConditionTests.cs
+++ b/src/DotLiquid.Tests/ConditionTests.cs
@@ -1,4 +1,6 @@
+using System.Collections.Generic;
 using System.Globalization;
+using System.Linq;
 using System.Text.RegularExpressions;
 using DotLiquid.Exceptions;
 using DotLiquid.NamingConventions;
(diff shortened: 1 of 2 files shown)


src/DotLiquid/Util/ObjectExtensionMethods.cs
```diff
@@ -52,8 +54,10 @@ public static bool BackCompatSafeTypeInsensitiveEqual(this object value, object
         {
             // NOTE(daviburg): Historically testing for equality cross integer and string boundaries resulted in not equal.
             // This ensures we preserve the behavior.
-            if (value is string && !(otherValue is char) || otherValue is string && !(value is char))
-            {
+            var aType = value.GetType();
+            var bType = otherValue.GetType();
+            if (aType != bType && (_numericTypes.Contains(aType) ^ _numericTypes.Contains(bType)))
```
daviburg, line 59: If we do that let's stick to logical operators with || rather than ^

src/DotLiquid/Util/ObjectExtensionMethods.cs
```diff
@@ -52,8 +54,10 @@ public static bool BackCompatSafeTypeInsensitiveEqual(this object value, object
         {
             // NOTE(daviburg): Historically testing for equality cross integer and string boundaries resulted in not equal.
             // This ensures we preserve the behavior.
-            if (value is string && !(otherValue is char) || otherValue is string && !(value is char))
```
daviburg, line 55: Could it be this condition was simply missing the condition where both types are equal, strings in this case?

daviburg (changes requested):

src/DotLiquid/Util/ObjectExtensionMethods.cs
microalps: This is shorter. Either both are numeric or both are non-numeric. XOR logical operator. https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/boolean-logical-operators

src/DotLiquid/Util/ObjectExtensionMethods.cs
microalps: No. My test passes without type equality as well, just shorter to test that first. 
```
                if ((_numericTypes.Contains(aType) ^ _numericTypes.Contains(bType)))
                {
                    return false;
                }
```
All conditional tests pass.

src/DotLiquid/Util/ObjectExtensionMethods.cs
```diff
@@ -50,11 +52,16 @@ public static object Send(this object value, string member, object[] parameters
...
-            // This ensures we preserve the behavior.
-            if (value is string && !(otherValue is char) || otherValue is string && !(value is char))
+            if (value != null && otherValue != null)
             {
-                return false;
+                // NOTE(daviburg): Historically testing for equality cross integer and string boundaries resulted in not equal.
+                // This ensures we preserve the behavior.
+                var aType = value.GetType();
+                var bType = otherValue.GetType();
+                if (aType != bType && (_numericTypes.Contains(aType) != _numericTypes.Contains(bType)))
```
daviburg, line 61: This is a significant departure from the earlier backward compatibility scope. The issue you have raised appears to be specific to testing contains in arrays for strings. But the code you wrote is a generic broad reaching type comparison for anything but 4 numeric types (which are not even all the numeric types in .NET).

Please scope the fix to the issue reported or,
- expand both the reported issue 
- expand validation tests to additional types
- learn historical behavior through testing

Otherwise we may just be creating yet more bugs to fix in the future.

microalps: I went back to pre-362 and conducted several tests.
1. This fails: int testing - `AssertEvaluatesTrue("array", "contains", "1.0");` This is because 1.0 is not int and IList.Contains must have equal types.
2. String testing - all my tests are successful (not so on current tip)
3- Char testing - not supported (have no preference to this, but added because your code referred to char types). 'A' evaluates to string and IList.Contains must have equal types.

So technically speaking entire premise of 362 breaks backward compatibility. HOWEVER I understand fixing point 1 is a FEATURE and is GOOD. My code continues in that vain by defining numeric types that are compatible instead of testing char vs string to determine if it's numeric (which broke point 2)

microalps: Let's discuss this elsewhere. I posted on gitter but happy to take this conversation to whatever is convenient.

daviburg: Yes, thank you for the investigation and will look up the conversation on gitter.

daviburg: Suggesting one-liner fix here:
if (value.GetType() != otherValue.GetType() && (value is string && !(otherValue is char) || otherValue is string && !(value is char)))
Simpler and better perf as we don't need to iterate through static arrays of types.

microalps: As we discussed on gitter, this one-liner has a slew of issues. In the effort to maintain consistency with pre-362 behavior, all tests added to this commit are also added to https://github.com/microalps/dotliquid/tree/pre-362-tests for comparison.

src/DotLiquid/Util/ObjectExtensionMethods.cs
```diff
@@ -1,11 +1,18 @@
...
 using System.Reflection;
 
 namespace DotLiquid.Util
 {
     public static class ObjectExtensionMethods
     {
+        private static HashSet<HashSet<Type>> _BackCompatComparableTypeBoundaries = new HashSet<HashSet<Type>>() {
+            new HashSet<Type> { typeof(decimal), typeof(double), typeof(float), typeof(int), typeof(uint), typeof(long), typeof(ulong), typeof(short), typeof(ushort) },
+            new HashSet<Type> { typeof(string), typeof(char),
+    }
```
microalps, line 13: noted extra comma and linebreak

src/DotLiquid.Tests/ConditionTests.cs
```diff
@@ -128,6 +165,112 @@ public void TestContainsWorksOnLongArrays()
...
+            var _array = new List<byte> { 0x01, 0x02, 0x03, 0x30 };
+            _context["array"] = _array.ToArray();
+            _context["first"] = _array.First();
+            _context["last"] = _array.Last();
+
+            AssertEvaluatesFalse(left: "array", op: "contains", right: "0");
+            AssertEvaluatesFalse(left: "array", op: "contains", right: "'0'");
+            AssertEvaluatesTrue(left: "array", op: "startsWith", right: "first");
+            AssertEvaluatesTrue(left: "array", op: "contains", right: "first");
+            AssertEvaluatesFalse(left: "array", op: "contains", right: "1");
```
daviburg, line 253: Was it possible to get a match with a byte at all? Is there a way to write the right operand as a byte?

src/DotLiquid.Tests/ConditionTests.cs
```diff
@@ -9,6 +11,40 @@ namespace DotLiquid.Tests
...
+
+                if (other is string @string)
+                    return Equals(@string);
+
+                return false;
+            }
+
+            public bool Equals(Car other)
+            {
+                return other.Make == this.Make && other.Model == this.Model;
```
daviburg, line 38:
```
      if (other == null)
         return false;

```
here and below

daviburg (approved):

src/DotLiquid.Tests/ConditionTests.cs
```diff
@@ -128,6 +165,112 @@ public void TestContainsWorksOnLongArrays()
...
+        public void TestStringArrays()
+        {
+            _context = new Context(CultureInfo.InvariantCulture);
+            var _array = new List<string>() { "Apple", "Orange", null, "Banana" };
+            _context["array"] = _array.ToArray();
+            _context["first"] = _array.First();
+            _context["last"] = _array.Last();
+
+            AssertEvaluatesTrue(left: "array", op: "contains", right: "'Apple'");
+            AssertEvaluatesTrue(left: "array", op: "startsWith", right: "first");
```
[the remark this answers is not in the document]
microalps: I added the first test in each array test where I had startsWith. Successful.

src/DotLiquid.Tests/ConditionTests.cs
```diff
@@ -128,6 +165,112 @@ public void TestContainsWorksOnLongArrays()
...
+        [Test]
+        public void TestStringArrays()
+        {
+            _context = new Context(CultureInfo.InvariantCulture);
+            var _array = new List<string>() { "Apple", "Orange", null, "Banana" };
+            _context["array"] = _array.ToArray();
+            _context["first"] = _array.First();
+            _context["last"] = _array.Last();
+
+            AssertEvaluatesTrue(left: "array", op: "contains", right: "'Apple'");
```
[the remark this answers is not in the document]
microalps: I think testing first/last contexts prove that sufficiently. That would only prove variable = string, which many other tests are proving (including contains). The eval of context variables is evaluated equally between contains and startsWith.

PR #418, opened 2021-02-05
C# Naming Convention

microalps:
This addresses several bugs with the C# naming convention:
- `{% if Array.Size > 0 %} ` did not evaluate - it expects size regardless of convention. `{{ Array | Size }}` does respect convention.
- C# devs use PascalCase and camelCase. The latter was supposed to be allowed, but was uppercasing the C instead of lower-casing it resulting in PascalCase being allowed twice. Reason this wasn't caught is there is a third convention allowed of all lowercase (at least conditionals) which most tests were using.
- Testing naming convention requires a change to the static property. While many tests correctly reverted, some used older syntax that did not. I followed previous practice in updating what I hope is all of them.

This introduced two tests that were failing - after review the tests themselves are incorrectly defining the expected behavior. The above bug fixed revealed these and my rational for changing the tests are noted below.

### DIFF
--- a/src/DotLiquid.Tests/StandardFilterTests.cs
+++ b/src/DotLiquid.Tests/StandardFilterTests.cs
@@ -598,8 +598,25 @@ namespace DotLiquid.Tests
         }
 
         [Test]
-        public void TestFirstLast()
+        public void TestFirstLastUsingRuby()
         {
+            var namingConvention = new NamingConventions.RubyNamingConvention();
+            TestFirstLast(namingConvention, (name) => namingConvention.GetMemberName(name));
+        }
+
+        [Test]
+        public void TestFirstLastUsingCSharp()
+        {
+            var namingConvention = new NamingConventions.CSharpNamingConvention();
+            TestFirstLast(namingConvention, (name) => char.ToUpperInvariant(name[0]) + name.Substring(1));
+        }
+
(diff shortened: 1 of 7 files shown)


daviburg (approved):

microalps:
I just noticed part of this commit was originally proposed as part of #274. Kudos for catching this first. See https://github.com/dotliquid/dotliquid/pull/274/commits/85bf389e685694a85bb877b35f9250721b251ebc

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -609,29 +626,37 @@ public void TestFirstLast()
 
             Helper.AssertTemplateResult(
                 expected: ".",
-                template: "{{ 'Ground control to Major Tom.' | last }}");
+                template: "{{ 'Ground control to Major Tom.' | " + lastFilter + " }}",
+                namingConvention: namingConvention);
             Helper.AssertTemplateResult(
                 expected: "Tom.",
-                template: "{{ 'Ground control to Major Tom.' | split: ' ' | last }}");
+                template: "{{ 'Ground control to Major Tom.' | "+ splitFilter + ": ' ' | " + lastFilter + " }}",
```
daviburg, line 633: nit nit: missing space character before first + sign

daviburg (approved):

PR #420, opened 2021-02-21
Strict Liquid Syntax (Backwards-Compatible)

### LINKED ISSUE #149: Losing offset when using Date filter on a DateTimeOffset object
I have a DateTimeOffset object with +4 hours offset but when using the date filter, it always loses this offset. This is due to parsing the input just after you convert it to a string (input.ToString()) on the date filter. 

I have worked around this by introducing my own filter that uses DateTimeOffset. I am more than happy to fix this and issue a pull request but i thought I would this here to see if you already made plans to fix it or not.

Thanks.

### LINKED ISSUE #214: Is there any way to do date arithmetic
Say I want to subtract a few days to the date, in liquid we can do something like:

`my_date |  date: "%s" | minus : 604800`

is there an equivalent or workaround in dotliquid?

### LINKED ISSUE #309: DateTimeOffset not supported?
### Dotliquid version
2.0.266

### Expected behavior
Some sort of output.

### Actual behavior
Empty string

```
        [Fact]
        public void CanRenderPoco()
        {
            var userCtx = new UserContext { Name = "Oliver", CreatedAt = new DateTimeOffset(2018, 7, 4, 15, 6, 41, TimeSpan.FromHours(10))};
            var t = Template.Parse("Hi {{name}}! You were created at {{createdAt}}");

            var result = t.Render(Hash.FromAnonymousObject(userCtx));

            Assert.Equal("Hi Oliver! You were created at xyz", result);
        }
```

```
Xunit.Sdk.EqualException
Assert.Equal() Failure
                                 ↓ (pos 31)
Expected: ···You were created at xyz
Actual:   ···You were created at 

```

### LINKED ISSUE #316: Util.ToStrFTime returns always String
### Dotliquid version
2.0.279
### Expected behavior
Output: August 20, 2018

Liquid template
```
{% assign myDate = "2018-08-17" %}
{{ myDate | date :"%s" | plus: 259200 | date :"%B %d, %Y"}}
```
### Actual behavior
Output: 1534464000259200

I would like to add some days to a date. I convert the date to seconds and then I add the days (converted in second) using the plus filter. This is the only way to add some days to a date using standard liquid filters.

The issue is with the Util.ToStrFTime, it returns a String and so the Plus function Concatenate the two strings.

```
return input is string
                ? string.Concat(input, operand)
                : DoMathsOperation(input, operand, Expression.Add);
```

What about changing the Util.ToStrFTime method return type? (i.e. return an Obejct).

Thanks

### LINKED ISSUE #322: date filter is not handling epoch integer
### Dotliquid version
```
<package id="DotLiquid" version="2.0.298" targetFramework="net461" />
```

### Expected behavior
**Input:**
```
{'payload': {'state': {'desired': {'enabled': True}, 'reported': {'telemetry': {'temperature': 4, 'humidity': 22, 'occupancy': 82}, 'version': '1.0.0', 'enabled': True}}, 'metadata': {'desired': {'enabled': {'timestamp': 1539200675}}, 'reported': {'telemetry': {'temperature': {'timestamp': 1539288920}, 'humidity': {'timestamp': 1539288800}, 'occupancy': {'timestamp': 1539288680}}, 'version': {'timestamp': 1539200675}, 'enabled': {'timestamp': 1539200675}}}, 'version': 4210}, 'snapshot': True, 'snapshot_count': 10, 'thing': 'Thermostat1', 'metadata': {'stream': 'Thermostat1Stream'}}
```
**Template:**
```
"timestamp": "{{ payload.metadata.reported.version.timestamp | date: "%Y/%m/%d %H:%M:%S" }}",
```

**Output:**
```
"timestamp": "2018/10/10 19:44:35",
```

### Actual behavior
**Output:**
```
"timestamp": "1539288680",
```

**Code**
```
namespace dotliquidapp
{
    class Program
    {
        static void Main(string[] args)
        {
            var rawTemplate = Properties.Settings.Default.Template;
            var inputJson = Properties.Settings.Default.Input;
            var inputDict = JsonConvert.DeserializeObject<Dictionary<string, object>>(inputJson, new JsonConverter[] { new MyConverter() });
            var template = Template.Parse(rawTemplate);

            var output = template.Render(Hash.FromDictionary(inputDict));
            Console.Out.WriteLine("------------Output--------------");
            Console.Out.WriteLine(output);
            Console.ReadLine();
        }
    }

    class MyConverter : CustomCreationConverter<IDictionary<string, object>>
    {
        public override IDictionary<string, object> Create(Type objectType)
        {
            return new Dictionary<string, object>();
        }

        public override bool CanConvert(Type objectType)
        {
            // in addition to handling IDictionary<string, object>
            // we want to handle the deserialization of dict value
            // which is of type object
            return objectType == typeof(object) || base.CanConvert(objectType);
        }

        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
        {
            if (reader.TokenType == JsonToken.StartObject
                || reader.TokenType == JsonToken.Null)
                return base.ReadJson(reader, objectType, existingValue, serializer);

            // if the next token is not an object
            // then fall back on standard deserializer (strings, numbers etc.)
            return serializer.Deserialize(reader);
        }
    }
}
```

### LINKED ISSUE #328: Date with minus filter. Parameter count mismatch
### Dotliquid version
2.0.298

### Expected behavior
C# Code
`
DotLiquid.Template template = DotLiquid.Template.Parse(email.Body);
Template.NamingConvention = new DotLiquid.NamingConventions.RubyNamingConvention();
var htmlContent = template.Render(DotLiquid.Hash.FromAnonymousObject(emailData));
`
HTML
` {% assign date_start = activity.created_on | date: "%s" %} ` returns something like 1542645650
`{% assign date_start = activity.created_on | date: "%s" | minus : 604800 %}` return 
1542645650 - 604800
### Actual behavior
` {% assign date_start = activity.created_on | date: "%s" %} ` returns something like 1542645650
`{% assign date_start = activity.created_on | date: "%s" | minus : 604800 %}` Gives me "Liquid error: Parameter count mismatch."

### LINKED ISSUE #385: Replace standard filter uses RegEx pattern matching instead of Shopify intended simple string matching
### Dotliquid version
Any

### Expected behavior
Match Shopify intend, use an extension / additional filter for regex replace
Intend is somewhat vague on Shopify spec which only says matching a string (although the absence of calling out escapes and pattern suggest they are not supported or expected). But they have an issue discussed in which the intend is made clear:
https://github.com/Shopify/liquid/issues/202

### Actual behavior
Standard filter implementation does a RegEx pattern matching. There is no additional filter to distinguish simple string matching from pattern matching

`{ "Date": "{{ 'MMM "EEE - SS BNBN KLJLsample\\' | replace: '\\', '\\' | replace: '"', '\"'}}"}` will generate
`{ "Date": "MMM \"EEE - SS BNBN KLJLsample\\\\"}`

`{ "Date": "{{ 'MMM "EEE - SS BNBN KLJLsample\\' | replace: '\', '\\' | replace: '"', '\"'}}"}` will fail with
`{ "Date": "Liquid error: parsing "\" - Illegal \ at end of pattern."}`

To match Shopify's behavior the second template should generate the first output.

### Discussion
This will be troublesome to fix as clients of dotLiquid are likely to have started relying on the non-compliant implementation so this will be a breaking change -> need for major version up.

### LINKED ISSUE #390: Capitalize does not follow Shopify spec - capitalizes each word

### LINKED ISSUE #400: Adding days doesn't work with Shopify's DotLiquid pattern
### Dotliquid version
Latest
### Expected behavior
Add 14 days to now
### Actual behavior
String concatenation instead of addition resulting in test: 16012856991209600

Template template = Template.Parse(@"test: {% assign seconds = 14 | times: 24 | times: 60 | times: 60 %}{{mydate | date: ""%s"" | plus: seconds | date: ""%a, %b %d, %y"" }}"); 
Console.WriteLine(template.Render(Hash.FromAnonymousObject(new { mydate = DateTime.Now })));

This is a common pattern in Shopify's DotLiquid to add days to dates.

You can test it here: https://dotnetfiddle.net/LoBHn0

### LINKED ISSUE #404: Number Conversion Inconsistent with Ruby Implementation
### Dotliquid version
2.0.366

### Expected behavior
`{{  '5' | plus: 5 }}` should output `10`
`{{  '5' | times: 5 }}` should output `25`
`{{  '5' | minus: 5 }}` should output `0`
`{{  '21' | divided_by: 3 }}` should output `7`
`{{ '8888' | modulo: 10 }}` should output `8`

### Actual behavior
`{{  '5' | plus: 5 }}` outputs `55`
`{{  '5' | times: 5 }}` outputs `55555`
`{{  '5' | minus: 5 }}` outputs `Liquid error: Parameter count mismatch.`
`{{  '21' | divided_by: 3 }}` outputs `Liquid error: Parameter count mismatch.`
`{{ '8888' | modulo:10 }}` outputs `Liquid error: Parameter count mismatch.`

Add test cases above to `StandardFilterTests`

### Description/Notes
In the ruby implementation from Shopify all mathematical operators use this method which attempts to convert the inputs to numbers regardless of their data type:

https://github.com/Shopify/liquid/blob/1850511334c555b74232166006a19e987ae7a41d/lib/liquid/standardfilters.rb#L453

However this implementation (`StandardFilters.cs`) uses `IsReal` which checks the .NET runtime type of the properties passed in.

There are also some cases such as `plus` and `times` which have explicitly different behavior in the case of `string` types compared with the `ruby` implementation. As such changing this to be consistent with ruby would be a breaking change.

Also in this implementation, if you change the operand to a floating point (`float`, `decimal` or `double`) we do attempt to convert the `input` in the cases of `minus`, `divided_by` and `modulo`. Which is in itself inconsistent behavior.

### LINKED ISSUE #419: Strict Liquid Syntax
This is a placeholder for various changes, for good or for worse, that are needed to make DotLiquid behave similar to the reference Liquid language. This will include breaking changes if enabled, but remain disabled by default (possibly until a major version release).

microalps:
# Describe in your message, what you're fixing. Reference the original issue, if possible.
Implements a way to opt-in to a more correct behavior in line with Ruby Liquid reference library that this is based on. This path was suggested by @daviburg to maintain backward compatibility. If opted in, Resolves #419 SubItems 1-4.
Resolves #316 , Resolves #328 , Resolves #400, Resolves #404, Resolves #322, Resolves #149, Resolves #309, Resolves #385, Resolves #390, Resolves #214, Closes #401, Closes #325

# Describe how if any API-breaking changes your PR may introduce.
The behavior does not change if you don't opt-in. All tests were run both prior to implementing the new behavior and after implementing the fixes. Tests marked V20 passed before and after while tests marked with V21 failed prior and passed after.
# State whether it's ready-to-review/-merge or if it's a WIP PR.
This is ready to merge for the issues addressed.

### DIFF
--- a/src/DotLiquid.Tests/StandardFilterTests.cs
+++ b/src/DotLiquid.Tests/StandardFilterTests.cs
@@ -10,6 +10,22 @@ namespace DotLiquid.Tests
     [TestFixture]
     public class StandardFilterTests
     {
+        private Context _contextV20;
+        private Context _contextV21;
+
+        [OneTimeSetUp]
+        public void SetUp()
+        {
+            _contextV20 = new Context(CultureInfo.InvariantCulture)
+            {
+                SyntaxCompatibilityLevel = SyntaxCompatibility.DotLiquid20
+            };
+            _contextV21 = new Context(CultureInfo.InvariantCulture)
+            {
+                SyntaxCompatibilityLevel = SyntaxCompatibility.DotLiquid21
+            };
+        }
+
         [Test]
         public void TestSize()
         {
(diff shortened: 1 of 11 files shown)


src/DotLiquid/SyntaxCompatibilityEnum.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    /// <summary>
+    /// Strict Liquid Syntax Compatibility Flag
+    /// </summary>
+    public enum SyntaxCompatibility
+    {
+        /// <summary>
+        /// Behavior as of DotLiquid 2.0
+        /// </summary>
+        COMPAT_2_0 = 200,
```
daviburg, line 11: Version2_0 (or Version_2_0)

In order to:
* Avoid all-caps
* Avoid non-standard abbreviation
* Substitute _ for . but not for space characters

src/DotLiquid/StandardFilters.cs
```diff
@@ -741,12 +763,20 @@ private static object DoMathsOperation(object input, object operand, Func<Expres
                 }
             }
 
-            return ExpressionUtility
+            try
+            {
+                return ExpressionUtility
                 .CreateExpression(
```
daviburg, line 769: Nit: one more tab to the right

src/DotLiquid/StandardFilters.cs
```diff
@@ -716,11 +734,15 @@ private static object DoMathsOperation(object input, object operand, Func<Expres
             if (input == null || operand == null)
                 return null;
 
+            bool shouldConvertStrings = false;
```
daviburg, line 737: Nit: The assignment of the variable, if condition, all can be done on a single line direct assignment.

src/DotLiquid/StandardFilters.cs
```diff
@@ -118,6 +118,11 @@ public static string Capitalize(string input)
             if (input.IsNullOrWhiteSpace())
                 return input;
 
+            if (Template.SyntaxCompatibilityLevel >= SyntaxCompatibility.COMPAT_2_1) {
+                string trimmed = input.TrimStart();
```
daviburg, line 122: nit: consider using var
I noticed at this time the code in this repo is mixing implicit and explicit types, not sure it is following a strict rule for it.

daviburg (approved):

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -739,13 +777,37 @@ public void TestPlus()
                 Helper.AssertTemplateResult("2", "{{ 1 | plus:1 }}");
                 Helper.AssertTemplateResult("5.5", "{{ 2  | plus:3.5 }}");
                 Helper.AssertTemplateResult("5.5", "{{ 3.5 | plus:2 }}");
-                Helper.AssertTemplateResult("11", "{{ '1' | plus:'1' }}");
```
mhornbacher, line 742: Perhaps just change this test to

`Helper.AssertTemplateResult("2", "{{ '1' | plus:'1' }}");` to assert the new behavior?

src/DotLiquid/SyntaxCompatibilityEnum.cs
microalps: You are right about case. Compat is definitely a used abbreviation (see referencesource.microsoft.com) but Version is a nice alternative. As far as underscore it seems the standard is to omit completely so Tls13 is 1.3.so it will be Version20 and Version21. Only Q is should we worry about Version 2.10 or will we bump to 3.x by then?

src/DotLiquid/StandardFilters.cs
microalps: Too many conditions to be readable IMHO.

microalps (commented):
There is no test for RemoveFirst behavior with both flags. In theory it follows ReplaceFirst, but unit tests are better.

microalps:
Before merging,  need to consider if static definition is recommended at this time or this should be added to RenderParameters and context. We can default to static property if not specified.

daviburg:
> Before merging, need to consider if static definition is recommended at this time or this should be added to RenderParameters and context. We can default to static property if not specified.

Static properties are challenging in shared hosting models like ours because we may be processing requests concurrently, requests which may not have the same versioning expectation.
On the other hand static properties is a precedence (e.g. Ruby date format), which we were not able to expose to customer in our hosting approach.

rarayudu (changes requested):

src/DotLiquid/Template.cs
```diff
@@ -58,6 +63,7 @@ static Template()
             RegexTimeOut = TimeSpan.FromSeconds(10);
             NamingConvention = new RubyNamingConvention();
             FileSystem = new BlankFileSystem();
+            SyntaxCompatibilityLevelDefault = SyntaxCompatibility.DotLiquid20;
```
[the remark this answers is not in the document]
microalps: I'd love to know the use case that you cant use the default and cant use the more flexible explicit RenderParameters. Having optional attributes become unwieldly as they are added to. If we change all options from static to the same 'default' behavior we would have a mess and implemented this we would have a mess. Im trying to think ahead.

microalps:
Rebased after merge of 421. Changed the naming of the property in Template to DefaultSyntaxCompatibilityLevel (similar to DefaultIsThreadSafe).

src/DotLiquid/Template.cs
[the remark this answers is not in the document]
microalps: See https://github.com/dotliquid/dotliquid/wiki/DotLiquid-Syntax-Compatibility#strict-liquid-syntax-work-in-progress for the no longer theoretical implementation of setting the syntax in a template.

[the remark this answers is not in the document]
rarayudu: We should be able to use the "public string Render(RenderParameters parameters)" overload by passing all the render parameters explicitly. Thought it would be easy for the users of "public string Render(Hash localVariables, IFormatProvider formatProvider = null)" to pass in the new syntax compatibility setting instead of creating caller creating render parameters. But Agree on your point that it might become a mess when we change more options.

rarayudu (approved):

PR #421, opened 2021-03-04
Math functions on uint and int and other combinations

### LINKED ISSUE #367: Plus filter doesn't work on typeof(uint) but does work on typeof(int)
### Dotliquid version
2.0.314
### Expected behavior
uint should work with the plus operator
### Actual behavior
`Liquid error: Parameter count mismatch.`
Pass a drop to a template with a single property of type uint. You should see the above error. Change to an int and it will work.

microalps:
Resolves #367

### DIFF
--- a/src/DotLiquid/Util/ExpressionUtility.cs
+++ b/src/DotLiquid/Util/ExpressionUtility.cs
@@ -21,13 +21,13 @@ namespace DotLiquid.Util
             // Using the promotion table at
             // https://docs.microsoft.com/en-us/dotnet/standard/base-types/conversion-tables
 
-            Add(typeof(Byte), typeof(UInt16), typeof(Int16), typeof(UInt32), typeof(Int32), typeof(UInt64), typeof(Int64), typeof(Single), typeof(Double), typeof(Decimal));
-            Add(typeof(SByte), typeof(Int16), typeof(Int32), typeof(Int64), typeof(Single), typeof(Double), typeof(Decimal));
-            Add(typeof(Int16), typeof(Int32), typeof(Int64), typeof(Single), typeof(Double), typeof(Decimal));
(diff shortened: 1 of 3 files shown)


src/DotLiquid/Util/ExpressionUtility.cs
```diff
@@ -21,13 +21,13 @@ static ExpressionUtility()
...
-            Add(typeof(Int32), typeof(Int64), typeof(Double), typeof(Decimal), typeof(Single));
-            Add(typeof(UInt32), typeof(Int64), typeof(UInt64), typeof(Double), typeof(Decimal), typeof(Single));
+            Add(typeof(Byte), typeof(UInt16), typeof(Int16), typeof(UInt32), typeof(Int32), typeof(UInt64), typeof(Int64), typeof(Decimal), typeof(Single), typeof(double));
+            Add(typeof(SByte), typeof(Int16), typeof(Int32), typeof(Int64), typeof(Decimal), typeof(Single), typeof(double));
+            Add(typeof(Int16), typeof(Int32), typeof(Int64), typeof(Decimal), typeof(Single), typeof(double));
+            Add(typeof(UInt16), typeof(UInt32), typeof(Int32), typeof(UInt64), typeof(Int64), typeof(Decimal), typeof(Single), typeof(double));
+            Add(typeof(Char), typeof(UInt16), typeof(UInt32), typeof(Int32), typeof(UInt64), typeof(Int64), typeof(Decimal), typeof(Single), typeof(double));
+            Add(typeof(Int32), typeof(Int64), typeof(Decimal), typeof(Single), typeof(double));
+            Add(typeof(UInt32), typeof(Int64), typeof(UInt64), typeof(Decimal), typeof(Single), typeof(double));
```
microalps, line 30: Changes to the order are required to ensure a consistent result whether you pass left, right or right, left. We try to use lowest required upgrade, e.g. decimal > single > double

daviburg (approved):

PR #422, opened 2021-03-05
Simplify and Reduce Reliance on Regex

### LINKED ISSUE #177: Optionally disable trimming
## Option to disable ERB-inspired whitespace trimming around tags

While profiling the application that I am working on, it has come to my attention that, on some of the templates that we currently use, `Template.Parse` takes two thirds of its time on ERB-like trimming. (The template is about 18000 characters long)

I would like to know whether you would be open to adding an option that would suppress trimming.

### LINKED ISSUE #199: `raw`\`endraw` can't escape string that contains {{ double opening curly brace
I was trying to do a trivial thing of rendering mixture of liquid and handlebars. But this case renders `{%raw%}` as if it's not a control tag.

Template:

``` liquid
{% for item in user.tasks -%}
  <li>{{ item.name }}</li>
  {%raw%}<li>{{ item.name {%endraw%} {%raw%}}}</li>  {%endraw%}
{% endfor -%}
```

Rendered as:

``` html
  <li>Documentation</li>
  <li>{{ item.name {%endraw%} {%raw%}}}</li>  
  <li>Code comments</li>
  <li>{{ item.name {%endraw%} {%raw%}}}</li>  
```

Should I somehow escape {{ symbols between raw\endraw tags?

### LINKED ISSUE #303: Bug on chained "replace" filters when second argument contains space characters.
### Dotliquid version: 
2.0.254

### Expected behavior
Hi.
I would expect to be able to use multiple _**" |replace: 'xxx','yyy' "**_ filters with space characters in the second parameter.

### Actual behavior
If more than one "replace" filters are chained, a space character in any but the last "replace" leads to following error message in rendered output: _**"Liquid error: Parameter count mismatch."**_

```
{%assign az='azerty'%}
1 {{az |replace: 'a','q'}}
2 {{az |replace: 'a','q '}}                    {%comment%} Note the space after 'q'  {%endcomment%}
3 {{az |replace: 'a','q'   |replace: 'z','w '}}{%comment%} Space after 'w'           {%endcomment%}
4 {{az |replace: 'a','q '  |replace: 'z','w'}} {%comment%} Space after 'q'           {%endcomment%}
5 {{az |replace: 'a','q r' |replace: 'z','w'}} {%comment%} Space between 'q' and 'r' {%endcomment%}
6 {{az |replace: 'a',' q'  |replace: 'z','w'}} {%comment%} Space before 'q'          {%endcomment%}
```

results in:
```
1 qzerty
2 q zerty
3 qw erty
4 Liquid error: Parameter count mismatch.
5 Liquid error: Parameter count mismatch.
6 Liquid error: Parameter count mismatch.
```
**Tests 4, 5 and 6** show that a space anywhere in the last argument of the first "replace" filter cause an error. As tested on other samples, any chained "replace" filter causes this error except for the last one.
**Tests 2 and 3** show that a space in the second argument of the last "replace" filter does NOT cause any problem.

If I'm not overseeing something, this looks like a bug to  me...
Thanks in advance for your attention.

### LINKED ISSUE #412: Parse error when string literals contain }} or %}
### Dotliquid version
Unknown (using Dotliquid online)

### Input
```liquid
{{ '{{ {% %} }}' }}
{% assign x = '{{ {% %} }}' %}
{{ x }} 
```

### Expected output
```
{{ {% %} }}

{{ {% %} }}
```

### Actual output
An exception is thrown

microalps:
- Backport https://github.com/Shopify/liquid/commit/292161865d24e3a8e58764c90dd011d9d494b155 Resolves #303
- Remove unused regex expressions
- Resolves #199, Resolves #412, Resolves #177

### DIFF
--- a/src/DotLiquid/DotLiquid.csproj
+++ b/src/DotLiquid/DotLiquid.csproj
@@ -5,7 +5,7 @@
     <AssemblyTitle>DotLiquid</AssemblyTitle>
     <NeutralLanguage>en-US</NeutralLanguage>
     <Authors>Tim Jones;Alessandro Petrelli</Authors>
-    <TargetFrameworks>netstandard1.3;net45</TargetFrameworks>
+    <TargetFrameworks>netstandard1.3;netstandard2.0;net45</TargetFrameworks>
     <AssemblyName>DotLiquid</AssemblyName>
     <AssemblyOriginatorKeyFile>../Formosatek-OpenSource.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
@@ -17,7 +17,6 @@
     <PackageLicenseUrl>http://www.apache.org/licenses/LICENSE-2.0</PackageLicenseUrl>
     <RepositoryType>git</RepositoryType>
     <RepositoryUrl>https://github.com/dotliquid/dotliquid</RepositoryUrl>
-    <NetStandardImplicitPackageVersion>1.6.0</NetStandardImplicitPackageVersion>
     <GenerateAssemblyTitleAttribute>false</GenerateAssemblyTitleAttribute>
     <GenerateAssemblyDescriptionAttribute>false</GenerateAssemblyDescriptionAttribute>
     <GenerateAssemblyConfigurationAttribute>false</GenerateAssemblyConfigurationAttribute>
@@ -34,7 +33,6 @@
 
   <ItemGroup>
     <PackageReference Include="SonarAnalyzer.CSharp" Version="5.11.0.1761" />
-    <PackageReference Include="System.ValueTuple" Version="4.3.0" />
   </ItemGroup>
 
   <ItemGroup>
@@ -47,9 +45,14 @@
     <DefineConstants>$(DefineConstants);CORE</DefineConstants>
   </PropertyGroup>
 
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.3' ">
+    <PackageReference Include="System.ValueTuple" Version="4.3.0" />
+  </ItemGroup>
+    
   <ItemGroup Condition=" '$(TargetFramework)' == 'net45' ">
     <Reference Include="System" />
     <Reference Include="Microsoft.CSharp" />
+    <PackageReference Include="System.ValueTuple" Version="4.3.0" />
   </ItemGroup>
 
 </Project>
--- a/src/DotLiquid/Tokenizer.cs
+++ b/src/DotLiquid/Tokenizer.cs
@@ -1,0 +1,155 @@
+using DotLiquid.Exceptions;
+using DotLiquid.Util;
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Text;
+using System.Text.RegularExpressions;
+
+namespace DotLiquid
+{
+    /// <summary>
+    /// Tokenizer uses a mix of Regular Expressions and text parsing as part of the compilation of strings into `blocks`.
+    /// </summary>
+    internal static class Tokenizer
+    {
+        private static readonly HashSet<char> SearchSingleQuoteEnd = new HashSet<char> { '\'' };
+        private static readonly HashSet<char> SearchDoubleQuoteEnd = new HashSet<char> { '"' };
+        private static readonly HashSet<char> SearchQuoteOrVariableEnd = new HashSet<char> { '}', '\'', '"' };
+        private static readonly HashSet<char> SearchQuoteOrTagEnd = new HashSet<char> { '%', '\'', '"' };
+        private static readonly Regex LiquidAnyStartingTagRegex = R.B(R.Q(@"({0})([-])?"), Liquid.AnyStartingTag);
+        private static readonly Regex TagNameRegex = R.B(R.Q(@"{0}\s*(\w+)"), Liquid.AnyStartingTag);
+        private static readonly ConcurrentDictionary<string, Regex> EndTagRegexes = new ConcurrentDictionary<string, Regex>();
+
+        /// <summary>
+        /// Splits a string into an array of `tokens` that represent either a tag, object/variable, or literal string
+        /// </summary>
+        /// <param name="source">The Liquid Template string</param>
+        /// <returns></returns>
+        /// <exception cref="SyntaxException"></exception>
+        internal static List<string> Tokenize(string source)
+        {
+            if (string.IsNullOrEmpty(source))
+                return new List<string>();
+
+            // Trim trailing whitespace.
+            source = Regex.Replace(source, string.Format(@"-({0}|{1})(\n|\r\n|[ \t]+)?", Liquid.VariableEnd, Liquid.TagEnd), "$1", RegexOptions.None, Template.RegexTimeOut);
+
+            var tokens = new List<string>();
+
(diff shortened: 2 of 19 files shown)


daviburg (approved):

microalps:
@mhornbacher Are you able to review this? @daviburg asked on gitter
> Any RegEx expert around to take a second look at #422 ?

microalps:
Going to look into #199 and #412 for inclusion in this PR.

src/DotLiquid/DotLiquid.csproj
```diff
@@ -5,7 +5,7 @@
     <AssemblyTitle>DotLiquid</AssemblyTitle>
     <NeutralLanguage>en-US</NeutralLanguage>
     <Authors>Tim Jones;Alessandro Petrelli</Authors>
-    <TargetFrameworks>netstandard1.3;net45</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net45;</TargetFrameworks>
```
daviburg, line 8: This has significant implication for consumers of the library. They no longer will be able to use dot liquid with older .NET standard. In the past this has been a blocking issue for PRs to merge to master.

daviburg (changes requested):
Tying to .NET Standard & Core change is a breaking change and blocking issue.

src/DotLiquid/DotLiquid.csproj
microalps: We have net45, are we supporting older than .NET 4.5?

src/DotLiquid.Website/ViewComponents/ContributorsViewComponent.cs
```diff
@@ -14,9 +13,9 @@ public class ContributorsViewComponent : ViewComponent
     {
         private readonly HttpClient _httpClient;
```
sebastienros, line 14: This should be static. Even if the data is cached, I understand that each call to this component will instantiate a new `HttpClient`, which is incorrect.

microalps: That was there before. However, now it is BETTER as this class is not even loaded because the view is cached and doesn't load it unnecessarily.

sebastienros: I understand it was there before. Many things in this PR were there before and you still changed it. I didn't notice it was cached, cool. And I made the <cache> tag helper, at least I am not useless here ;)

src/DotLiquid/DotLiquid.csproj
microalps: I've reincluded 1.3 and made a dedicated build for 2.0 to take advantage of the performance gains. Even running the test suite is about 33% longer on 1.3 vs 2.0

microalps:
I am aware that I only partially replaced the whitespace control regex, but it should be sufficient for now. I haven't done major benchmarks but it seems to be several times faster now due to this change alone.

src/DotLiquid/Tokenizer.cs
```diff
@@ -0,0 +1,139 @@
...
+using DotLiquid.Util;
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Text;
+using System.Text.RegularExpressions;
+
+namespace DotLiquid
+{
+    internal static class Tokenizer
```
daviburg, line 11: Consider adding XML comments to the class and its method so the intend is clear. It will help code maintainers to distinguish intend from actual implementation while troubleshooting issues.

src/DotLiquid/Util/CharEnumerator.cs
```diff
@@ -0,0 +1,73 @@
...
+using System.Text;
+
+namespace DotLiquid.Util
+{
+    internal class CharEnumerator : IEnumerator<char>
+    {
+        private readonly String str;
+        private int index;
+
+        internal CharEnumerator(String str)
```
daviburg, line 13: String already implements `IEnumerable<char>` so you can call GetEnumerator and get the `IEnumerator<char>`. This class appears unnecessary.

daviburg (changes requested):

sebastienros:
> Reduce Reliance on Regex

I am supportive of this effort, but I need to share my experience. When I was using DL, I realized it was not fast enough for my needs and started looking into improving the performance. I arrived with the "easy" conclusion that regexes were an issue and started to replace the most visible and simple one with a specialized piece of code instead, with a simple `Split` call AFAIR. But when I mentioned my change and the perf improvements in forum it got shut down. The reason was that the philosophy of the project was to copy as much as possible the original Ruby implementation. This is when I decided to build my own engine (if I had known about Scriban I would have use it instead).

Here the goal is different, mostly to fix a bug that is hard with a regex, so I hope (and believe) that's a more acceptable reason to do so.

src/DotLiquid/Tokenizer.cs
microalps: Will consider. It is internal, so it is not required. Likewise the CharEnumerator.

src/DotLiquid/Util/CharEnumerator.cs
microalps: Two reasons for this:
1. NET Standard 1.3 doesn't support it. Hence my attempt at bumping which was shot down
2. The .NET version is sealed and marks key properties as private. With our version, we exposed Position and Remaining, as well as Next and Previous characters so we can 'Peek' more like a stream.

microalps:
> Here the goal is different, mostly to fix a bug that is hard with a regex, so I hope (and believe) that's a more acceptable reason to do so.

Correct. Performance was considered only so as not to have a regression. My testing showed it was equivalent - until I snuck in the whitespace handling which pushed it way over the top (and that was only partial, so I can't wait to make another PR with the other half)

src/DotLiquid/Util/CharEnumerator.cs
daviburg: oh, ok, makes sense then.

src/DotLiquid/Tokenizer.cs
daviburg: Yes for CharEnumerator it would make even more sense to add XML comments to answer questions just like I had in this review (i.e. why do we even have such class).

microalps: Done for both.

daviburg (approved):

daviburg:
As the blocking issue of change of target support has been addressed, I am removing my request for change.
I'm a bit weary of the absence of input from engineers who objected to Sebastien earlier similar work.

microalps:
> I'm a bit weary of the absence of input from engineers who objected to Sebastien earlier similar work.

Would love to know who so we can ping them. Search for `is:pr author:sebastienros` returned no results.

sebastienros:
It was a discussion on gitter, I found it, and honestly it's not as strict as what I remembered. So I'd say as nobody seems to have any issue with it, just go on. I assume I just realized that at some point this would be limited. But your changes are scoped.

microalps:
Rebased on top of CodeCov PR and squashed bug fix commits into their parent.

daviburg:
@microalps when trying to look the full coverage report I get this error

Unauthorized
Github API: Forbidden Check on Codecov's status or see our docs for common support.

Error 403

Issue #289, opened 2017-12-13
Flatten() in sort() while using IEnumerable<IDictionary>


rarayudu:
In the StandardFilter.cs the sort() method is doing Flatten() on each item if the given input is of type IEnumerable.

Any reason why we are doing this? This is causing issues when trying to sort a list of dictionary objects based on a property as each dictionary item in the list is flattened. And because of this the specific if case which we have later for IDictionary type objects won't hit.

Can we not do flattening for the type IDictionary list, similar to what we are doing for type hash list.

microalps:
Valid Q - But closing as it is a duplicate of #228. More info in the related issue.

PR #434, opened 2021-05-25
Where filter implementation - resolves #347, resolves #374.

### LINKED ISSUE #347: DotLiquid - Array filters should support 'where' filter clause
### Dotliquid version
Latest Dot Liquid
### Expected behavior
'where' filter should work
### Actual behavior
'where' clause is not working

The original Liquid project supports 'where' clause 
https://help.shopify.com/en/themes/liquid/filters/array-filters#where

### LINKED ISSUE #374: Filter "where" not implemented
### Dotliquid version

2.0.333

### Expected behavior

I should be able to use the `where` filter, that is available and documented [HERE](https://shopify.github.io/liquid/filters/where/)

### Actual behavior

`where` filter not available

Just try to do a template with `where` filter in it.

I can try to provide a PR with the implementation for evaluation.

robin-parker:
Added where filter and associated tests base upon Shopify documentation.


microalps:
I'm sure @daviburg  is going to ask - how does this differ from #389 ? Is there any pros/cons to each implementation?

robin-parker:
@microalps, this PR include all the shopify samples as test cases, other than that no clear pros/cons I can see.

daviburg:
> @microalps, this PR include all the shopify samples as test cases

That's a pretty good start! 😊

src/DotLiquid/StandardFilters.cs
```diff
@@ -931,6 +931,62 @@ public static IEnumerable Compact(object input)
...
+        {
+            if (input == null)
+                return null;
+
+            if (property.IsNullOrWhiteSpace()){
+                throw new ArgumentNullException($"'{nameof(property)}' cannot be null or empty.", nameof(property));
+            }
+
+            List<object> inputList;
+            if(input is IEnumerable<Hash> enumerableHash && !string.IsNullOrEmpty(property))
```
daviburg, line 952: nit: space character after if keyword and before opening parenthesis

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -1399,5 +1406,105 @@ public void TestCompact()
...
+            // First Shopify sample
+            Helper.AssertTemplateResult(
+                expected: "\r\n\r\nKitchen products:\r\n\r\n- Spatula\r\n\r\n- Garlic press\r\n",
+                template: @"{% assign kitchen_products = products | where: ""type"", ""kitchen"" %}
+
+Kitchen products:
+{% for product in kitchen_products %}
+- {{ product.title }}
+{% endfor %}",
+                localVariables: Hash.FromAnonymousObject(new  { products }) );
```
daviburg, line 1464: Nit: extraneous double space character after new keyword.

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -1399,5 +1406,105 @@ public void TestCompact()
...
+        }
+
+        // Second sample from specification at https://shopify.github.io/liquid/filters/where/
+        // Say instead you have a list of products and you only want to show those that are available to buy.
+        // You can where with a property name but no target value to include all products with a truthy "available" value.
+        [Test]
+        public void TestWhere_ShopifySample2()
+        {
+            var products = new System.Collections.Generic.List<Hash>();
```
daviburg, line 1473: Could you declare a using System.Collections.Generic; atop the file instead?

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -1399,5 +1406,105 @@ public void TestCompact()
...
+            products.Add(new Hash { {"title","Coffee mug"}, {"available",true} });
```
daviburg, line 1474: I believe List <> supports collection initializer, which would simplify the code here.

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -1399,5 +1406,105 @@ public void TestCompact()
...
+            products.Add(new Hash { {"title","Limited edition sneakers"} /*no 'available' property*/ });
```
daviburg, line 1475: Avoid mixing C-style comments if possible.

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -1399,5 +1406,105 @@ public void TestCompact()
...
+            products.Add(new Hash { {"title","Limited edition sneakers"}, {"available",false} }); // 'available' = false
```
daviburg, line 1476: nit: spacing, here and elsewhere, the comma character should be followed by a space character

src/DotLiquid/StandardFilters.cs
```diff
@@ -931,6 +931,62 @@ public static IEnumerable Compact(object input)
...
+        /// <param name="property">The name of the property to filter by</param>
+        /// <param name="value">Value to retain, if null object containing this property are retained</param>
+        /// <returns></returns>
+        public static IEnumerable Where(object input, string property, string value = null)
+        {
+            if (input == null)
+                return null;
+
+            if (property.IsNullOrWhiteSpace()){
+                throw new ArgumentNullException($"'{nameof(property)}' cannot be null or empty.", nameof(property));
```
daviburg, line 948: use name arguments because both arguments have the same type so their order is ambiguous to human code reader. Named arguments allows human to assert that the order is correct.

src/DotLiquid/StandardFilters.cs
```diff
@@ -931,6 +931,62 @@ public static IEnumerable Compact(object input)
...
+            }
+
+            List<object> inputList;
+            if(input is IEnumerable<Hash> enumerableHash && !string.IsNullOrEmpty(property))
+                inputList = enumerableHash.Cast<object>().ToList();
```
daviburg, line 953: Use chained ternary operator.

Also when using if blocks, consistently use or don't use curvy bracket blocks to continue the existing code style in the file/repo.

src/DotLiquid/StandardFilters.cs
```diff
@@ -931,6 +931,62 @@ public static IEnumerable Compact(object input)
...
+            else if (input is IEnumerable enumerableInput)
+                inputList = enumerableInput.Flatten().Cast<object>().ToList();
```
microalps, line 955: See #228 and #289 - I'd like to ask the same Q here.

daviburg, line 955: What happens to the enumerable content if the enumerated type is a value type? i.e. not objects.

src/DotLiquid/StandardFilters.cs
```diff
@@ -931,6 +931,62 @@ public static IEnumerable Compact(object input)
...
+            else
+            { 
+                inputList = new List<object>(new[] { input });
+            }
+
+            if (!inputList.Any())
+                return inputList;
+
+            return inputList.FindAll(s => Matches(s, property, value));
```
daviburg, line 964: Why are we not using the Linq native Where method?

src/DotLiquid/StandardFilters.cs
```diff
@@ -931,6 +931,62 @@ public static IEnumerable Compact(object input)
...
+            }
+
+            List<object> inputList;
+            if(input is IEnumerable<Hash> enumerableHash && !string.IsNullOrEmpty(property))
+                inputList = enumerableHash.Cast<object>().ToList();
+            else if (input is IEnumerable enumerableInput)
+                inputList = enumerableInput.Flatten().Cast<object>().ToList();
+            else
+            { 
+                inputList = new List<object>(new[] { input });
```
microalps, line 958: Is there a test case for this? Is this compliant with Ruby Liquid - I didn't see this logic in https://github.com/Shopify/liquid/blob/master/lib/liquid/standardfilters.rb#L190

src/DotLiquid/StandardFilters.cs
```diff
@@ -931,6 +931,62 @@ public static IEnumerable Compact(object input)
...
+            }
+
+            if (!inputList.Any())
+                return inputList;
+
+            return inputList.FindAll(s => Matches(s, property, value));
+        }
+
+        private static bool Matches(object any, string property, string value)
```
daviburg, line 967: In theory the implementation of PR#389 with a .NET IEqualityComparer hash provider will perform better than something doing a full match check on every value.

src/DotLiquid/StandardFilters.cs
```diff
@@ -931,6 +931,62 @@ public static IEnumerable Compact(object input)
...
+            bool matches = false;
+            if (any is IDictionary && ((IDictionary)any).Contains(property))
+            {
+                // the 'any' dictionary has the filter property
+                var propertyValue = ((IDictionary)any)[property];
+                if (value == null) {
+                    // If no filter value provided, check the property value is truthy.
+                    matches = propertyValue!=null && !false.Equals(propertyValue);
+                } else {
+                    matches = value.Equals( propertyValue );
```
daviburg, line 978: don't add extra space character after opening parenthesis or before closing parenthesis

daviburg (changes requested):

src/DotLiquid/StandardFilters.cs
```diff
@@ -931,6 +931,62 @@ public static IEnumerable Compact(object input)
...
+            if (!inputList.Any())
+                return inputList;
+
+            return inputList.FindAll(s => Matches(s, property, value));
+        }
+
+        private static bool Matches(object any, string property, string value)
+        {
+            bool matches = false;
+            if (any is IDictionary && ((IDictionary)any).Contains(property))
```
microalps, line 970: Skip the cast and declare in your test. https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/patterns#declaration-and-type-patterns

microalps (commented):
Sorry to harp on your WIP, just noted a few stuff. DaviBurg is the resident code-review expert though, so you can expect his feedback too.

daviburg:
> Sorry to harp on your WIP, just noted a few stuff. DaviBurg is the resident code-review expert though, so you can expect his feedback too.

Just trying to help. Please don't hold back in your own helping with review input.

src/DotLiquid/StandardFilters.cs
robin-parker: The .editorconfig was incorrectly configured, it’s fixed now and formatted.

src/DotLiquid.Tests/StandardFilterTests.cs
robin-parker: Fixed by .editorconfig

src/DotLiquid.Tests/StandardFilterTests.cs
robin-parker: Fixed by .editorconfig

src/DotLiquid.Tests/StandardFilterTests.cs
robin-parker: Done

src/DotLiquid.Tests/StandardFilterTests.cs
robin-parker: Done

src/DotLiquid.Tests/StandardFilterTests.cs
robin-parker: OK, done

src/DotLiquid.Tests/StandardFilterTests.cs
robin-parker: Fixed by .editorconfig

src/DotLiquid/StandardFilters.cs
robin-parker: Done

src/DotLiquid/StandardFilters.cs
robin-parker: They get autoboxed as object and an empty list is returned – test case added

src/DotLiquid/StandardFilters.cs
robin-parker: Done

src/DotLiquid/StandardFilters.cs
robin-parker: Updated - I'm new to C# and wasn't aware of Linq Where.

src/DotLiquid/StandardFilters.cs
robin-parker: Done

src/DotLiquid/StandardFilters.cs
robin-parker: Flatten() has been removed, it was copied from Sort

src/DotLiquid/StandardFilters.cs
robin-parker: Unused code, I’ve removed it.

src/DotLiquid/StandardFilters.cs
robin-parker: Thanks, cool feature.

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -1399,5 +1407,113 @@ public void TestCompact()
...
+
+        // Second sample from specification at https://shopify.github.io/liquid/filters/where/
+        // Say instead you have a list of products and you only want to show those that are available to buy.
+        // You can where with a property name but no target value to include all products with a truthy "available" value.
+        [Test]
+        public void TestWhere_ShopifySample2()
+        {
+            List<Hash> products = new List<Hash> {
+                new Hash { { "title", "Coffee mug" }, { "available", true } },
+                new Hash { { "title", "Limited edition sneakers" } }, //no 'available' property
```
daviburg, line 1483: nit nit: space character after // comment starting escape characters. i.e.
`// no 'available' property`

src/DotLiquid/StandardFilters.cs
```diff
@@ -931,6 +931,53 @@ public static IEnumerable Compact(object input)
...
+        {
+            if (input == null)
+                return null;
+
+            if (property.IsNullOrWhiteSpace())
+                throw new ArgumentNullException(paramName: nameof(property), message: $"'{nameof(property)}' cannot be null or empty.");
+
+            List<object> inputList = (input is IEnumerable enumerableInput) ? enumerableInput.Cast<object>().ToList() : null;
+
+            return inputList.Where(s => s.Matches(property, target_value));
```
daviburg, line 952: please spell out with an English word what the 's' variable is, for code readability.

src/DotLiquid/StandardFilters.cs
```diff
@@ -931,6 +931,53 @@ public static IEnumerable Compact(object input)
...
+        {
+            bool matches = false;
+            if (any is IDictionary dictionary && dictionary.Contains(property))
+            {
+                // the 'any' dictionary has the filter property
+                var propertyValue = dictionary[property];
+                if (target_value == null)
+                {
+                    // If no filter value provided, check the property value is truthy.
+                    matches = propertyValue != null && !false.Equals(propertyValue);
```
daviburg, line 965: how does the default equality behave with regards to casing and local culture?
i.e. what should be the behavior for False (vs false)?
Is there an invariant version of the equals method for when the local is not English?

src/DotLiquid/StandardFilters.cs
```diff
@@ -931,6 +931,53 @@ public static IEnumerable Compact(object input)
...
+                }
+                else
+                {
+                    matches = target_value.Equals(propertyValue);
```
daviburg, line 969: Specific string comparison culture consideration, e.g. CurrentCulture. Check if there is a pattern to follow from other existing string comparisons.

src/DotLiquid/StandardFilters.cs
```diff
@@ -931,6 +931,53 @@ public static IEnumerable Compact(object input)
...
+                }
+            }
+            else if (any.RespondTo(property))
```
daviburg, line 972: Can any be null? i.e. can there be a null entry in the input list?

src/DotLiquid/StandardFilters.cs
```diff
@@ -931,6 +931,53 @@ public static IEnumerable Compact(object input)
...
+
+            if (property.IsNullOrWhiteSpace())
+                throw new ArgumentNullException(paramName: nameof(property), message: $"'{nameof(property)}' cannot be null or empty.");
+
+            List<object> inputList = (input is IEnumerable enumerableInput) ? enumerableInput.Cast<object>().ToList() : null;
+
+            return inputList.Where(s => s.Matches(property, target_value));
+        }
+
+        private static bool Matches(this object any, string property, string target_value)
```
daviburg, line 955: This matches helper method implements the specific Shopify liquid concept of Where filter match. See if there is an appropriate method name to describe that purpose. Either way, do XML comment this method to explain what it is going and what are each of the input parameters.

daviburg (changes requested):

src/DotLiquid/StandardFilters.cs
daviburg: That's alright. Thank you for your contribution. Impressive for someone just starting!

daviburg:
@microalps I only have small change requests from my second pass review and expect to sign-off the next iteration. Could you check if you are also good with the code? I am thinking we don't need to strongly push for an Equality pattern given that the Shopify Where standard filter has a very different meaning than the Linq Where method.

daviburg:
@robin-parker invited you also to [DotLiquid Gitter](https://gitter.im/dotliquid/dotliquid?utm_source=share-link&utm_medium=link&utm_campaign=share-link) so you can talk with other contributors.

src/DotLiquid/StandardFilters.cs
```diff
@@ -931,6 +931,53 @@ public static IEnumerable Compact(object input)
...
+        /// <returns></returns>
+        public static IEnumerable Where(object input, string property, string target_value = null)
+        {
+            if (input == null)
+                return null;
+
+            if (property.IsNullOrWhiteSpace())
+                throw new ArgumentNullException(paramName: nameof(property), message: $"'{nameof(property)}' cannot be null or empty.");
+
+            List<object> inputList = (input is IEnumerable enumerableInput) ? enumerableInput.Cast<object>().ToList() : null;
```
daviburg, line 950: If this line returns null the next line will likely fail with null deref exception. Please add test coverage for this and fix as appropiate.

src/DotLiquid/StandardFilters.cs
daviburg, line 965: Codecov report says you only have partial coverage for this line, please investigate what test case you might be missing.

src/DotLiquid.Tests/StandardFilterTests.cs
robin-parker: Done

src/DotLiquid/StandardFilters.cs
robin-parker: Done

src/DotLiquid/StandardFilters.cs
robin-parker: Refactored to introduce a TruthyAndFalsy class with IsTruthy/IsFalse methods, added test cases to cover common examples for false and true.

src/DotLiquid/StandardFilters.cs
robin-parker: Test case and null-check added.

src/DotLiquid/StandardFilters.cs
robin-parker: Renamed to HasMatchingProperty() and added comment.

src/DotLiquid/StandardFilters.cs
robin-parker: Change type of input to IEnumerable, removed superfluous if/else and added test case coverage for nulls

src/DotLiquid/StandardFilters.cs
robin-parker: Simplified method and introduced StringComparison.CurrentCulture for equality.

src/DotLiquid/StandardFilters.cs
robin-parker: Done

src/DotLiquid.Tests/TruthyAndFalsyTests.cs
```diff
@@ -0,0 +1,60 @@
+﻿using System;
+using NUnit.Framework;
+
+namespace DotLiquid.Tests
+{
+
```
daviburg, line 6: nit: extraneous empty line

src/DotLiquid.Tests/TruthyAndFalsyTests.cs
```diff
@@ -0,0 +1,60 @@
...
+    [TestFixture]
+    public class TruthyAndFalsyTests
```
daviburg, line 8: Should you have the specification sample templates as tests also here?

i.e. https://shopify.github.io/liquid/basics/truthy-and-falsy/

src/DotLiquid/StandardFilters.cs
```diff
@@ -931,6 +931,43 @@ public static IEnumerable Compact(object input)
...
+
+        /// <summary>
+        /// Checks if the given object has a matching property name.
+        /// * If target_value is provided, then the property is compared to target_value
+        /// * If target_value is null, then the property is checked for "Truthyness".
+        /// <param name="any">an object to be assessed</param>
+        /// <param name="propertyName">The name of the property to test for</param>
+        /// <param name="target_value">target property value</param>
+        /// <returns></returns>
+        /// </summary>
```
daviburg, line 961: Summary end tag should be after line 956

src/DotLiquid/StandardFilters.cs
```diff
@@ -931,6 +931,43 @@ public static IEnumerable Compact(object input)
...
+        private static bool HasMatchingProperty(this object any, string propertyName, string target_value)
+        {
+            object propertyValue = any is IDictionary dictionary ? dictionary[propertyName]
```
daviburg, line 964: The dictionary array accessor will throw KeyNotFoundException if the property is not found. You probably didn't plan for exceptions here.

https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.idictionary-2.item?view=net-5.0#exceptions

 You probably want to use ContainsKey instead 

https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.idictionary-2.containskey?view=net-5.0

daviburg: And if this is comment is accurate, there is likely a gap in the test cases.

src/DotLiquid/TruthyAndFalsy.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// </summary>
+    internal static class TruthyAndFalsy
+    {
+        /// <summary>
+        /// All values in Liquid are Truthy except nil and false.
+        /// </summary>
+        /// <param name="any">any object</param>
+        /// <returns></returns>
+        public static bool IsTruthy(this object any)
+        {
```
daviburg, line 21: For single-liners you can use expression i.e. 

`public static bool IsTruthy(this object any)  => !IsFalsy(any);`

src/DotLiquid/TruthyAndFalsy.cs
```diff
@@ -0,0 +1,35 @@
...
+        }
+
+        /// <summary>
+        /// The only values that are Falsy in Liquid are nil and false.
+        /// </summary>
+        /// <param name="any">any object</param>
+        /// <returns></returns>
+        public static bool IsFalsy(this object any)
+        {
+            return any == null || (any is bool _bool && _bool == false) || "false".Equals(any.ToString().ToLowerInvariant());
```
daviburg, line 32: I'm hesitant about this .ToString()
Is this test meaningful for any type other than string? if this is just for the string type, then `any is string _string `would be a safer approach.

daviburg (changes requested):

daviburg:
Apologies, I thought this time I would be able to sign-off.

src/DotLiquid.Tests/TruthyAndFalsyTests.cs
robin-parker: Removed

src/DotLiquid.Tests/TruthyAndFalsyTests.cs
robin-parker: Done

src/DotLiquid/StandardFilters.cs
robin-parker: Done

src/DotLiquid/TruthyAndFalsy.cs
robin-parker: Done

src/DotLiquid/TruthyAndFalsy.cs
robin-parker: It's just intended for string, so amended as suggested.

src/DotLiquid/StandardFilters.cs
robin-parker: Reinstated original if/else block with contains(key).
I can't spot the gap in test cases, could you elaborate on the case I've missed?

robin-parker:
> Apologies, I thought this time I would be able to sign-off.

No problem, hopefully third times a charm.

src/DotLiquid/StandardFilters.cs
daviburg: The missing case is any is implementing IDictionary but dictionary does not contain propertyName. Should result in dictionary[propertyName] throws KeyNotFoundException

src/DotLiquid/TruthyAndFalsy.cs
```diff
@@ -0,0 +1,34 @@
...
+        /// <summary>
+        /// The only values that are Falsy in Liquid are nil and false.
+        /// </summary>
+        /// <param name="any">any object</param>
+        /// <returns></returns>
+        public static bool IsFalsy(this object any)
+        {
+            return any == null
+                || (any is bool _bool && _bool == false)
+                || (any is string _string && "false".Equals(_string.ToLowerInvariant()));
```
daviburg, line 31: I forgot, specify string comparison type instead, as built-in .NET string comparison

"false".Equals(_string, StringComparison.InvariantCultureIgnoreCase)

src/DotLiquid/StandardFilters.cs
```diff
@@ -931,6 +931,51 @@ public static IEnumerable Compact(object input)
...
         }
+
+        /// <summary>
+        /// Creates an array including only the objects with a given property value, or any truthy value by default.
+        /// </summary>
+        /// <param name="input">an array to be filtered</param>
+        /// <param name="propertyName">The name of the property to filter by</param>
+        /// <param name="target_value">Value to retain, if null object containing this property are retained</param>
+        /// <returns></returns>
+        public static IEnumerable Where(IEnumerable input, string propertyName, string target_value = null)
```
daviburg, line 942: The Shopify specification is ambiguous but I think it would be correct to interpret the target value as of any type. This profile assumes the target value is a string. But say we filter a by a price property, the target value may be a decimal type instead. e.g.

{% assign ten_bucks_products = products | where: "price", 9.99 %}

There is probably precedence in the code base on what is the pattern to deal with ambiguous types.

src/DotLiquid/StandardFilters.cs
```diff
@@ -931,6 +931,51 @@ public static IEnumerable Compact(object input)
             ary.RemoveAll(item => item == null);
             return ary;
         }
+
+        /// <summary>
+        /// Creates an array including only the objects with a given property value, or any truthy value by default.
+        /// </summary>
+        /// <param name="input">an array to be filtered</param>
+        /// <param name="propertyName">The name of the property to filter by</param>
+        /// <param name="target_value">Value to retain, if null object containing this property are retained</param>
```
daviburg, line 940: "Value to retain, if null object containing this property are retained"

Null objects would not be retained, would they?

This parameter description is likely wrong. The target value specifies what value the matched named property must have for the containing objects to be selected by the filter.

src/DotLiquid/StandardFilters.cs
daviburg: I think existing filters dealing with ambiguously typed inputs use `object` parameter type then test the type (is string, is int, is double, is DateTimeOffset, etc.)

We have existing object to object equality implementation there, which likely you can call:
https://github.com/dotliquid/dotliquid/blob/ddaa286eb24f46cbe19a00a35ef5ea4987c24a8a/src/DotLiquid/Condition.cs#L116

daviburg, line 942: rename target_value to targetValue to continue following camel casing convention.

src/DotLiquid/StandardFilters.cs
robin-parker: That test case is already present see "Tartan flat cap" in TestWhere_ShopifySample3.

src/DotLiquid/TruthyAndFalsy.cs
robin-parker: StringComparison.InvariantCultureIgnoreCase in not available in netcore1.0, I've used StringComparison.OrdinalIgnoreCase, which has been used elsewhere in StandardFilters.cs

src/DotLiquid/StandardFilters.cs
robin-parker: @daviburg, the Condition constructor requires string for left/right, would an acceptable alternative be to call SafeTypeInsensitiveEqual in ObjectExtensionMethods.cs?

Also, is ObjectExtensionMethods.cs a better class for IsTruthy/IsFalsy? I could easily refactor them.

src/DotLiquid/StandardFilters.cs
robin-parker: Comment has been updated

src/DotLiquid/StandardFilters.cs
robin-parker: Done

src/DotLiquid/TruthyAndFalsy.cs
daviburg: Thanks for checking what's available with .NET Core 1.0. Good call to re-use what's already used elsewhere in the file.

src/DotLiquid/StandardFilters.cs
daviburg: Yes ObjectExtensionMethods is the right place to look for variable equality helper. If what exists there doesn't work for the object equality you need, I would add the newly needed variant there.

As the ObjectExtensionMethods existing class is pretty small, and so would be the new ThruthyAndFalsy file, I agree it is good to refactor the new extensions into the existing file.

src/DotLiquid.Tests/Util/ObjectExtensionMethodsTests.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        private static readonly object NIL = null;
+
+        [Test]
+        public void TestSafeTypeInsensitiveEqual()
+        {
+            // Not equal
+            Assert.False(NIL.SafeTypeInsensitiveEqual("nil"));
+            Assert.False("nil".SafeTypeInsensitiveEqual(null));
+            Assert.False("a string".SafeTypeInsensitiveEqual("A STRING")); // diffreent case string equality
```
daviburg, line 20: nit: misspelled 'diffreent '

src/DotLiquid.Tests/Util/ObjectExtensionMethodsTests.cs
```diff
@@ -0,0 +1,31 @@
+using System.Collections.Generic;
```
daviburg, line 1: Nit: best to keep all the using statements in a single block (no empty line) and organize them with System* first, then everything else, each sub-group alphabetically ordered.

src/DotLiquid/StandardFilters.cs
robin-parker: Refactored.

src/DotLiquid.Tests/Util/ObjectExtensionMethodsTests.cs
robin-parker: Done

src/DotLiquid.Tests/Util/ObjectExtensionMethodsTests.cs
robin-parker: Fixed.

daviburg (approved):

PR #435, opened 2021-05-26
Strftime enhancements, Resolves #251

### LINKED ISSUE #251: %-d for date Ruby format is not rendered
### Dotliquid version
2.0.145

### Expected behavior
I'm using ` | date: "%-d %b %Y"` as a date format. I'd expect the `%-d` to be rendered the same way as in Ruby. That is, the day without left padding zero (if it's single digit).

I'm using `Liquid.UseRubyDateFormat = true;`.

### Actual behavior
The `%-d` stays in final output.

```
Thread.CurrentThread.CurrentCulture = new CultureInfo("en-US");
var result = Template.Parse("{{ foo | date: \"%-d %b %Y\" }}").Render(Hash.FromAnonymousObject(new { foo = DateTime.UtcNow }));
Console.WriteLine(result);
```

robin-parker:
Implemented several missing directives and flags, also added documentation to cover deviation from Ruby (e.g. %N, %9N, %Z).

### DIFF
--- a/src/DotLiquid.Tests/Util/StrFTimeTests.cs
+++ b/src/DotLiquid.Tests/Util/StrFTimeTests.cs
@@ -1,6 +1,5 @@
 using System;
 using System.Globalization;
-using System.Threading;
 using DotLiquid.Util;
 using NUnit.Framework;
 
@@ -115,32 +114,32 @@ namespace DotLiquid.Tests.Util
             Assert.That(now.DateTime.ToStrFTime("%Z"), Is.EqualTo(timeZoneOffset));
         }
 
-        [TestCase("%z", ExpectedResult = "Z")] // ISO 8601 offset from UTC in timezone (1 minute=1, 1 hour=100) If timezone cannot be determined, no characters (+100)
-        [TestCase("%:z", ExpectedResult = "Z")] // hour and minute offset from UTC with a colon (e.g. +09:00)
+        [TestCase("%z", ExpectedResult = "+0000")] // hour and minute offset from UTC without a colon
+        [TestCase("%:z", ExpectedResult = "+00:00")] // hour and minute offset from UTC with a colon
         public string TestTimeZoneUTC(string format)
         {
             using (CultureHelper.SetCulture("en-GB"))
             {
                 Assert.That(CultureInfo.CurrentCulture, Is.EqualTo(new CultureInfo("en-GB")));
-                return DateTime.UtcNow.ToStrFTime(format);
+                return DateTimeOffset.UtcNow.ToStrFTime(format);
             }
         }
 
-        [TestCase("%z", ExpectedResult = "+0100")] // ISO 8601 offset from UTC in timezone (1 minute=1, 1 hour=100) If timezone cannot be determined, no characters (+100)
-        [TestCase("%:z", ExpectedResult = "+01:00")] // hour and minute offset from UTC with a colon (e.g. +09:00)
+        [TestCase("%z", ExpectedResult = "-0500")] // hour and minute offset from UTC without a colon
+        [TestCase("%:z", ExpectedResult = "-05:00")] // hour and minute offset from UTC with a colon
         public string TestTimeZoneLocal(string format)
         {
             using (CultureHelper.SetCulture("en-GB"))
             {
                 Assert.That(CultureInfo.CurrentCulture, Is.EqualTo(new CultureInfo("en-GB")));
-                // Jun-10 is during UK British Summer Time (BST / +01:00)
-                return DateTimeOffset.Parse("2012-06-10T14:32:14+01:00").ToStrFTime(format);
+                return DateTimeOffset.Parse("2012-06-10T14:32:14-05:00").ToStrFTime(format);
             }
         }
 
-        [TestCase("%G", ExpectedResult = "2013")] // ISO 8601 offset from UTC in timezone (1 minute=1, 1 hour=100) If timezone cannot be determined, no characters (+100)
-        [TestCase("%g", ExpectedResult = "13")] // hour and minute offset from UTC with a colon (e.g. +09:00)
-        [TestCase("%V", ExpectedResult = "01")] // hour and minute offset from UTC with a colon (e.g. +09:00)
+        // '2012-12-31' is considered to be in the first week of 2013 (according to ISO-8601).
+        [TestCase("%G", ExpectedResult = "2013")]
+        [TestCase("%g", ExpectedResult = "13")]
+        [TestCase("%V", ExpectedResult = "01")]
         public string TestIso8601WeekBasedDates(string format)
         {
             using (CultureHelper.SetCulture("en-GB"))
--- a/src/DotLiquid.Tests/Util/StrFTimeTests.cs
+++ b/src/DotLiquid.Tests/Util/StrFTimeTests.cs
@@ -74,6 +74,7 @@ namespace DotLiquid.Tests.Util
         [TestCase("%_2Y", ExpectedResult = "2012")] // Ensure space-pad does not truncate.
         // %z, %:z, %::z --> See TestTimeZoneUTC/TestTimeZoneLocal
         [TestCase("%", ExpectedResult = "%")]
+        [TestCase("%%", ExpectedResult = "%")]
         public string TestFormat(string format)
         {
             using (CultureHelper.SetCulture("en-GB"))
--- a/src/DotLiquid.Tests/Util/StrFTimeTests.cs
+++ b/src/DotLiquid.Tests/Util/StrFTimeTests.cs
@@ -10,39 +10,77 @@ namespace DotLiquid.Tests.Util
     [TestFixture]
     public class StrFTimeTests
     {
-        [TestCase("%a", ExpectedResult = "Sun")]
-        [TestCase("%A", ExpectedResult = "Sunday")]
-        [TestCase("%b", ExpectedResult = "Jan")]
-        [TestCase("%B", ExpectedResult = "January")]
-        [TestCase("%c", ExpectedResult = "Sun Jan 08 14:32:14 2012")]
-        [TestCase("%C", ExpectedResult = "20")]
-        [TestCase("%d", ExpectedResult = "08")]
-        [TestCase("%e", ExpectedResult = " 8")]
-        [TestCase("%h", ExpectedResult = "Jan")]
-        [TestCase("%H", ExpectedResult = "14")]
-        [TestCase("%I", ExpectedResult = "02")]
-        [TestCase("%j", ExpectedResult = "008")]
-        [TestCase("%k", ExpectedResult = "14")]
-        [TestCase("%l", ExpectedResult = "2")]
-        [TestCase("%m", ExpectedResult = "01")]
-        [TestCase("%M", ExpectedResult = "32")]
-        [TestCase("%P", ExpectedResult = "pm")]
-        [TestCase("%p", ExpectedResult = "PM")]
-        [TestCase("%S", ExpectedResult = "14")]
-        [TestCase("%u", ExpectedResult = "7")]
-        [TestCase("%U", ExpectedResult = "02")]
-        [TestCase("%W", ExpectedResult = "01")]
-        [TestCase("%w", ExpectedResult = "0")]
-        [TestCase("%x", ExpectedResult = "08/01/2012")]
-        [TestCase("%X", ExpectedResult = "14:32:14")]
-        [TestCase("%y", ExpectedResult = "12")]
-        [TestCase("%Y", ExpectedResult = "2012")]
+        // Descriptions below are taken from the Ruby Time.strftime documentation.
+        [TestCase("%a", ExpectedResult = "Sun")] // Abbreviated weekday name (Mon)
+        [TestCase("%A", ExpectedResult = "Sunday")] // Full weekday name (Monday)
+        [TestCase("%b", ExpectedResult = "Jan")] // Abbreviated month name (Jan)
+        [TestCase("%B", ExpectedResult = "January")] // Full month name (January)
+        [TestCase("%^B", ExpectedResult = "JANUARY")] // Upcased month name
+        [TestCase("%^_10B", ExpectedResult = "   JANUARY")] // Upcase, then Space-pad
+        [TestCase("%_^10B", ExpectedResult = "   JANUARY")] // Space-pad, then Upcase
+        [TestCase("%c", ExpectedResult = "Sun Jan 08 14:32:14 2012")] // Date and time representation (Thu Aug 23 14:55:02 2001)
+        [TestCase("%C", ExpectedResult = "20")] //Century (20 in 2009)
+        [TestCase("%d", ExpectedResult = "08")] // Day of Month, zero-padded (01..31)
+        [TestCase("%-d", ExpectedResult = "8")] // Remove zero-padding
+        [TestCase("%D", ExpectedResult = "01/08/12")] // Short date (%m/%d/%y)
+        [TestCase("%e", ExpectedResult = " 8")] // Day of month, space-padded ( 1..31)
+        // %E - not specified
+        // %f - not specified
+        [TestCase("%F", ExpectedResult = "2012-01-08")] // ISO 8601 date (2001-08-23)
+        [TestCase("%g", ExpectedResult = "12")] // Week-based year, last two digits (00..99)
+        [TestCase("%G", ExpectedResult = "2012")]  // Week based year (2001)
+        [TestCase("%h", ExpectedResult = "Jan")] // Abbreviated month name (Jan)
+        [TestCase("%H", ExpectedResult = "14")] // Hour in 24h format, zero-padded (00..23)
+        // %i - not specified
+        [TestCase("%I", ExpectedResult = "02")] // Hour in 12h format, zero-padded (01..12)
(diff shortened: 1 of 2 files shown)


microalps:
Just to open discussion we have Culture in comtext that can differ than current culture. Should we take the opportunity to correct that here?

src/DotLiquid.Tests/Util/StrFTimeTests.cs
```diff
@@ -10,39 +9,77 @@ namespace DotLiquid.Tests.Util
...
+        [TestCase("%V", ExpectedResult = "01")] // Week number of the week-based year (01..53)
+        [TestCase("%w", ExpectedResult = "0")] // Day of the week, Sunday is 0 (0..6)
+        [TestCase("%W", ExpectedResult = "01")] // Week number with the first Monday as the first day of week one (00..53)
+        [TestCase("%x", ExpectedResult = "01/08/12")] // Locale Date representation (08/23/01)
+        [TestCase("%X", ExpectedResult = "14:32:14")] // Locale Time representation (14:55:02)
+        [TestCase("%y", ExpectedResult = "12")] // Year, last two digits (00-99)
+        [TestCase("%Y", ExpectedResult = "2012")] // Year with century (2001)
+        [TestCase("%_2Y", ExpectedResult = "2012")] // Ensure space-pad does not truncate.
+        // %z, %:z, %::z --> See TestTimeZoneUTC/TestTimeZoneLocal
         [TestCase("%", ExpectedResult = "%")]
```
sebastienros, line 76: Isn't it `%%` that is supposed to render `%`?

sebastienros: Actually it's both, here is the result in ruby:

```ruby
puts d.strftime("%%x % %% %%% %%%% %%%%%")
```

```
%x % % %% %% %%%
```

src/DotLiquid/Util/StrFTime.cs
```diff
@@ -18,85 +20,197 @@ public static class StrFTime
...
             { "H", (dateTime) => dateTime.ToString("HH", CultureInfo.CurrentCulture) },
+            // i - not specified
             { "I", (dateTime) => dateTime.ToString("hh", CultureInfo.CurrentCulture) },
             { "j", (dateTime) => dateTime.DayOfYear.ToString().PadLeft(3, '0') },
-            { "m", (dateTime) => dateTime.ToString("MM", CultureInfo.CurrentCulture) },
+            // J - not specified
             { "k", (dateTime) => dateTime.ToString("%H", CultureInfo.CurrentCulture) },
-            { "l", (dateTime) => dateTime.ToString("%h", CultureInfo.CurrentCulture) },
+            // K - not specified
+            { "l", (dateTime) => dateTime.ToString("%h", CultureInfo.CurrentCulture).PadLeft(2, ' ') },
```
daviburg, line 38: "l" format is changed. The change should be behind an opt-in flag (use the compatibility version pattern from microalps).
Same for other formats you adjusted.

src/DotLiquid/Util/StrFTime.cs
```diff
@@ -18,85 +20,197 @@ public static class StrFTime
...
+            if (day >= DayOfWeek.Monday && day <= DayOfWeek.Wednesday)
+            {
+                dateTime = dateTime.AddDays(3);
             }
 
-            return output.ToString();
+            switch (directive)
+            {
+                case "G":
+                    return dateTime.ToString("yyyy");
```
daviburg, line 205: likely should have ", CultureInfo.CurrentCulture" or some explicit culture format specification.

src/DotLiquid/Util/StrFTime.cs
```diff
@@ -18,85 +20,197 @@ public static class StrFTime
...
+        /// <summary>
+        /// Applies formatting consistent to the rules specified by the Ruby Time.strftime function.
+        /// The following exceptions apply;
+        /// - `%N` (Not standard) - DateTime does not support nanoseconds precision, so the default is microsecond (6dp).
+        /// - `%9N` (Not implemented) - DateTime does not support nanoseconds precision.
+        /// - `%Z` (Not standard) - DotLiquid returns an offset (+00:00), Ruby states this should return a timezone description (e.g. 'GMT')
+        /// <see href="https://help.shopify.com/themes/liquid/filters/additional-filters#date"/>
+        /// <see href="https://ruby-doc.org/core-3.0.0/Time.html#method-i-strftime"/>
+        /// </summary>
+        /// <param name="dateTime"/>
```
daviburg, line 93: Best to keep param documentation consistent. Ideally provide a text description for each parameter. If we omit the text description, then at least we should be consistent on how we omit it (i.e. self-closing XML element vs explicit open-close element, vs no element at all)

src/DotLiquid/Util/StrFTime.cs
```diff
@@ -18,85 +20,197 @@ public static class StrFTime
...
+                x => StrFTimeMatchEvaluator(
+                    x.Groups[0].Value,
+                    x.Groups["flag"].Captures.Cast<Capture>().Select(y => y.Value).ToList(),
+                    x.Groups["width"].Captures.Cast<Capture>().Select(y => (int?)Convert.ToInt32(y.Value)).FirstOrDefault(),
+                    x.Groups["directive"].Captures.Cast<Capture>().Select(y => y.Value).FirstOrDefault(),
+                    dateTimeOffset
+                    ));
+        }
+
+        private static String StrFTimeMatchEvaluator(String orig, IEnumerable<String> flags, int? width, String directive, object obj)
```
daviburg, line 126: The concepts used here such as 'directive' may not be obvious. Best provide XML comment of what the method is doing and what the parameters are.

src/DotLiquid/Util/StrFTime.cs
```diff
@@ -18,85 +20,197 @@ public static class StrFTime
...
+                return orig; // This is an unspecified directive
+
+            return flags.ToList().Aggregate(result, (current, flag) => ApplyFlag(flag, (width ?? 2), current));
+        }
+
+        // Pre-process the directive for some of the quirkier cases, such as '%:z' and '%3N'.
+        private static String PreProcessDirective(String directive, IEnumerable<String> flags, int? width)
         {
-            var output = new System.Text.StringBuilder();
+            String result = directive;
```
daviburg, line 145: best use 'var' for obvious types.

src/DotLiquid/Util/StrFTime.cs
```diff
@@ -18,85 +20,197 @@ public static class StrFTime
...
+            String flagString = String.Join("", flags);
+            if ("z".Equals(directive) && ":".Equals(flagString))
+            {
+                result = flagString + directive;
```
daviburg, line 149: it may be a cleaner pattern to return directly the result rather than mutate the 'result' local variable.

daviburg (changes requested):

src/DotLiquid/Util/StrFTime.cs
microalps: Keeping bug fixes behind an opt-in flag just causes frustrations. If you were using "Ruby" date then you were familiar enough to know it was wrong and used it anyways without reporting a bug. This is different than Liquid filters which are/were barely documented and were not bugs but simply deviations from the reference implementation (many times not so obvious).

daviburg: ok

src/DotLiquid/Util/StrFTime.cs
robin-parker: All occurrences of DateTime.ToString() updated to include CultureInfo.CurrentCulture.

src/DotLiquid/Util/StrFTime.cs
robin-parker: Standardized all <param /> elements with descriptions.

src/DotLiquid/Util/StrFTime.cs
robin-parker: Added XML comment.

src/DotLiquid/Util/StrFTime.cs
robin-parker: Replaced local variable mutation with return statements.

src/DotLiquid/Util/StrFTime.cs
robin-parker: Done

robin-parker:
> Just to open discussion we have Culture in context that can differ than current culture. Should we take the opportunity to correct that here?

@microalps - I've updated all instances of DateTime.ToString() that did not already specify a FormatProvider to use CultureInfo.CurrentCulture for consistency.

src/DotLiquid/Util/StrFTime.cs
```diff
@@ -1,11 +1,16 @@
...
 using System.Collections.Generic;
 using System.Globalization;
+using System.Linq;
+using System.Text.RegularExpressions;
 
 namespace DotLiquid.Util
 {
     public static class StrFTime
     {
+        // A Regex capturing al suportied flags and directives
```
daviburg, line 11: 'all supported'

src/DotLiquid/Util/StrFTime.cs
```diff
@@ -16,87 +21,201 @@ public static class StrFTime
...
+        /// <summary>
+        /// Formats a date using a ruby date format string
+        /// </summary>
+        /// <param name="dateTimeOffset">date-time object to be formatted</param>
+        /// <param name="format">the required format</param>
+        /// <returns>a string version of date-time matching pattern.</returns>
+        public static string ToStrFTime(this DateTimeOffset dateTimeOffset, string format)
+        {
+            return Regex.Replace(input: format, pattern: PATTERN_REGEX,
+                evaluator: x => StrFTimeMatchEvaluator(
```
daviburg, line 120: consider using an English word instead of 'x' for input parameter. What is the content of 'x'?

src/DotLiquid/Util/StrFTime.cs
```diff
@@ -16,87 +21,201 @@ public static class StrFTime
...
+                    x.Groups[0].Value,
```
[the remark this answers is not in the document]
daviburg: As you have multiple groups, which one will be group 0?

src/DotLiquid/Util/StrFTime.cs
```diff
@@ -16,87 +21,201 @@ public static class StrFTime
...
+                    x.Groups["flag"].Captures.Cast<Capture>().Select(y => y.Value).ToList(),
```
daviburg, line 122: Consider declaring string constants for the names of the capture groups, use the same constant here and in the declaration of the Regex (use string interpolation then to insert the constants into the regex string)

src/DotLiquid/Util/StrFTime.cs
```diff
@@ -16,87 +21,201 @@ public static class StrFTime
...
+                    x.Groups["width"].Captures.Cast<Capture>().Select(y => (int?)Convert.ToInt32(y.Value)).FirstOrDefault(),
```
daviburg, line 123: rename 'y' to an English word describing what is contained in the variable.

src/DotLiquid/Util/StrFTime.cs
```diff
@@ -16,87 +21,201 @@ public static class StrFTime
...
+        /// <param name="width">optional minimum field width</param>
+        /// <param name="directive">The required data value, such as Y for a 4-digit year</param>
+        /// <param name="obj">the source date-time object</param>
+        /// </summary>
+        private static String StrFTimeMatchEvaluator(String orig, IEnumerable<String> flags, int? width, String directive, object obj)
+        {
+            var result = orig;
+            directive = PreProcessDirective(directive, flags, width);
+
+            bool isDateTimeOffset = obj is DateTimeOffset;
```
daviburg, line 146: use pattern matching

var isDateTimeOffset = obj is DateTimeOffset dateTimeOffset;

then you don't need to cast again when the pattern matches, instead use dateTimeOffset below

src/DotLiquid/Util/StrFTime.cs
```diff
@@ -16,87 +21,201 @@ public static class StrFTime
...
-        /// </summary>
-        /// <param name="dateTime"></param>
-        /// <param name="pattern"></param>
-        /// <returns></returns>
-        public static string ToStrFTime(this DateTimeOffset dateTime, string pattern)
+        // Pre-process the directive for some of the quirkier cases, such as '%:z' and '%3N'.
+        private static String PreProcessDirective(String directive, IEnumerable<String> flags, int? width)
         {
-            var output = new System.Text.StringBuilder();
+            var flagString = String.Join("", flags);
```
daviburg, line 160: String.Concat as not inserting any separator

src/DotLiquid/Util/StrFTime.cs
```diff
@@ -16,87 +21,201 @@ public static class StrFTime
...
+            if ("z".Equals(directive) && ":".Equals(flagString))
```
daviburg, line 161: Equals takes a comparison type. I think you want culture insensitive / ordinal comparison here. But case sensitive, right?

src/DotLiquid/Util/StrFTime.cs
robin-parker: Fixed

src/DotLiquid/Util/StrFTime.cs
```diff
@@ -16,87 +21,201 @@ public static class StrFTime
             { "b", (dateTime) => dateTime.ToString("MMM", CultureInfo.CurrentCulture) },
             { "B", (dateTime) => dateTime.ToString("MMMM", CultureInfo.CurrentCulture) },
             { "c", (dateTime) => dateTime.ToString("ddd MMM dd HH:mm:ss yyyy", CultureInfo.CurrentCulture) },
-            { "C", (dateTime) => ((int)Math.Floor(Convert.ToDouble(dateTime.ToString("yyyy"))/100)).ToString() },
+            { "C", (dateTime) => ((int)Math.Floor(Convert.ToDouble(dateTime.ToString("yyyy", CultureInfo.CurrentCulture))/100)).ToString() },
```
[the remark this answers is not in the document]
robin-parker: Added

src/DotLiquid/Util/StrFTime.cs
```diff
@@ -16,87 +21,201 @@ public static class StrFTime
...
+            { ":z", (dateTime) => dateTime.ToString("%K", CultureInfo.CurrentCulture) },
+            // ::z - not implemented
             { "Z", (dateTime) => dateTime.ToString("zzz", CultureInfo.CurrentCulture) },
-            { "%", (dateTime) => "%" }
+            { "%", (dateTime) => "%" } // A % sign
         };
 
         private static readonly Dictionary<string, DateTimeOffsetDelegate> OffsetFormats = new Dictionary<string, DateTimeOffsetDelegate>
         {
             { "s", (dateTimeOffset) => ((long)(dateTimeOffset - new DateTimeOffset(1970, 1, 1, 0,0,0, TimeSpan.Zero)).TotalSeconds).ToString() },
```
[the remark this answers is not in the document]
robin-parker: Done

src/DotLiquid/Util/StrFTime.cs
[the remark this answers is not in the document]
robin-parker: Added named parameters.
Groups[0] represent the full specifier, for example %s

src/DotLiquid/Util/StrFTime.cs
robin-parker: Changed to specifier.

src/DotLiquid/Util/StrFTime.cs
robin-parker: Done.

src/DotLiquid/Util/StrFTime.cs
robin-parker: Done

src/DotLiquid/Util/StrFTime.cs
robin-parker: This triggered a CS0165 compiler error for netstandard2.0, to resolve I created a separate named pattern in the if and else blocks.

src/DotLiquid/Util/StrFTime.cs
robin-parker: Changed

src/DotLiquid/Util/StrFTime.cs
robin-parker: Done

src/DotLiquid/Util/StrFTime.cs
```diff
@@ -16,87 +26,199 @@ public static class StrFTime
...
+        /// Formats a date-time for a single specifier within the requested format. For ease of processing
+        /// the specifier is broken down into it's consituent parts of: flags, width, directive.
+        ///
+        /// A specifier consists of a percent (%) character, zero or more flags, optional minimum field width and a conversion directive as follows:
+        /// <param name="specifier">the entire specifier, such as %Y for a 4-digit year</param>
+        /// <param name="flags">zero or more flags to change formatting, such as '0' (zero padding).</param>
+        /// <param name="width">optional minimum field width</param>
+        /// <param name="directive">The required data value, such as Y for a 4-digit year</param>
+        /// <param name="source">the source date-time object</param>
+        /// </summary>
```
daviburg, line 144: Misplaced end tag `/// </summary>`
The end tag for summary must be above the parameter elements

daviburg: Everything else looks good.

daviburg (approved):

PR #436, opened 2021-06-15
Fix ArgumentNullException upon calling Regex.Match() due to invalid variable name

WouterJanson:
Encountered this while moving out templates from Mustache to Liquid, a if statement can be written as follow in Mustache:
```Mustache
{{#person}}
  {{.}}
{{/person}}
```
Where `{{.}}` refers to the person variable itself, so the Liquid equivalent would be:
```Liquid
{% if person %}
  {{ person }}
{% endif %}
```

Obviously it's not gonna work when the Mustache template is being processed by dotliquid, but if you would have `{{.}}` in your liquid template you'd get an `ArgumentNullException` when the library calls:
```csharp
Match firstPartSquareBracketedMatch = SquareBracketedRegex.Match(firstPart);
```

This is because a null is passed into `.Match()` due to the parts list having a length of zero.

---

To fix this, I've added an explicit null check on the variable `firstPart` and return a `VariableNotFoundException` if requested. This should result in a clearer error message to the end user, in order to fix their template.

### DIFF
--- a/src/DotLiquid.Tests/ContextTests.cs
+++ b/src/DotLiquid.Tests/ContextTests.cs
@@ -279,6 +279,17 @@ namespace DotLiquid.Tests
         }
 
         [Test]
+        public void TestVariableNotFoundExceptionFromInvalidVariableName()
+        {
+            Template template = Template.Parse("{{ . }}");
+            string rendered = template.Render();
+
+            Assert.AreEqual("", rendered);
+            Assert.AreEqual(1, template.Errors.Count);
+            Assert.AreEqual(string.Format(Liquid.ResourceManager.GetString("VariableNotFoundException"), "."), template.Errors[0].Message);
+        }
+
+        [Test]
         public void TestScoping()
         {
             Assert.DoesNotThrow(() =>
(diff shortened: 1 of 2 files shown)


src/DotLiquid.Tests/ContextTests.cs
```diff
@@ -278,6 +278,17 @@ public void TestVariableNotFoundExceptionIgnoredForUnlessStatement()
...
 
+        [Test]
+        public void TestVariableNotFoundExceptionFromInvalidVariableName()
+        {
+            Template template = Template.Parse("{{ . }}");
+            string rendered = template.Render();
+
+            Assert.AreEqual("", rendered);
+            Assert.AreEqual(1, template.Errors.Count);
+            Assert.AreEqual(string.Format(Liquid.ResourceManager.GetString("VariableNotFoundException"), "."), template.Errors[0].Message);
```
microalps, line 289: Might want to follow the model of checking for exception similar to TestHashWithDefaultProc. This is checking a string output which is good but lacking type checking.

microalps (commented):
What is the behavior on other incomplete notations such as "variable." Or "variable[index"?

src/DotLiquid.Tests/ContextTests.cs
microalps: I actually started looking into this - I see this has a different behavior and your test is valid.

microalps:
Thank you @WouterJanson for your contribution. This, together with additional changes and tests, have been incorporated into #441. Closing this PR. Feel free to continue discussion there.

PR #437, opened 2021-06-17
Concat, EscapeOnce and Reverse filters (#330)

### LINKED ISSUE #350: ExpandoObject is not resolving
### Dotliquid version
2.0.314.0

### Expected behavior
```cs
public class ExpandoModel
{
    public int IntProp { get; set; }
    public string StrProp { get; set; }
    public ExpandoObject Props { get; set; }
}
private static void RunUsingExpando()
{
    const string templateString = @"Int: '{{IntProp}}'; String: '{{StrProp}}'; Expando: '{{Props.Key1}}'";
    Template.NamingConvention = new CSharpNamingConvention();
    Template.RegisterSafeType(typeof(ExpandoModel), new[] { "IntProp", "StrProp", "Props" });
    var t = Template.Parse(templateString);
    var model = new ExpandoModel()
    {
        IntProp = 23,
        StrProp = "from string prop",
        Props = new ExpandoObject()
    };
    var dic = (IDictionary<string, object>)model.Props;
    dic.Add("Key1", "from expando prop");

    string output = t.Render(Hash.FromAnonymousObject(model));
    Console.WriteLine("RunUsingExpando --> " + output);
}
```

**Expected Output:**
>Int: '23'; String: 'from string prop'; Expando: 'from expando prop'

### Actual behavior
**Actual Output:**
>Int: '23'; String: 'from string prop'; Expando: ''

### Observations
If I replace 
```public ExpandoObject Props { get; set; }```
with 
```public Dictionary<string, object> Props { get; set; }``` <-- this works

### LINKED ISSUE #417: Cannot bind property of an ExpandoObject that is inside an array
Looks like we are unable bind a property of an ExpandoObject when that object is inside an array and when we use `Hash.FromDictionary`

[NETfiddle Demo](https://dotnetfiddle.net/Ky9s0x)

`
   
            public class Program
            {
                public static void Main()
                {
                    // this does not work		
                    var modelString = "{\"States\": [{\"Name\": \"Texas\",\"Code\": \"TX\"}, {\"Name\": \"New York\",\"Code\": \"NY\"}]}";
                    var template = "State Is:{{States[0].Name}}";
                    RenderFromDictionary(modelString, template);

                    //this works
                    modelString = "{\"States\": [\"Texas\",\"New York\"]}";
                    template = "State Is:{{States[0]}}";
                    RenderFromDictionary(modelString, template);

                    //this works
                    var anonymoudmodel = new { States = new[] { new { Name = "Texas", Code = "TX" }, new { Name = "New York", Code = "NY" } } };
                    template = "State Is:{{States[0].Name}}";
                    RenderFromAnonymousObject(anonymoudmodel, template);
                }

                private static void RenderFromDictionary(string modelString, string template)
                {
                    var model = JsonConvert.DeserializeObject<ExpandoObject>(modelString);
                    var templateModel = Hash.FromDictionary(model);
                    var html = Template.Parse(template).Render(templateModel);
                    Console.WriteLine(html);
                }

                private static void RenderFromAnonymousObject(dynamic model, string template)
                {
                    var templateModel = Hash.FromAnonymousObject(model);
                    var html = Template.Parse(template).Render(templateModel);
                    Console.WriteLine(html);
                }
            }

`

### PR DESCRIPTION
**Pull request by robin-parker**
- Implements some of the missing standard filters listed in issue #330.
  - Doesn't cover `Sort` or `SortNatural`, which are in PR #431.
  - Doesn't cover `Echo` or `Liquid`, which will move to #438 
- Enhanced Split filter to 'split a string into an array' as mentioned in the [reverse filter documentation](https://shopify.github.io/liquid/filters/reverse/).
- Improvements to ExpandoObject evaluation, resolves #350, resolves #417
  - Removed need and usage of recursive Hash.FromDictionary
  - Replaced inner/nested Hash in `for` and `cycle` tags with generic Dictionary.

### CODE REVIEW

**Comment by microalps** [comment]
Can we have the original description updated to note what is included? I see changes to split - which has no open PR. Also, for resolving multiple tickets, github requires you to write resolves #xxx, resolves #yyy a simple comma doesn't denote resolving two issues.

File: src/DotLiquid/StandardFilters.cs [hunk 1]
```diff
@@ -224,6 +240,18 @@ public static string TruncateWords(string input, int words = 15, string truncate
         /// <returns></returns>
         public static string[] Split(string input, string pattern)
         {
+            // If the pattern is empty and the input is a string convert to an array ...
+            // ... as specified in the Liquid Reverse filter example.
+            if (string.IsNullOrEmpty(pattern) && !string.IsNullOrEmpty(input))
+            {
+                var stringList = new List<string>();
```
sebastienros, line 247: Less allocations if you create a `char[]` with the correct size, and then use a for loop to iterate the chars.

src/DotLiquid/Context.cs
```diff
@@ -685,7 +694,7 @@ private static object Liquidize(object obj)
 
             if (obj.GetType().GetTypeInfo().GetCustomAttributes(typeof(LiquidTypeAttribute), false).Any())
             {
-                var attr = (LiquidTypeAttribute) obj.GetType().GetTypeInfo().GetCustomAttributes(typeof(LiquidTypeAttribute), false).First();
+                var attr = (LiquidTypeAttribute)obj.GetType().GetTypeInfo().GetCustomAttributes(typeof(LiquidTypeAttribute), false).First();
```
sebastienros, line 697: Probably faster (perf) to use `FirstOrDefault` and check for null instead of call `GetCustomAttributes()` twice. Even with the internal caching the resulting array is recreated.

src/DotLiquid.Tests/ContextTests.cs
```diff
@@ -821,11 +870,11 @@ public void TestArrayContainingLambdaAsVariable()
...
-            _context["callcount"] = (Proc) (c =>
-            {
-                ++global;
-                return global.ToString();
-            });
+            _context["callcount"] = (Proc)(c =>
+           {
+               ++global;
+               return global.ToString();
+           });
```
microalps, line 877: Seems to be incorrect change - 873 has 12 spaces and 874 has 11.

microalps (changes requested):

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -222,6 +253,48 @@ public void TestSort_OnDictionaryWithPropertyOnlyInSomeElement_ReturnsSortedDict
...
+
+  <h4>Little black dress</h4>
+",
+                template: @"{% assign products_by_company = collection.products | sort_natural: 'company' %}
+{% for product in products_by_company %}
+  <h4>{{ product.title }}</h4>
+{% endfor %}",
+                localVariables: Hash.FromAnonymousObject(new { collection }));
+        }
+
```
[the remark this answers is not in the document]
robin-parker: Removed

src/DotLiquid/StandardFilters.cs
robin-parker: Replaced by code from microalps.

src/DotLiquid/StandardFilters.cs
```diff
@@ -224,6 +240,18 @@ public static string TruncateWords(string input, int words = 15, string truncate
...
+            if (string.IsNullOrEmpty(pattern) && !string.IsNullOrEmpty(input))
+            {
+                var stringList = new List<string>();
+                foreach (var @char in input.ToCharArray())
+                {
+                    stringList.Add(@char.ToString());
+                }
+                return stringList.ToArray();
+            }
+
```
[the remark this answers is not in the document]
robin-parker: Thanks, I've made the recommended changes.

src/DotLiquid.Tests/ContextTests.cs
robin-parker: Fixed.

src/DotLiquid/Context.cs
```diff
@@ -585,23 +589,28 @@ private object LookupAndEvaluate(object obj, object key)
...
             else if (obj is IIndexable indexableObj)
-            { 
+            {
                 value = indexableObj[key];
             }
             else
-            { 
+            {
                 throw new NotSupportedException();
             }
```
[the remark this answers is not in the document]
robin-parker: I didn't change these lines, they were cleaned up by the code formatter.

src/DotLiquid/Context.cs
robin-parker: I didn't change this line, it was cleaned up by the code formatter.

src/DotLiquid/Context.cs
```diff
@@ -585,23 +589,28 @@ private object LookupAndEvaluate(object obj, object key)
...
             if (obj is IDictionary dictionaryObj)
-            { 
+            {
                 value = dictionaryObj[key];
             }
+            // Resolve #350/#417, add support for rendering of a nested  ExpandoObject
+            else if (obj is IDictionary<string, object> dictionaryObject)
+            {
+                value = dictionaryObject[key.ToString()];
+            }
```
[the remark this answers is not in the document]
robin-parker: I didn't know rendering performance was an issue, I'll take a look at whether there is sufficient test coverage to verify refactoring and get back to you.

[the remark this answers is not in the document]
robin-parker: Combined IsHashOrArrayLikeObject/LookupAndEvaluate methods and updated Hash.Contains() as agreed.

src/DotLiquid/Context.cs
```diff
@@ -430,14 +430,15 @@ private object FindVariable(string key)
...
-                    if ((variable = LookupAndEvaluate(e, key)) != null)
+                    if (TryEvaluateHashOrArrayLikeObject(e, key, out variable))
                     {
                         scope = e;
                         break;
                     }
             }
             scope = scope ?? Environments.LastOrDefault() ?? Scopes.Last();
-            variable = variable ?? LookupAndEvaluate(scope, key);
+            if (variable == null)
```
microalps, line 440: By not saving result of Try, a explicit null value in a dictionary will recompute. This is inefficient. See my PR how I rewrote the logic here, but feel free to offer another solution.

src/DotLiquid/Context.cs
```diff
@@ -510,38 +511,12 @@ private object Variable(string markup, bool notifyNotFound)
...
+                    if (TryEvaluateHashOrArrayLikeObject(@object, part, out res))
                     {
                         // If its a proc we will replace the entry with the proc
-                        object res = LookupAndEvaluate(@object, part);
                         @object = Liquidize(res);
                     }
-                    // Some special cases. If the part wasn't in square brackets and
-                    // no key with the same name was found we interpret following calls
-                    // as commands and call them on the current object
-                    else if (!partResolved && (@object is IEnumerable) && (Template.NamingConvention.OperatorEquals(part as string, "size") || Template.NamingConvention.OperatorEquals(part as string, "first") || Template.NamingConvention.OperatorEquals(part as string, "last")))
```
microalps, line 529: This logic is incorrectly removed. We should not compute first/last from square brackets. PartResolved is true if it matched square brackets. Also, the three other places using TryEvaluateHashOrArrayLikeObject now are computing first last when previously did not. This is a behavior change that seems unnecessary.

src/DotLiquid/Context.cs
```diff
@@ -561,73 +536,63 @@ private object Variable(string markup, bool notifyNotFound)
...
-                { 
                     throw new NotSupportedException();
-                }
-                return newValue;
+
+                value = newValue;
             }
 
-            return value;
+            return value != null;
```
microalps, line 595: Value can be explicitly nuull and should return true.

src/DotLiquid/Context.cs
```diff
@@ -723,7 +688,9 @@ private void SquashInstanceAssignsWithEnvironments()
                 foreach (Hash env in Environments)
                     if (env.ContainsKey(k))
                     {
-                        tempAssigns[k] = LookupAndEvaluate(env, k);
+                        object value = null;
+                        TryEvaluateHashOrArrayLikeObject(env, k, out value);
```
microalps, line 692: Try out var

src/DotLiquid/Hash.cs
```diff
@@ -240,22 +241,22 @@ public void Clear()
...
         IDictionaryEnumerator IDictionary.GetEnumerator()
         {
-            return ((IDictionary) _nestedDictionary).GetEnumerator();
+            return ((IDictionary)_nestedDictionary).GetEnumerator();
         }
 
         public bool Contains(KeyValuePair<string, object> item)
         {
-            return ((IDictionary<string, object>) _nestedDictionary).Contains(item);
+            return _defaultValue != null || _lambda != null || ((IDictionary<string, object>)_nestedDictionary).Contains(item);
```
microalps, line 249: This change doesnt seem correct. It doesn't contain that key/value pair only that key especially is defaultValue is not equal to the value of kvp

src/DotLiquid/Hash.cs
```diff
@@ -274,32 +275,32 @@ public int Count
...
         public bool IsFixedSize
         {
-            get { return ((IDictionary) _nestedDictionary).IsFixedSize; }
+            get { return ((IDictionary)_nestedDictionary).IsFixedSize; }
         }
 
         public bool ContainsKey(string key)
         {
-            return _nestedDictionary.ContainsKey(key);
+            return Contains(key);
```
microalps, line 303: Needs research why we have two alias methods. I only changed and tested this one in my commit. Maybe we can mark the other obsolete?

src/DotLiquid.Tests/ContextTests.cs
```diff
@@ -531,14 +536,22 @@ public void TestAccessHashesWithHashAccessVariables()
...
+        // See https://github.com/Shopify/liquid/blob/50d1a2ffc91e69c26af2eea278cbc9c4930126a9/test/integration/drop_test.rb#L232
         public void TestHashNotationOnlyForHashAccess()
         {
             _context["array"] = new[] { 1, 2, 3, 4, 5 };
-            _context["hash"] = new { first = "Hello" };
+            _context["hash"] = new { last = "goodbye", first = "Hello" };
+            _context["emptyArray"] = new int[0];
 
             Assert.AreEqual(1, _context["array.first"]);
-            Assert.AreEqual(null, _context["array['first']"]);
```
microalps, line 540: See my other comments this behavior should not be changed. Is there a spec or test case in Liquid that says this is valid syntax?

robin-parker: Does this [test](https://github.com/Shopify/liquid/blob/50d1a2ffc91e69c26af2eea278cbc9c4930126a9/test/integration/drop_test.rb#L232.) mean array["first"] is supported by Ruby?

src/DotLiquid/Hash.cs
robin-parker: The Contains/ContainsKey methods are for different interfaces implemented by Hash:
- `System.Collections.Generic.IDictionary<string, object>` declares `ContainsKey(string)`
- `System.Collections.IDictionary` declares `Contains(object) `

src/DotLiquid.Tests/ContextTests.cs
microalps: Seems so. I will confirm on an actual ruby install, but in the interim please note this change in the description of issues fixed and mark this resolved.

src/DotLiquid/Context.cs
microalps: As noted elsewhere this can be marked resolved

src/DotLiquid/Context.cs
robin-parker: Incorporated changes from @microalps dotliquid/render-perf branch.

src/DotLiquid/Context.cs
robin-parker: Done

src/DotLiquid/Hash.cs
robin-parker: Removed _defaultValue!=null check for all Contain(...) methods.

src/DotLiquid/Context.cs
robin-parker: Done

src/DotLiquid.Tests/ContextTests.cs
microalps: See test_hash_notation_only_for_hash_access - https://github.com/Shopify/liquid/blob/master/test/integration/context_test.rb this is not valid after all. I'm not sure what that other test is, but a quick check against my ruby install confirms array.first != array["first"]

src/DotLiquid/Context.cs
microalps: See my other comment - this change is not consistent with reference Ruby liquid.

microalps:
My performance changes have been separated into #441 - please roll back to before all perf changes and I will rebase my commit for perf after it merges

src/DotLiquid/Context.cs
robin-parker: Changes reverted

src/DotLiquid/Hash.cs
robin-parker: Change reverted, new tests retained to document pre-existing behaviour.

src/DotLiquid.Tests/ContextTests.cs
robin-parker: Reverted.

robin-parker:
> My performance changes have been separated into #441 - please roll back to before all perf changes and I will rebase my commit for perf after it merges

@microalps, I have reverted all changes except for a few new tests in HashTests which document the existing Default/Lambda constructor behaviour.

src/DotLiquid.Tests/HashTests.cs
```diff
@@ -117,5 +118,50 @@ public void TestUpperTwoScenarioWithSameClass()
             IncludeBaseClassPropertiesOrNot(true);
         }
         #endregion
+
+        [Test]
+        public void TestDefaultConstructor()
```
microalps, line 123: TestDefaultValueConstructor might be more accurate.

src/DotLiquid.Tests/HashTests.cs
```diff
@@ -117,5 +118,50 @@ public void TestUpperTwoScenarioWithSameClass()
...
+        {
+            var hash = new Hash(0); // default value of zero
+            hash["key"] = "value";
+
+            Assert.False(hash.Contains("unknown-key"));
+            Assert.False(hash.ContainsKey("unknown-key"));
```
microalps, line 129: For clarification - this is just temporarily false but behavior change will be introduced in #441

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -1572,5 +1603,80 @@ public void TestWhere_NonStringCompare()
...
+        public void TestConcat()
+        {
+            var array1 = new String[] { "one", "two" };
+            var array2 = new String[] { "alpha", "bravo" };
+            var arrayCombined = new String[] { "one", "two", "alpha", "bravo" };
+
+            Assert.AreEqual(null, StandardFilters.Concat(null, null));
+            Assert.AreEqual(array1, StandardFilters.Concat(array1, null));
+            Assert.AreEqual(array1, StandardFilters.Concat(null, array1));
+            Assert.AreEqual(arrayCombined, StandardFilters.Concat(array1, array2));
```
microalps, line 1617: Should IEnumerable return be tested by CollectionAssert.AreEqual?

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -1572,5 +1603,80 @@ public void TestWhere_NonStringCompare()
...
+
+        [Test]
+        public void TestReverse()
+        {
+            var array = new String[] { "one", "two", "three" };
+            var arrayReversed = new String[] { "three", "two", "one" };
+
+            Assert.AreEqual(null, StandardFilters.Reverse(null));
+            Assert.AreEqual(arrayReversed, StandardFilters.Reverse(array));
+            Assert.AreEqual(array, StandardFilters.Reverse(arrayReversed));
```
microalps, line 1655: Should IEnumerable return be tested by CollectionAssert.AreEqual?

microalps (changes requested):
Looks pretty good to me, now for @daviburg to take a stab

src/DotLiquid.Tests/HashTests.cs
robin-parker: Agreed.

src/DotLiquid.Tests/HashTests.cs
robin-parker: Yes, I had to change from true to false whilst reverting perf changes.

src/DotLiquid.Tests/StandardFilterTests.cs
robin-parker: Changed to CollectionAssert.AreEqual.

src/DotLiquid.Tests/StandardFilterTests.cs
robin-parker: CollectionAssert.AreEqual

src/DotLiquid.Tests/ContextTests.cs
```diff
@@ -785,34 +790,78 @@ public void TestNestedDynamicAsVariable()
...
+        [Test]
+        public void TestNestedExpandoTemplate_Issue350()
+        {
+            var model = new ExpandoModel()
+            {
+                IntProp = 23,
+                StrProp = "from string prop",
+                Props = new ExpandoObject()
+            };
+            var dic = (IDictionary<string, object>)model.Props;
```
daviburg, line 810: Avoid uncommon abbreviations and use fully-form words from the English dictionary for code read-ability.

src/DotLiquid.Tests/ContextTests.cs
```diff
@@ -785,34 +790,78 @@ public void TestNestedDynamicAsVariable()
...
+            dic.Add("Key1", "from expando prop");
```
daviburg, line 811: Nit: when two or more arguments have the same type, it is safer to specify the argument names to remove the ambiguity for human reader of the code (compiler won't care, but compiler is not capable to say if the order was invalid).

src/DotLiquid.Tests/ContextTests.cs
```diff
@@ -785,34 +790,78 @@ public void TestNestedDynamicAsVariable()
...
+
+            Template.RegisterSafeType(typeof(ExpandoModel), new[] { "IntProp", "StrProp", "Props" });
+            const string templateString = @"Int: '{{IntProp}}'; String: '{{StrProp}}'; Expando: '{{Props.Key1}}'";
+            var t = Template.Parse(templateString);
```
daviburg, line 815: Same for single letter variables. Just don't use them. Always use a word from the English dictionary that describes the purpose of the variable.

daviburg: Note that there might be older code that uses ambiguous names. Don't copy that :-)

src/DotLiquid.Tests/ContextTests.cs
```diff
@@ -785,34 +790,78 @@ public void TestNestedDynamicAsVariable()
             Assert.AreEqual("Hello", _context["dynamic.lambda.name"]);
         }
 
+        private class ExpandoModel
```
daviburg, line 793: Consider grouping related privates and test methods in a separate class to avoid this test class from growing very large and difficult to maintain.

src/DotLiquid.Tests/ContextTests.cs
```diff
@@ -785,34 +790,78 @@ public void TestNestedDynamicAsVariable()
...
+            Assert.AreEqual(expected: "Int: '23'; String: 'from string prop'; Expando: 'from expando prop'",
+                            actual: t.Render(Hash.FromAnonymousObject(model)));
+        }
+
+        private class DataJson : Drop
+        {
+            public Dictionary<string, object> Data { get; set; }
+        }
+
+        // Test case for [Issue #417](https://github.com/dotliquid/dotliquid/issues/417)
```
daviburg, line 825: Nit: non-standard comment format at this position. `//` should be used for in-method comments above a line of code. Above a method the XML comment format `/// <summary>...` should be used.

src/DotLiquid.Tests/HashTests.cs
daviburg: Please add a code comment `//` above so future maintainers know if/that it is ok to change.

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -1572,5 +1603,81 @@ public void TestWhere_NonStringCompare()
...
+            var arrayReversed = new String[] { "three", "two", "one" };
+
+            CollectionAssert.AreEqual(null, StandardFilters.Reverse(null));
+            CollectionAssert.AreEqual(arrayReversed, StandardFilters.Reverse(array));
+            CollectionAssert.AreEqual(array, StandardFilters.Reverse(arrayReversed));
+            CollectionAssert.AreEqual(new[] { 3, 2, 2, 1 }, StandardFilters.Reverse(new[] { 1, 2, 2, 3 }));
+        }
+
+        /// <summary>
+        /// Reverses the order of the items in an array. reverse cannot reverse a string.
```
daviburg, line 1660: Should you have a test case for calling reverse on the string? While this shouldn't work, it is important that we have explicit test as some users will attempt that before they discover they need to first split, then reverse, then join.

src/DotLiquid/Context.cs
```diff
@@ -585,23 +589,28 @@ private object LookupAndEvaluate(object obj, object key)
         {
             object value;
             if (obj is IDictionary dictionaryObj)
-            { 
+            {
                 value = dictionaryObj[key];
             }
+            // Resolve #350/#417, add support for rendering of a nested  ExpandoObject
```
daviburg, line 595: nit: duplicate space character after 'nested'

src/DotLiquid/StandardFilters.cs
```diff
@@ -155,6 +156,21 @@ public static string Escape(string input)
...
+        /// <summary>
+        /// Escapes a string without changing existing escaped entities.
+        /// It doesn’t change strings that don’t have anything to escape.
+        /// </summary>
+        /// <param name="input">String to escape</param>
+        /// <returns>Escaped string</returns>
+        /// <see href="https://shopify.github.io/liquid/filters/escape_once/"/>
+        public static string EscapeOnce(string input)
+        {
+            if (string.IsNullOrEmpty(input))
```
daviburg, line 168: Use a ternary single line return statement. i.e. return condition ? a:b;

src/DotLiquid/StandardFilters.cs
```diff
@@ -218,14 +234,21 @@ public static string TruncateWords(string input, int words = 15, string truncate
 
         /// <summary>
         /// Split input string into an array of substrings separated by given pattern.
+        ///
+        /// If the pattern is empty the input string is converted to an array of 1-char
```
daviburg, line 238: Add this in a `remarks` XML element after the summary element

src/DotLiquid/StandardFilters.cs
```diff
@@ -218,14 +234,21 @@ public static string TruncateWords(string input, int words = 15, string truncate
...
         /// <param name="pattern"></param>
         /// <returns></returns>
         public static string[] Split(string input, string pattern)
         {
-            return input.IsNullOrWhiteSpace()
-                ? new[] { input }
+            if (input.IsNullOrWhiteSpace())
+                return new[] { input };
+
+            // If the pattern is empty convert to an array as specified in the Liquid Reverse filter example.
```
daviburg, line 249: of the Shopify specification. There are typically two if not more Shopify pages for each filter, so consider adding the exacting example URL.

src/DotLiquid/StandardFilters.cs
```diff
@@ -976,6 +999,46 @@ private static bool HasMatchingProperty(this object any, string propertyName, ob
                 ? propertyValue.IsTruthy()
                 : propertyValue.SafeTypeInsensitiveEqual(targetValue);
         }
+
+        /// <summary>
+        /// Concatenates (joins together) multiple arrays.
+        /// The resulting array contains all the items from the input arrays.
+        ///
+        /// Will not remove duplicate entries from the concatenated array
```
daviburg, line 1007: move to remarks element

src/DotLiquid/StandardFilters.cs
```diff
@@ -976,6 +999,46 @@ private static bool HasMatchingProperty(this object any, string propertyName, ob
...
+        /// <summary>
+        /// Concatenates (joins together) multiple arrays.
+        /// The resulting array contains all the items from the input arrays.
+        ///
+        /// Will not remove duplicate entries from the concatenated array
+        /// unless you also use the uniq filter.
+        /// </summary>
+        /// <param name="left">left hand (start) of the new concatenated array</param>
+        /// <param name="right">array to be appended to left</param>
+        /// <returns></returns>
```
daviburg, line 1012: omit empty XML comment elements

src/DotLiquid/StandardFilters.cs
```diff
@@ -976,6 +999,46 @@ private static bool HasMatchingProperty(this object any, string propertyName, ob
...
+        /// <see href="https://shopify.github.io/liquid/filters/concat/"/>
+        public static IEnumerable Concat(IEnumerable left, IEnumerable right)
+        {
+            // If either side is null, return the other side.
+            if (left == null)
+                return right;
+            else if (right == null)
+                return left;
+
+            List<object> leftList = left.Cast<object>().ToList();
```
daviburg, line 1022: likely you can use `var` here and on next line. And as the variables are used only once, you should consider just inlining the whole thing.

daviburg (changes requested):

daviburg:
> Looks pretty good to me, now for @daviburg to take a stab

Thanks for tagging me.

src/DotLiquid.Tests/ContextTests.cs
microalps: Davi, give it a break, it's clearly a key/value and only one signature exists. Agree with you in general, not in this case.

daviburg: Yes that's fine here.

src/DotLiquid/Context.cs
```diff
@@ -585,23 +589,28 @@ private object LookupAndEvaluate(object obj, object key)
         {
             object value;
             if (obj is IDictionary dictionaryObj)
-            { 
+            {
                 value = dictionaryObj[key];
             }
+            // Resolve #350/#417, add support for rendering of a nested  ExpandoObject
+            else if (obj is IDictionary<string, object> dictionaryObject)
```
microalps, line 596: Does this code replace the need for similar type checking and recursive Hash.FromDictionary calls?

src/DotLiquid.Tests/HashTests.cs
robin-parker: Comment added.

src/DotLiquid.Tests/ContextTests.cs
robin-parker: Fixed, I blindly copied this test from issue #350 without cleaning it up.

src/DotLiquid.Tests/ContextTests.cs
robin-parker: Fixed.

src/DotLiquid.Tests/StandardFilterTests.cs
robin-parker: Added test case.

src/DotLiquid/Context.cs
robin-parker: Removed

src/DotLiquid/StandardFilters.cs
robin-parker: Done

src/DotLiquid/StandardFilters.cs
robin-parker: Done.

src/DotLiquid/StandardFilters.cs
robin-parker: Added URL to reverse filter.

src/DotLiquid/StandardFilters.cs
robin-parker: Done

src/DotLiquid/StandardFilters.cs
robin-parker: Removed

src/DotLiquid/StandardFilters.cs
robin-parker: Inlined.

src/DotLiquid.Tests/ContextTests.cs
robin-parker: Removed one unused private class and relocated the other to test class region.

src/DotLiquid/Context.cs
robin-parker: I can take a look for recursive Hash.FromDictionary calls, did you have any specific locations in mind using similar type checking?

microalps: I am referring to this block:
```
            foreach (var keyValue in dictionary)
            {
                    if (keyValue.Value is IDictionary<string, object> innerDictionary)
                    {
                        result.Add(keyValue.Key, FromDictionary(innerDictionary));
                    }
                    else
                    {
                        result.Add(keyValue);
                    }
            }
```
It can be simplified after your change:
```
            foreach (var keyValue in dictionary)
            {
                    result.Add(keyValue);
            }
```

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -1654,6 +1654,7 @@ public void TestReverse()
             CollectionAssert.AreEqual(arrayReversed, StandardFilters.Reverse(array));
             CollectionAssert.AreEqual(array, StandardFilters.Reverse(arrayReversed));
             CollectionAssert.AreEqual(new[] { 3, 2, 2, 1 }, StandardFilters.Reverse(new[] { 1, 2, 2, 3 }));
+            Assert.AreEqual(".moT rojaM ot lortnoc dnuorG", StandardFilters.Reverse("Ground control to Major Tom."));
```
microalps, line 1657: This is incorrect outcome according to the [docs](https://shopify.github.io/liquid/filters/reverse/):
> reverse cannot be used directly on a string
The docs do not note what SHOULD be the outcome, but a quick Liquid test shows no change and no exception.

robin-parker: Its a simple change to faithfully implement the reverse filter description, but it seems odd to add code to remove potentially useful behaviour.
Should we prioritise compliance with the docs over functionality? (I’d vote for keeping string reversal).

robin-parker: Changed to prevent string reversal.

src/DotLiquid/Hash.cs
```diff
@@ -165,10 +152,14 @@ public Hash()
 
         #endregion
 
-        public void Merge(IDictionary<string, object> otherValues)
+        /// <summary>
+        /// Merge the contents of the given Dicionary into this Hash.
+        /// <summary>
+        public Hash Merge(IDictionary<string, object> otherValues)
```
microalps, line 158: This is a breaking change on a public method. There is no reason for this change other than skimping on two lines.

microalps (changes requested):

src/DotLiquid/Hash.cs
robin-parker: What does a change of return type from void to Hash break?
My intention is to enable fluent use of Merge, for example:
`var merged = Hash.FromDictionary(dictionary1).Merge(dictionary2);`

src/DotLiquid/Context.cs
robin-parker: @microalps, whilst looking at this I've found what I think is a pre-existing bug (#442) in for loops when a Hash is created with nested Dictionary objects.
I'll submit a separate PR for that bug fix.

daviburg: @microalps ok to resolve this conversation here then?

src/DotLiquid/Hash.cs
daviburg: If the developer rebuilds the code it should be fine. But if there is just an assembly redirect to a new version, I am not sure the method resolution at runtime will still work

src/DotLiquid/Context.cs
microalps: Resolves #442

src/DotLiquid/Hash.cs
microalps: I requested change because (a) I'm not sure like @daviburg says and (b) there was no reason for it - nothing in our code is fluent style (c) if we are chaining a constructor with a merge - that just means we are missing a constructor with a dictionary (See https://github.com/dotliquid/dotliquid/pull/441/files#diff-fc52e9cceed37df79819e1e24e3657a1e091af0d2a293047135da7a2b991f4f9R163 where I did just that). FYI - DataSet Merge returns void as do pretty much every method in .NET with the name merge.

robin-parker: I've reverted the Merge changes. Also FWIW, I don't think these changes will resolve #442.

Is there a guide anywhere I can check to avoid future style conflicts?

daviburg: I don't think the dotLiquid project has yet a documented code style. But that would be a good thing to start now that we have several concurrently active contributors (yah!).

microalps (approved):

daviburg (commented):
Only code style comments left.

daviburg (approved):

PR #440, opened 2021-06-29
Raw-Type End Tag with Whitespace Control Char

### LINKED ISSUE #439: comment tag was never closed
### Dotliquid version
2.1.457

### Expected behavior
Liquid template parse successfully

### Actual behavior
Liquid template parse failed with `SyntaxException`: 

```
DotLiquid.Exceptions.SyntaxException : comment tag was never closed
  Stack Trace:
     at DotLiquid.Tokenizer.Tokenize(String source)
   at DotLiquid.Template.Tokenize(String source)
   at DotLiquid.Template.ParseInternal(String source)
   at DotLiquid.Template.Parse(String source)
```

Parse this liquid file. It succeeds in 2.1.405.
```liquid
{%- comment -%} a {%- endcomment -%}
```

https://github.com/dotnet/docfx/pull/7442

microalps:
Resolves #439

### DIFF
--- a/src/DotLiquid.Tests/Tags/RawTests.cs
+++ b/src/DotLiquid.Tests/Tags/RawTests.cs
@@ -21,6 +21,15 @@ namespace DotLiquid.Tests.Tags
         }
 
         [Test]
+        public void TestRawWithErbLikeTrimmingWhitespace()
+        {
+            Helper.AssertTemplateResult("{{ test }}", "{%- raw %}{{ test }}{%- endraw %}");
+            Helper.AssertTemplateResult("{{ test }}", "{% raw -%}{{ test }}{% endraw -%}");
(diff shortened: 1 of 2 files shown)


src/DotLiquid.Tests/Tags/RawTests.cs
```diff
@@ -20,6 +20,15 @@ public void TestOutputInRaw ()
                 "{% raw %}{{ test }}{% endraw %}");
         }
 
+        [Test]
+        public void TestRawWithErbLikeTrimmingWhitespace()
+        {
+            Helper.AssertTemplateResult("{{ test }}", "{%- raw %}{{ test }}{%- endraw %}");
```
daviburg, line 26: If I understand this correctly DotLiquid completely ignores whitespace control from Shopify's spec. So this is just a temporary stop gap while the real work is implementing the whitespace control feature/

daviburg (approved):

PR #441, opened 2021-07-06
Strict Liquid Syntax (Part 2) + Render Performance

### LINKED ISSUE #390: Capitalize does not follow Shopify spec - capitalizes each word

### LINKED ISSUE #393: Sort filter does not follow casing behavior from Shopify's specification
### Dotliquid version
Current as of 6-22-2020

### Expected behavior
Shopify specification compliant behavior case sensitive sorting
```

Sally Snake, giraffe, octopus, zebra
```
https://shopify.github.io/liquid/filters/sort/

### Actual behavior
Sort natural filter case insensitive behavior is implemented instead

```

giraffe, octopus, Sally Snake, zebra
```

https://shopify.github.io/liquid/filters/sort_natural/

Use online try liquid with the Shopify sample

```
{% assign my_array = "zebra, octopus, giraffe, Sally Snake" | split: ", " %}

{{ my_array | sort | join: ", " }}
```

### LINKED ISSUE #411: Whitespace control is incompatible with Shopify's docs
### Sample code
```liquid
# Without whitespace control
{% assign name = 'John G. Chalmers-Smith' %}
{% if name and name.size > 10 %}
    Wow, {{ name }}, you have a long name!
{% else %}
    Hello there!
{% endif %}

# With whitespace control
{% assign name = 'John G. Chalmers-Smith' -%}
{%- if name and name.size > 10 -%}
    Wow, {{ name }}, you have a long name!
{%- else -%}
    Hello there!
{%- endif -%}
```
### Dotliquid version
Unknown (using DotLiquid Online)

### Expected behavior
```
# Without whitespace control


    Wow, John G. Chalmers-Smith, you have a long name!


# With whitespace control
Wow, John G. Chalmers-Smith, you have a long name!
```
Generated by [LiquidJS](https://liquidjs.com/playground.html#IyBXaXRob3V0IHdoaXRlc3BhY2UgY29udHJvbA0KeyUgYXNzaWduIG5hbWUgPSAnSm9obiBHLiBDaGFsbWVycy1TbWl0aCcgJX0NCnslIGlmIG5hbWUgYW5kIG5hbWUuc2l6ZSA+IDEwICV9DQogICAgV293LCB7eyBuYW1lIH19LCB5b3UgaGF2ZSBhIGxvbmcgbmFtZSENCnslIGVsc2UgJX0NCiAgICBIZWxsbyB0aGVyZSENCnslIGVuZGlmICV9DQoNCiMgV2l0aCB3aGl0ZXNwYWNlIGNvbnRyb2wNCnslIGFzc2lnbiBuYW1lID0gJ0pvaG4gRy4gQ2hhbG1lcnMtU21pdGgnIC0lfQ0KeyUtIGlmIG5hbWUgYW5kIG5hbWUuc2l6ZSA+IDEwIC0lfQ0KICAgIFdvdywge3sgbmFtZSB9fSwgeW91IGhhdmUgYSBsb25nIG5hbWUhDQp7JS0gZWxzZSAtJX0NCiAgICBIZWxsbyB0aGVyZSENCnslLSBlbmRpZiAtJX0=)

### Actual behavior
```
# Without whitespace control


    Wow, John G. Chalmers-Smith, you have a long name!


# With whitespace control
    Wow, John G. Chalmers-Smith, you have a long name!
```
Generated by http://dotliquidmarkup.org/try-online
### References
- https://shopify.github.io/liquid/basics/whitespace/

### LINKED ISSUE #413: [Security] Unrecognized filters don't trigger a rendering error/exception
### Dotliquid version
2.0.366

### Expected behavior

Assuming that no custom filters are registered, attempting to render a template like `{{ "hello world" | unknownfilter }}` should throw an exception.

This is an important problem because template filters are often used as a guard against information leakage (e.g. masking secrets) and injection vulnerabilities (string escaping, etc). Ideally developers would write tests to ensure that these facilities are working properly (I found this issue by writing such tests), but in the real world that often doesn't happen. Many devs will incorrectly assume that unrecognized filters will trigger an exception, and therefore assume that if the template renders without throwing, their filters are safeguarding their system correctly, and explicit tests to verify this will be treated as a low priority due to these assumptions.

### Actual behavior

The missing filter is ignored and the template renders as if it were written `{{ "hello world" }}`.

### LINKED ISSUE #445: ArgumentNullException due to invalid variable name
### Dotliquid version
4.1.x

### Expected behavior
Clear error either (a) output into the template with the variable that could not be understood, or (b) rethrown, as per the render settings

### Actual behavior
ArgumentNullException with no detail what caused this

From #436:
 
Encountered this while moving out templates from Mustache to Liquid, a if statement can be written as follow in Mustache:
```Mustache
{{#person}}
  {{.}}
{{/person}}
```
Where `{{.}}` refers to the person variable itself, so the Liquid equivalent would be:
```Liquid
{% if person %}
  {{ person }}
{% endif %}
```

Obviously it's not gonna work when the Mustache template is being processed by dotliquid, but if you would have `{{.}}` in your liquid template you'd get an `ArgumentNullException`

microalps:
This improves upon render time with a slew of under the hood changes. Notable changes:
For all versions:
- Removed regex parsing of variables that are clearly going to fail. For example, the variable doesn't need to be tested for literal if it doesn't start with a quote
- IsHashOrArrayLikeObject and LookupAndEvaluate had much of the same code - now it's combined into TryEvaluateHashOrArrayLikeObject - originally discussed with and developed by @robin-parker in #437 
- Resolves #445 originally found by @wouterjanson in #436
- Adds a way to get back other behaviors of Capitalize
- Implements SortNatural (originally reported in #330) - this is recommended EVEN IF not switching to DotLiquid 2.2 to future-proof your templates and maintain cross-compatibility

With an opt-in flag for DotLiquid 2.2 Syntax
- Variable segment parsing is no longer using Regex - this is for the parts between curly braces {{ }} as well as as if, case, and when tags. 
- Resolves #411 reported by @jolleekin
- Removes non-standard short-hand comments
- Updated Capitalize filters to be inline with Reference Library. Resolves #390 
- Resolves #413 
- Resolves #393 making the Sort filter case-sensitive. Coupled with the new SortNatural, provides a path forward compliant with the Liquid spec

It remains unknown how much this impacts the benchmarks on https://github.com/RicoSuter/NJsonSchema/pull/1346 ... would love to hear from that team

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,5 +1,5 @@
-﻿# Build script for dotliquid, see https://www.appveyor.com/docs/appveyor-yml for reference
-version: 2.1.{build}
+# Build script for dotliquid, see https://www.appveyor.com/docs/appveyor-yml for reference
+version: 2.2.{build}
 image: Visual Studio 2017
 
 cache:
--- a/src/DotLiquid.Tests/ExtendedFilterTests.cs
+++ b/src/DotLiquid.Tests/ExtendedFilterTests.cs
@@ -31,8 +31,7 @@ namespace DotLiquid.Tests
 
             Helper.AssertTemplateResult(
                 expected: "Title",
-                template: "{{ 'title' | titleize }}",
-                syntax: context.SyntaxCompatibilityLevel);
+                template: "{{ 'title' | titleize }}");
         }
 
         [Test]
@@ -46,8 +45,7 @@ namespace DotLiquid.Tests
 
             Helper.AssertTemplateResult(
                 expected: "My great title",
-                template: "{{ 'my great title' | upcase_first }}",
-                syntax: context.SyntaxCompatibilityLevel);
+                template: "{{ 'my great title' | upcase_first }}");
         }
     }
 }
--- a/src/DotLiquid.Tests/ExtendedFilterTests.cs
+++ b/src/DotLiquid.Tests/ExtendedFilterTests.cs
@@ -1,0 +1,53 @@
+using NUnit.Framework;
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace DotLiquid.Tests
+{
+    [TestFixture]
+    public class ExtendedFilterTests
+    {
+        private Context _context;
+
+        [OneTimeSetUp]
+        public void SetUp()
+        {
+            _context = new Context(CultureInfo.InvariantCulture);
+            Template.RegisterFilter(typeof(ExtendedFilters));
+        }
+
+        [Test]
+        public void TestTitleize()
+        {
+            var context = _context;
+            Assert.AreEqual(null, ExtendedFilters.Titleize(context: context, input: null));
+            Assert.AreEqual("", ExtendedFilters.Titleize(context: context, input: ""));
+            Assert.AreEqual(" ", ExtendedFilters.Titleize(context: context, input: " "));
+            Assert.AreEqual("That Is One Sentence.", ExtendedFilters.Titleize(context: context, input: "That is one sentence."));
+
+            Helper.AssertTemplateResult(
+                expected: "Title",
+                template: "{{ 'title' | titleize }}",
+                syntax: context.SyntaxCompatibilityLevel);
+        }
+
+        [Test]
+        public void TestUpcaseFirst()
+        {
+            var context = _context;
+            Assert.AreEqual(null, ExtendedFilters.UpcaseFirst(context: context, input: null));
+            Assert.AreEqual("", ExtendedFilters.UpcaseFirst(context: context, input: ""));
+            Assert.AreEqual(" ", ExtendedFilters.UpcaseFirst(context: context, input: " "));
+            Assert.AreEqual(" My boss is Mr. Doe.", ExtendedFilters.UpcaseFirst(context: context, input: " my boss is Mr. Doe."));
+
+            Helper.AssertTemplateResult(
+                expected: "My great title",
+                template: "{{ 'my great title' | upcase_first }}",
+                syntax: context.SyntaxCompatibilityLevel);
+        }
+    }
+}
--- a/src/DotLiquid.Tests/ParsingQuirksTests.cs
+++ b/src/DotLiquid.Tests/ParsingQuirksTests.cs
@@ -126,9 +126,9 @@ namespace DotLiquid.Tests
         [Test]
         public void TestShortHandSyntaxIsIgnored()
         {
-            //These tests are based on actual handling on Ruby Liquid, not indicative of wanted behavior
+            //These tests are based on actual handling on Ruby Liquid, not indicative of wanted behavior. Behavior for legacy dotliquid parser is in TestEmptyLiteral
             Assert.AreEqual("}", Template.Parse("{{{}}}", SyntaxCompatibility.DotLiquid22).Render());
-            Assert.AreEqual(string.Empty, Template.Parse("{{%%}}", SyntaxCompatibility.DotLiquid22).Render());
+            Assert.AreEqual("{##}", Template.Parse("{##}", SyntaxCompatibility.DotLiquid22).Render());
         }
     }
 }
--- a/src/DotLiquid.Tests/ParsingQuirksTests.cs
+++ b/src/DotLiquid.Tests/ParsingQuirksTests.cs
@@ -74,14 +74,16 @@ namespace DotLiquid.Tests
         [TestCase(".y")]
         public void TestVariableNotTerminatedFromInvalidVariableName(string variableName)
         {
-            Template template = Template.Parse("{{ " + variableName + " }}");
+            var template = Template.Parse("{{ " + variableName + " }}");
             SyntaxException ex = Assert.Throws<SyntaxException>(() => template.Render(new RenderParameters(System.Globalization.CultureInfo.InvariantCulture)
             {
                 LocalVariables = Hash.FromAnonymousObject(new { x = "" }),
                 ErrorsOutputMode = ErrorsOutputMode.Rethrow,
                 SyntaxCompatibilityLevel = SyntaxCompatibility.DotLiquid22
             }));
-            Assert.AreEqual(string.Format(Liquid.ResourceManager.GetString("VariableNotTerminatedException"), variableName), ex.Message);
+            Assert.AreEqual(
+                expected: string.Format(Liquid.ResourceManager.GetString("VariableNotTerminatedException"), variableName),
+                actual: ex.Message);
 
             template = Template.Parse("{{ x[" + variableName + "] }}");
             ex = Assert.Throws<SyntaxException>(() => template.Render(new RenderParameters(System.Globalization.CultureInfo.InvariantCulture)
@@ -90,7 +92,9 @@ namespace DotLiquid.Tests
                 ErrorsOutputMode = ErrorsOutputMode.Rethrow,
                 SyntaxCompatibilityLevel = SyntaxCompatibility.DotLiquid22
             }));
-            Assert.AreEqual(string.Format(Liquid.ResourceManager.GetString("VariableNotTerminatedException"), variableName), ex.Message);
+            Assert.AreEqual(
+                expected: string.Format(Liquid.ResourceManager.GetString("VariableNotTerminatedException"), variableName),
+                actual: ex.Message);
         }
 
         [Test]
@@ -103,7 +107,9 @@ namespace DotLiquid.Tests
                 ErrorsOutputMode = ErrorsOutputMode.Rethrow,
                 SyntaxCompatibilityLevel = SyntaxCompatibility.DotLiquid22
             }));
-            Assert.AreEqual(string.Format(Liquid.ResourceManager.GetString("VariableNotTerminatedException"), "["), ex.Message);
+            Assert.AreEqual(
+                expected: string.Format(Liquid.ResourceManager.GetString("VariableNotTerminatedException"), "["),
+                actual: ex.Message);
         }
 
         [TestCase("[\"]")]
@@ -112,7 +118,9 @@ namespace DotLiquid.Tests
         public void TestVariableTokenizerNotTerminated(string variableName)
         {
             var ex = Assert.Throws<SyntaxException>(() => Tokenizer.GetVariableEnumerator(variableName).MoveNext());
-            Assert.AreEqual(string.Format(Liquid.ResourceManager.GetString("VariableNotTerminatedException"), variableName), ex.Message);
+            Assert.AreEqual(
+                expected: string.Format(Liquid.ResourceManager.GetString("VariableNotTerminatedException"), variableName),
+                actual: ex.Message);
         }
 
         [Test]
--- a/src/DotLiquid.Tests/ParsingQuirksTests.cs
+++ b/src/DotLiquid.Tests/ParsingQuirksTests.cs
@@ -65,5 +65,54 @@ namespace DotLiquid.Tests
             Helper.AssertTemplateResult("{{ {% %} }}", "{{ '{{ {% %} }}' }}");
             Helper.AssertTemplateResult("{{ {% %} }}", "{% assign x = '{{ {% %} }}' %}{{x}}");
         }
+
+        [TestCase(".")]
+        [TestCase("x.")]
+        [TestCase("$x")]
+        [TestCase("x?")]
+        [TestCase("x¿")]
+        [TestCase(".y")]
+        public void TestVariableNotTerminatedFromInvalidVariableName(string variableName)
+        {
+            Template template = Template.Parse("{{ " + variableName + " }}");
+            SyntaxException ex = Assert.Throws<SyntaxException>(() => template.Render(new RenderParameters(System.Globalization.CultureInfo.InvariantCulture)
+            {
+                LocalVariables = Hash.FromAnonymousObject(new { x = "" }),
+                ErrorsOutputMode = ErrorsOutputMode.Rethrow,
+                SyntaxCompatibilityLevel = SyntaxCompatibility.DotLiquid22
+            }));
+            Assert.AreEqual(string.Format(Liquid.ResourceManager.GetString("VariableNotTerminatedException"), variableName), ex.Message);
+
+            template = Template.Parse("{{ x[" + variableName + "] }}");
+            ex = Assert.Throws<SyntaxException>(() => template.Render(new RenderParameters(System.Globalization.CultureInfo.InvariantCulture)
+            {
+                LocalVariables = Hash.FromAnonymousObject(new { x = new { x = "" } }),
+                ErrorsOutputMode = ErrorsOutputMode.Rethrow,
+                SyntaxCompatibilityLevel = SyntaxCompatibility.DotLiquid22
+            }));
+            Assert.AreEqual(string.Format(Liquid.ResourceManager.GetString("VariableNotTerminatedException"), variableName), ex.Message);
+        }
+
+        [Test]
+        public void TestNestedVariableNotTerminated()
+        {
+            var template = Template.Parse("{{ x[[] }}");
+            var ex = Assert.Throws<SyntaxException>(() => template.Render(new RenderParameters(System.Globalization.CultureInfo.InvariantCulture)
+            {
+                LocalVariables = Hash.FromAnonymousObject(new { x = new { x = "" } }),
+                ErrorsOutputMode = ErrorsOutputMode.Rethrow,
+                SyntaxCompatibilityLevel = SyntaxCompatibility.DotLiquid22
+            }));
+            Assert.AreEqual(string.Format(Liquid.ResourceManager.GetString("VariableNotTerminatedException"), "["), ex.Message);
+        }
+
+        [TestCase("[\"]")]
+        [TestCase("[\"\"")]
+        [TestCase("[']")]
+        public void TestVariableTokenizerNotTerminated(string variableName)
+        {
+            var ex = Assert.Throws<SyntaxException>(() => Tokenizer.GetVariableEnumerator(variableName).MoveNext());
+            Assert.AreEqual(string.Format(Liquid.ResourceManager.GetString("VariableNotTerminatedException"), variableName), ex.Message);
+        }
     }
 }
--- a/src/DotLiquid.Tests/ParsingQuirksTests.cs
+++ b/src/DotLiquid.Tests/ParsingQuirksTests.cs
@@ -114,5 +114,13 @@ namespace DotLiquid.Tests
             var ex = Assert.Throws<SyntaxException>(() => Tokenizer.GetVariableEnumerator(variableName).MoveNext());
             Assert.AreEqual(string.Format(Liquid.ResourceManager.GetString("VariableNotTerminatedException"), variableName), ex.Message);
         }
+
+        [Test]
+        public void TestShortHandSyntaxIsIgnored()
+        {
+            //These tests are based on actual handling on Ruby Liquid, not indicative of wanted behavior
+            Assert.AreEqual("}", Template.Parse("{{{}}}", SyntaxCompatibility.DotLiquid22).Render());
+            Assert.AreEqual(string.Empty, Template.Parse("{{%%}}", SyntaxCompatibility.DotLiquid22).Render());
+        }
     }
 }
--- a/src/DotLiquid.Tests/ParsingQuirksTests.cs
+++ b/src/DotLiquid.Tests/ParsingQuirksTests.cs
@@ -126,7 +126,7 @@ namespace DotLiquid.Tests
         [Test]
         public void TestShortHandSyntaxIsIgnored()
         {
-            //These tests are based on actual handling on Ruby Liquid, not indicative of wanted behavior. Behavior for legacy dotliquid parser is in TestEmptyLiteral
+            // These tests are based on actual handling on Ruby Liquid, not indicative of wanted behavior. Behavior for legacy dotliquid parser is in TestEmptyLiteral
             Assert.AreEqual("}", Template.Parse("{{{}}}", SyntaxCompatibility.DotLiquid22).Render());
             Assert.AreEqual("{##}", Template.Parse("{##}", SyntaxCompatibility.DotLiquid22).Render());
         }
--- a/src/DotLiquid.Tests/Tags/CommentTests.cs
+++ b/src/DotLiquid.Tests/Tags/CommentTests.cs
@@ -1,0 +1,45 @@
+using System;
+using System.Collections.Generic;
+using DotLiquid.Exceptions;
+using NUnit.Framework;
+using DotLiquid.Tags;
+
+namespace DotLiquid.Tests.Tags
+{
+    [TestFixture]
+    public class CommentTests
+    {
+        [Test]
+        public void TestEmptyComment()
+        {
+            Assert.AreEqual(string.Empty, Template.Parse("{% comment %}{% endcomment %}").Render());
+            // Next test is specific to legacy parser and was removed from Ruby Liquid. Test that it is ignored is in TestShortHandSyntaxIsIgnored
+            Assert.AreEqual(string.Empty, Template.Parse("{##}", SyntaxCompatibility.DotLiquid20).Render());
+        }
+
+        [Test]
+        public void TestSimpleCommentValue()
+        {
+            Assert.AreEqual("", Template.Parse("{% comment %}howdy{% endcomment %}").Render());
+        }
+
+        [Test]
+        public void TestCommentsIgnoreLiquidMarkup()
+        {
+            Assert.AreEqual(
+                expected: "",
+                actual: Template.Parse("{% comment %}{% if 'gnomeslab' contains 'liquid' %}yes{% else %}no{ % endif %}{% endcomment %}").Render());
+        }
+
+        [Test]
+        public void TestCommentShorthand()
+        {
+            Assert.AreEqual("{% comment %}gnomeslab{% endcomment %}", Comment.FromShortHand("{# gnomeslab #}"));
+            Assert.AreEqual(null, Comment.FromShortHand(null));
+
+            Assert.AreEqual(
+                expected: "",
+                actual: Template.Parse("{#{% if 'gnomeslab' contains 'liquid' %}yes{ % endif %}#}", SyntaxCompatibility.DotLiquid20).Render());
+        }
+    }
+}
--- a/src/DotLiquid.Tests/Tags/CommentTests.cs
+++ b/src/DotLiquid.Tests/Tags/CommentTests.cs
@@ -1,8 +1,5 @@
-using System;
-using System.Collections.Generic;
-using DotLiquid.Exceptions;
-using NUnit.Framework;
 using DotLiquid.Tags;
+using NUnit.Framework;
 
 namespace DotLiquid.Tests.Tags
 {
@@ -13,6 +10,7 @@ namespace DotLiquid.Tests.Tags
         public void TestEmptyComment()
         {
             Assert.AreEqual(string.Empty, Template.Parse("{% comment %}{% endcomment %}").Render());
+
             // Next test is specific to legacy parser and was removed from Ruby Liquid. Test that it is ignored is in TestShortHandSyntaxIsIgnored
             Assert.AreEqual(string.Empty, Template.Parse("{##}", SyntaxCompatibility.DotLiquid20).Render());
         }
--- a/src/DotLiquid.Tests/Tags/LiteralTests.cs
+++ b/src/DotLiquid.Tests/Tags/LiteralTests.cs
@@ -14,8 +14,7 @@ namespace DotLiquid.Tests.Tags
         {
             Template t = Template.Parse("{% literal %}{% endliteral %}");
             Assert.AreEqual(string.Empty, t.Render());
-            t = Template.Parse("{{{}}}");
-            Assert.AreEqual(string.Empty, t.Render());
+            Assert.AreEqual(string.Empty, Template.Parse("{{{}}}", SyntaxCompatibility.DotLiquid20).Render());
         }
 
         [Test]
@@ -35,14 +34,14 @@ namespace DotLiquid.Tests.Tags
         [Test]
         public void TestShorthandSyntax()
         {
-            Template t = Template.Parse("{{{{% if 'gnomeslab' contains 'liquid' %}yes{ % endif %}}}}");
+            Template t = Template.Parse("{{{{% if 'gnomeslab' contains 'liquid' %}yes{ % endif %}}}}", SyntaxCompatibility.DotLiquid20);
             Assert.AreEqual("{% if 'gnomeslab' contains 'liquid' %}yes{ % endif %}", t.Render());
         }
 
         [Test]
         public void TestLiteralsDontRemoveComments()
         {
-            Template t = Template.Parse("{{{ {# comment #} }}}");
+            Template t = Template.Parse("{{{ {# comment #} }}}", SyntaxCompatibility.DotLiquid20);
             Assert.AreEqual("{# comment #}", t.Render());
         }
 
--- a/src/DotLiquid.Tests/Tags/LiteralTests.cs
+++ b/src/DotLiquid.Tests/Tags/LiteralTests.cs
@@ -13,6 +13,7 @@ namespace DotLiquid.Tests.Tags
         public void TestEmptyLiteral()
         {
             Assert.AreEqual(string.Empty, Template.Parse("{% literal %}{% endliteral %}").Render());
+
             // Next test is specific to legacy parser and was removed from Ruby Liquid. Test that it is ignored is in TestShortHandSyntaxIsIgnored
             Assert.AreEqual(string.Empty, Template.Parse("{{{}}}", SyntaxCompatibility.DotLiquid20).Render());
         }
--- a/src/DotLiquid.Tests/Tags/LiteralTests.cs
+++ b/src/DotLiquid.Tests/Tags/LiteralTests.cs
@@ -12,37 +12,36 @@ namespace DotLiquid.Tests.Tags
         [Test]
         public void TestEmptyLiteral()
         {
-            Template t = Template.Parse("{% literal %}{% endliteral %}");
-            Assert.AreEqual(string.Empty, t.Render());
+            Assert.AreEqual(string.Empty, Template.Parse("{% literal %}{% endliteral %}").Render());
             Assert.AreEqual(string.Empty, Template.Parse("{{{}}}", SyntaxCompatibility.DotLiquid20).Render());
         }
 
         [Test]
         public void TestSimpleLiteralValue()
         {
-            Template t = Template.Parse("{% literal %}howdy{% endliteral %}");
-            Assert.AreEqual("howdy", t.Render());
+            Assert.AreEqual("howdy", Template.Parse("{% literal %}howdy{% endliteral %}").Render());
         }
 
         [Test]
         public void TestLiteralsIgnoreLiquidMarkup()
         {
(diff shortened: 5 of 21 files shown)


microalps:
Performance (based on Fluid's benchmark) vs latest NuGet
```
|              Method |        Job | BuildConfiguration |      Mean |    Error |   StdDev | Ratio | RatioSD |   Gen 0 |  Gen 1 | Gen 2 | Allocated |
|-------------------- |----------- |------------------- |----------:|---------:|---------:|------:|--------:|--------:|-------:|------:|----------:|
|               Parse | Job-SCOQHT |              NuGet |  19.99 us | 0.394 us | 0.679 us |  1.00 |    0.00 |  2.6245 | 0.0305 |     - |     16 KB |
|               Parse | DefaultJob |            Default |  19.40 us | 0.386 us | 0.716 us |  0.97 |    0.04 |  2.6245 | 0.0305 |     - |     16 KB |
|                     |            |                    |           |          |          |       |         |         |        |       |           |
|            ParseBig | Job-SCOQHT |              NuGet |  75.11 us | 0.949 us | 0.841 us |  1.00 |    0.00 | 15.3809 | 1.0986 |     - |     94 KB |
|            ParseBig | DefaultJob |            Default |  74.19 us | 0.947 us | 0.840 us |  0.99 |    0.02 | 15.3809 | 1.0986 |     - |     94 KB |
|                     |            |                    |           |          |          |       |         |         |        |       |           |
|    RenderNestedHash | Job-SCOQHT |              NuGet | 492.93 us | 4.901 us | 4.345 us |  1.00 |    0.00 | 66.4063 | 9.7656 |     - |    410 KB |
|    RenderNestedHash | DefaultJob |            Default | 277.51 us | 3.998 us | 3.339 us |  0.56 |    0.01 | 46.3867 | 6.8359 |     - |    287 KB |
|                     |            |                    |           |          |          |       |         |         |        |       |           |
| RenderILiquidizable | Job-SCOQHT |              NuGet | 965.68 us | 8.052 us | 7.532 us |  1.00 |    0.00 | 73.2422 | 8.7891 |     - |    452 KB |
| RenderILiquidizable | DefaultJob |            Default | 564.06 us | 2.323 us | 2.060 us |  0.58 |    0.00 | 49.8047 | 5.8594 |     - |    310 KB |
```

src/DotLiquid/Template.cs
```diff
@@ -8,6 +8,7 @@
 using DotLiquid.FileSystems;
 using DotLiquid.Util;
 using DotLiquid.NamingConventions;
+using System.Collections.Concurrent;
```
daviburg, line 11: nit: insert new using with existing system using statements

src/DotLiquid/Template.cs
```diff
@@ -195,16 +199,19 @@ public static void RegisterValueTypeTransformer(Type type, Func<object, object>
...
-            return null;
+                var interfaces = type.GetTypeInfo().ImplementedInterfaces;
+                foreach (var interfaceType in interfaces)
+                {
+                    if (ValueTypeTransformers.TryGetValue(interfaceType, out transformer))
+                        return transformer;
+                    if (interfaceType.GetTypeInfo().IsGenericType && ValueTypeTransformers.TryGetValue(
+                        interfaceType.GetGenericTypeDefinition(), out transformer))
+                        return transformer;
+                }
```
daviburg, line 212: nit: insert single empty line after closing a curly bracket code block.

src/DotLiquid/Template.cs
```diff
@@ -195,16 +199,19 @@ public static void RegisterValueTypeTransformer(Type type, Func<object, object>
...
-                    return transformer;
-            }
-            return null;
+                var interfaces = type.GetTypeInfo().ImplementedInterfaces;
+                foreach (var interfaceType in interfaces)
+                {
+                    if (ValueTypeTransformers.TryGetValue(interfaceType, out transformer))
+                        return transformer;
+                    if (interfaceType.GetTypeInfo().IsGenericType && ValueTypeTransformers.TryGetValue(
+                        interfaceType.GetGenericTypeDefinition(), out transformer))
```
daviburg, line 210: nit, formatting: consider starting each line with a condition. If you need to multi-line parameters, consider starting each line with a parameter and putting each parameter on its own line

src/DotLiquid/Tokenizer.cs
```diff
@@ -90,13 +99,68 @@ internal static List<string> Tokenize(string source)
...
+        /// </summary>
+        /// <param name="source">The Liquid Variable string</param>
+        /// <returns></returns>
+        /// <exception cref="SyntaxException"></exception>
+        internal static IEnumerator<string> GetVariableEnumerator(string source)
+        {
+            if (string.IsNullOrEmpty(source))
+                yield break;
+
+            using (var markupEnumerator = new CharEnumerator(source))
```
daviburg, line 113: This using lasts until the end of method so it can be simplified as

`using var markupEnumerator = new CharEnumerator(source);`

and no code block.

src/DotLiquid/Tokenizer.cs
```diff
@@ -90,13 +99,68 @@ internal static List<string> Tokenize(string source)
...
+            {
+                while (markupEnumerator.HasNext())
+                {
+                    var isComplete = false;
+                    var sb = new StringBuilder();
```
daviburg, line 118: nit: consider more description variable name than `sb` - what is the string builder's purpose or content?

daviburg (changes requested):

src/DotLiquid/Tokenizer.cs
microalps: This won't compile for .net 4.5 or core 1.x - 
Feature 'using declarations' is not available in C# 7.3. Please use language version 8.0 or greater.

src/DotLiquid/Tokenizer.cs
microalps: Will fix this in this method, but as a parameter for the other methods it is consistent with Microsoft's own StringWriter constructor.

daviburg: We haven't always been very good at following our own best practices guidance. And old habits are pretty resilient to change :-)

src/DotLiquid/Hash.cs
```diff
@@ -123,15 +104,20 @@ private static void FromAnonymousObject(object anonymousObject, Hash hash, bool
             return expr.Compile();
         }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DotLiquid.Hash">Hash</see> class and populates it with the contents of a dictionary
+        /// </summary>
+        /// <param name="dictionary">The Dictionary object</param>
+        /// <returns></returns>
```
daviburg, line 111: Nit: avoid empty XML elements in comments. Each element can be omitted entirely if there is nothing to say. Or add what you see as appropriate in the returns element.

src/DotLiquid/Tokenizer.cs
```diff
@@ -124,32 +185,84 @@ private static void ReadToEnd(StringBuilder sb, DotLiquid.Util.CharEnumerator ma
...
+        /// Reads a token until the end of a variable segment and advances the enumerator position
+        /// </summary>
+        /// <param name="sb">The StringBuilder to write to</param>
+        /// <param name="markupEnumerator">The string enumerator</param>
+        /// <returns>True if valid variable, otherwise false</returns>
+        private static bool ReadToEndOfVariable(StringBuilder sb, CharEnumerator markupEnumerator)
+        {
+            while (markupEnumerator.HasNext())
+            {
+                char nextChar = markupEnumerator.Next;
```
daviburg, line 234: nit: consider using `var`

src/DotLiquid/Tokenizer.cs
```diff
@@ -124,32 +185,84 @@ private static void ReadToEnd(StringBuilder sb, DotLiquid.Util.CharEnumerator ma
...
+                    return true;
+
+                if (!ReadWordChar(sb, markupEnumerator))
+                    return false;
+            };
+            return true;
+        }
+
+        /// <summary>
+        /// Reads a single character and advances the enumerator position
```
daviburg, line 245: nit: `/// Reads a single word-character and advances the enumerator position`

src/DotLiquid/Tokenizer.cs
```diff
@@ -124,32 +185,84 @@ private static void ReadToEnd(StringBuilder sb, DotLiquid.Util.CharEnumerator ma
...
+        /// </summary>
+        /// <param name="sb">The StringBuilder to write to</param>
+        /// <param name="markupEnumerator">The string enumerator</param>
+        /// <returns>True if upcoming character is a word character, otherwise false</returns>
+        private static bool ReadWordChar(StringBuilder sb, CharEnumerator markupEnumerator)
+        {
+            char nextChar = markupEnumerator.Next;
+            if (nextChar < 128) // avoid regex for standard ascii
```
daviburg, line 253: nit: `// For better performance, avoid regex for standard ascii`

microalps:
Here is an updated benchmark, accounting for the differences in behavior between DotLiquid 2.1 and 2.2
```
|              Method |        Job | BuildConfiguration | SyntaxCompatibilityLevel |      Mean |     Error |    StdDev | Ratio | RatioSD |   Gen 0 |  Gen 1 | Gen 2 | Allocated |
|-------------------- |----------- |------------------- |------------------------- |----------:|----------:|----------:|------:|--------:|--------:|-------:|------:|----------:|
|               Parse | Job-XXIMLP |              NuGet |              DotLiquid20 |  19.15 us |  0.114 us |  0.107 us |  1.00 |    0.00 |  2.6550 | 0.0305 |     - |     16 KB |
|               Parse | DefaultJob |            Default |              DotLiquid20 |  19.00 us |  0.353 us |  0.346 us |  0.99 |    0.02 |  2.6550 | 0.0305 |     - |     16 KB |
|                     |            |                    |                          |           |           |           |       |         |         |        |       |           |
|            ParseBig | Job-XXIMLP |              NuGet |              DotLiquid20 |  75.91 us |  0.721 us |  0.639 us |  1.00 |    0.00 | 15.8691 | 1.2207 |     - |     98 KB |
|            ParseBig | DefaultJob |            Default |              DotLiquid20 |  76.06 us |  1.520 us |  1.492 us |  1.00 |    0.02 | 15.8691 | 1.2207 |     - |     98 KB |
|                     |            |                    |                          |           |           |           |       |         |         |        |       |           |
|    RenderNestedHash | Job-XXIMLP |              NuGet |              DotLiquid20 | 492.01 us |  4.058 us |  3.796 us |  1.00 |    0.00 | 66.4063 | 9.7656 |     - |    410 KB |
|    RenderNestedHash | DefaultJob |            Default |              DotLiquid20 | 330.10 us |  6.483 us |  6.064 us |  0.67 |    0.01 | 58.5938 | 8.7891 |     - |    362 KB |
|                     |            |                    |                          |           |           |           |       |         |         |        |       |           |
| RenderILiquidizable | Job-XXIMLP |              NuGet |              DotLiquid20 | 947.22 us |  7.149 us |  5.970 us |  1.00 |    0.00 | 73.2422 | 8.7891 |     - |    452 KB |
| RenderILiquidizable | DefaultJob |            Default |              DotLiquid20 | 650.13 us | 11.948 us | 18.951 us |  0.69 |    0.02 | 62.5000 | 6.8359 |     - |    386 KB |
|                     |            |                    |                          |           |           |           |       |         |         |        |       |           |
|               Parse | Job-XXIMLP |              NuGet |              DotLiquid22 |  19.50 us |  0.365 us |  0.342 us |  1.00 |    0.00 |  2.6550 | 0.0305 |     - |     16 KB |
|               Parse | DefaultJob |            Default |              DotLiquid22 |  18.40 us |  0.281 us |  0.263 us |  0.94 |    0.02 |  2.6245 | 0.0305 |     - |     16 KB |
|                     |            |                    |                          |           |           |           |       |         |         |        |       |           |
|            ParseBig | Job-XXIMLP |              NuGet |              DotLiquid22 |  76.41 us |  1.116 us |  1.044 us |  1.00 |    0.00 | 15.8691 | 1.2207 |     - |     98 KB |
|            ParseBig | DefaultJob |            Default |              DotLiquid22 |  70.04 us |  1.351 us |  1.757 us |  0.92 |    0.04 | 15.2588 | 1.0986 |     - |     94 KB |
|                     |            |                    |                          |           |           |           |       |         |         |        |       |           |
|    RenderNestedHash | Job-XXIMLP |              NuGet |              DotLiquid22 | 495.92 us |  8.054 us |  7.533 us |  1.00 |    0.00 | 66.4063 | 9.7656 |     - |    410 KB |
|    RenderNestedHash | DefaultJob |            Default |              DotLiquid22 | 272.86 us |  1.418 us |  1.257 us |  0.55 |    0.01 | 46.3867 | 6.8359 |     - |    285 KB |
|                     |            |                    |                          |           |           |           |       |         |         |        |       |           |
| RenderILiquidizable | Job-XXIMLP |              NuGet |              DotLiquid22 | 957.73 us | 12.200 us | 11.412 us |  1.00 |    0.00 | 72.2656 | 7.8125 |     - |    452 KB |
| RenderILiquidizable | DefaultJob |            Default |              DotLiquid22 | 578.03 us |  3.955 us |  3.506 us |  0.60 |    0.01 | 49.8047 | 5.8594 |     - |    309 KB |
```
Summary: 2.2 gives a 10% edge over 2.1 on render and on parse operations. Overall, rendering can be 65% - 80% faster when utilizing all the performance tweaks.

src/DotLiquid.Tests/TemplateTests.cs
```diff
@@ -8,47 +8,56 @@ namespace DotLiquid.Tests
...
+            var v20 = Tokenizer.Tokenize(input, SyntaxCompatibility.DotLiquid20);
+            var v22 = Tokenizer.Tokenize(input, SyntaxCompatibility.DotLiquid22);
+            Assert.AreEqual(v20, v22);
+            return v22;
+        }
+
         [Test]
         public void TestTokenizeStrings()
         {
-            CollectionAssert.AreEqual(new[] { " " }, Template.Tokenize(" "));
```
daviburg, line 14: Keep the class prefix for clarity, especially now that there are multiple Tokenize methods in this context.

src/DotLiquid.Tests/TemplateTests.cs
```diff
@@ -8,47 +8,56 @@ namespace DotLiquid.Tests
...
+            CollectionAssert.AreEqual(new[] { "  ", "{% comment %}", " ", "{% endcomment %}", " " }, Tokenize("  {% comment %} {% endcomment %} "));
         }
 
         [Test]
         public void TestInstanceAssignsPersistOnSameTemplateObjectBetweenParses()
         {
             Template t = new Template();
-            Assert.AreEqual("from instance assigns", t.ParseInternal("{% assign foo = 'from instance assigns' %}{{ foo }}").Render());
-            Assert.AreEqual("from instance assigns", t.ParseInternal("{{ foo }}").Render());
+            Assert.AreEqual("from instance assigns", t.ParseInternal("{% assign foo = 'from instance assigns' %}{{ foo }}", SyntaxCompatibility.DotLiquid22).Render());
```
daviburg, line 50: 2.2 would be the new default syntax compatibility version, so only explicit 2.1 version compatibility input is required, is it not?

src/DotLiquid.Tests/TemplateTests.cs
```diff
@@ -8,47 +8,56 @@ namespace DotLiquid.Tests
     [TestFixture]
     public class TemplateTests
     {
+
+        private System.Collections.Generic.List<string> Tokenize(string input)
+        {
+            var v20 = Tokenizer.Tokenize(input, SyntaxCompatibility.DotLiquid20);
```
daviburg, line 14: Why 2.0 not 2.1 to compare with 2.2?

src/DotLiquid/Context.cs
```diff
@@ -422,29 +433,42 @@ private object Resolve(string key, bool notifyNotFound = true)
...
-        private object FindVariable(string key)
+        private bool TryFindVariable(string key, out object variable)
         {
+            bool foundVariable = false;
             Hash scope = Scopes.FirstOrDefault(s => s.ContainsKey(key));
-            object variable = null;
+            object ret = null;
             if (scope == null)
             {
                 foreach (Hash e in Environments)
```
daviburg, line 445: nit: while you are here, consider renaming 'e' to 'environment' for clarity.

src/DotLiquid/Context.cs
```diff
@@ -422,29 +433,42 @@ private object Resolve(string key, bool notifyNotFound = true)
...
         /// <param name="key"></param>
+        /// <param name="variable"></param>
         /// <returns></returns>
-        private object FindVariable(string key)
+        private bool TryFindVariable(string key, out object variable)
         {
+            bool foundVariable = false;
             Hash scope = Scopes.FirstOrDefault(s => s.ContainsKey(key));
-            object variable = null;
+            object ret = null;
```
daviburg, line 442: I like that you eliminated the mutability of the previous local by having a different variable for the non-liquidized version, but what would be a more descriptive name than 'ret'?

src/DotLiquid/Context.cs
```diff
@@ -422,29 +433,42 @@ private object Resolve(string key, bool notifyNotFound = true)
...
+            }
+            else
+            {
+                foundVariable = TryEvaluateHashOrArrayLikeObject(scope, key, out ret);
             }
-            scope = scope ?? Environments.LastOrDefault() ?? Scopes.Last();
-            variable = variable ?? LookupAndEvaluate(scope, key);
 
-            variable = Liquidize(variable);
+            variable = Liquidize(ret);
```
daviburg, line 466: I think the expected contract from a TryXyz method is that the return object is assigned null if the attempt fails. Here it assigns even if `foundVariable` is false.

src/DotLiquid/Context.cs
```diff
@@ -461,33 +485,39 @@ private object FindVariable(string key)
...
 
             // try to resolve the rest of the parts (starting from the second item in the list)
-            for (int i = 1; i < parts.Count; ++i)
+            while (partsEnumerator.MoveNext())
             {
-                var forEachPart = parts[i];
-                Match partSquareBracketedMatch = SquareBracketedRegex.Match(forEachPart);
-                bool partResolved = partSquareBracketedMatch.Success;
+                string forEachPart = partsEnumerator.Current;
+                bool partResolved = forEachPart[0] == '[';
```
daviburg, line 516: R.Scan never returns a null string, is that right?

src/DotLiquid/Context.cs
```diff
@@ -461,33 +485,39 @@ private object FindVariable(string key)
...
 
                 object part = forEachPart;
                 if (partResolved)
-                    part = Resolve(partSquareBracketedMatch.Groups[1].Value);
+                    part = Resolve(forEachPart.Substring(1, forEachPart.Length - 2));
```
daviburg, line 520: Can the part be malformed as just "[" here and lead to an exception because length is 1 so the input becomes -1?

daviburg (changes requested):

src/DotLiquid.Tests/TemplateTests.cs
microalps: Template.Tokenize doesn't exist any more. It is a separate class Tokenizer.Tokenize now - and that expects a specific syntax as an argument. Hence the Tokenize above that actually tests multiple versions against each other.

src/DotLiquid.Tests/TemplateTests.cs
microalps: ParseInternal has no default - it is internal. The default, however, is actually 2.0 to introduce no breaking changes. This is computed in Parse method based on Template.DefaultSyntax property.
The reason why I specified 2.2 is because the only difference in handling is whitespace and there is no whitespace in this. By specifying 2.2 I ensure no regression with this commit. It could go either way.

src/DotLiquid/Context.cs
microalps: To me that is short for return - a reserved keyword. Would you prefer foundValue?

src/DotLiquid/Context.cs
microalps: Probably doesn't hurt because ret is null in that case. But makes sense.

src/DotLiquid/Context.cs
microalps: No, ToList() would return an empty object not null.

src/DotLiquid/Context.cs
microalps: tokenizer would throw an exception in that case. [ must be ended with ] and [" must end with "] - by itself it is not a word-character and not a complete token. Regex expression also works this way (but in that case, it would ignore the [ and not return as part of any token).

src/DotLiquid/Context.cs
microalps: Actually, you can't skip assignment - compiler throws an error - The out parameter 'variable' must be assigned to before control leaves the current method

src/DotLiquid.Tests/ExtendedFilterTests.cs
```diff
@@ -0,0 +1,53 @@
...
+
+namespace DotLiquid.Tests
+{
+    [TestFixture]
+    public class ExtendedFilterTests
+    {
+        private Context _context;
+
+        [OneTimeSetUp]
+        public void SetUp()
```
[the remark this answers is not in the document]
microalps: It matches the spelling throughout tests and with the line before of OneTimeSetUpAttribute

src/DotLiquid.Tests/TemplateTests.cs
microalps: 2.0 and 2.1 have no difference in parsing step (Tokenizer.Tokenize is part of Template.Parse). 2.2 introduces the whitespace control handling. Comparing earliest, the default shipped with DotLiquid, to latest probably makes the most sense.

src/DotLiquid.Tests/ExtendedFilterTests.cs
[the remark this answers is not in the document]
daviburg: British it is then.

src/DotLiquid/Context.cs
daviburg: Yes which is why engineers typically assign the out variable to `null` for when there isn't a match.

src/DotLiquid/Context.cs
daviburg: It is not the return here because both the return type now is the boolean for found or not found, and the out variable needs to be liquidize before been assigned.
As to `foundValue`, does value vs variable carry the nuance that one is liquidized and not the other?

src/DotLiquid.Tests/TemplateTests.cs
daviburg: So here `TemplateTests.Tokenize(` would make it clear which Tokenize is at play.

src/DotLiquid/Context.cs
microalps: Which is exactly what will happen, the underlying variable will be null, liquidized and still be null. I don't think there is a case where foundVariable is false and the variable is not null.

src/DotLiquid.Tests/TemplateTests.cs
microalps: I'll rename it to TokenizeMultiVersions? Will that be clearer what it is doing and also deal with your issue?

src/DotLiquid/Context.cs
daviburg: oh!

src/DotLiquid.Tests/TemplateTests.cs
daviburg: Sure, that will remove ambiguity. Personally I always put prefixes for clarity, but that's fine when at least the method name doesn't overlap.

appveyor.yml
```diff
@@ -1,5 +1,5 @@
-﻿# Build script for dotliquid, see https://www.appveyor.com/docs/appveyor-yml for reference
-version: 2.1.{build}
+# Build script for dotliquid, see https://www.appveyor.com/docs/appveyor-yml for reference
+version: 2.2.{build}
```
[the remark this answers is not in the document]
microalps: No, that would be a breaking change for most users that don't opt in to a version. Versions were only introduced recently with 2.1.x branch.

src/DotLiquid/Tokenizer.cs
```diff
@@ -124,32 +194,100 @@ private static void ReadToEnd(StringBuilder sb, DotLiquid.Util.CharEnumerator ma
...
-                                return;
+                                if (previousCharIsWhitespaceControl)
+                                {
+                                    // Remove hyphen from token
+                                    sb.Remove(sb.Length - 3, 1);
+
+                                    // Trim trailing whitespace by skipping ahead beyond the tag end
+                                    while (markupEnumerator.Remaining > 0)
+                                    {
+                                        if (((uint)markupEnumerator.Next - '\t') <= 5 || markupEnumerator.Next == ' ')
```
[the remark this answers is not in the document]
microalps: I based this on MS stating they do this for RegEx character classes in .NET 5.0. I guess it is more performant and technically we should change the other trim to potentially use this. For now, this will do. Once it is accepted we can do performance testing to see if there is any value to changing.

src/DotLiquid.Tests/ParsingQuirksTests.cs
```diff
@@ -65,5 +65,70 @@ public void TestLiquidTagsInQuotes()
...
+            var ex = Assert.Throws<SyntaxException>(() => Tokenizer.GetVariableEnumerator(variableName).MoveNext());
+            Assert.AreEqual(
+                expected: string.Format(Liquid.ResourceManager.GetString("VariableNotTerminatedException"), variableName),
+                actual: ex.Message);
+        }
+
+        [Test]
+        public void TestShortHandSyntaxIsIgnored()
+        {
+            //These tests are based on actual handling on Ruby Liquid, not indicative of wanted behavior
```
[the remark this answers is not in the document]
microalps: Put a comment in both places

src/DotLiquid.Tests/Tags/CommentTests.cs
```diff
@@ -0,0 +1,45 @@
...
+namespace DotLiquid.Tests.Tags
+{
+    [TestFixture]
+    public class CommentTests
+    {
+        [Test]
+        public void TestEmptyComment()
+        {
+            Assert.AreEqual(string.Empty, Template.Parse("{% comment %}{% endcomment %}").Render());
+            // Next test is specific to legacy parser and was removed from Ruby Liquid. Test that it is ignored is in TestShortHandSyntaxIsIgnored
```
daviburg, line 16: nit nit: have an empty line before an in-line code comment line
```suggestion

            // Next test is specific to legacy parser and was removed from Ruby Liquid. Test that it is ignored is in TestShortHandSyntaxIsIgnored
```

src/DotLiquid.Tests/ParsingQuirksTests.cs
```diff
@@ -65,5 +65,70 @@ public void TestLiquidTagsInQuotes()
...
+            var ex = Assert.Throws<SyntaxException>(() => Tokenizer.GetVariableEnumerator(variableName).MoveNext());
+            Assert.AreEqual(
+                expected: string.Format(Liquid.ResourceManager.GetString("VariableNotTerminatedException"), variableName),
+                actual: ex.Message);
+        }
+
+        [Test]
+        public void TestShortHandSyntaxIsIgnored()
+        {
+            //These tests are based on actual handling on Ruby Liquid, not indicative of wanted behavior. Behavior for legacy dotliquid parser is in TestEmptyLiteral
```
daviburg, line 129: nit nit: space character after // and before the first word (These)
```suggestion
            // These tests are based on actual handling on Ruby Liquid, not indicative of wanted behavior. Behavior for legacy dotliquid parser is in TestEmptyLiteral
```

src/DotLiquid.Tests/Tags/LiteralTests.cs
```diff
@@ -12,44 +12,44 @@ public class LiteralTests
         [Test]
         public void TestEmptyLiteral()
         {
-            Template t = Template.Parse("{% literal %}{% endliteral %}");
-            Assert.AreEqual(string.Empty, t.Render());
-            t = Template.Parse("{{{}}}");
-            Assert.AreEqual(string.Empty, t.Render());
+            Assert.AreEqual(string.Empty, Template.Parse("{% literal %}{% endliteral %}").Render());
+            // Next test is specific to legacy parser and was removed from Ruby Liquid. Test that it is ignored is in TestShortHandSyntaxIsIgnored
```
daviburg, line 16: nit: empty line to be added before comment line
```suggestion

            // Next test is specific to legacy parser and was removed from Ruby Liquid. Test that it is ignored is in TestShortHandSyntaxIsIgnored
```

src/DotLiquid.Tests/TemplateTests.cs
```diff
@@ -8,47 +8,56 @@ namespace DotLiquid.Tests
     [TestFixture]
     public class TemplateTests
     {
+
```
daviburg, line 11: nit: no empty line after opening class curvy bracket code block
```suggestion
```

src/DotLiquid.Tests/TemplateTests.cs
```diff
@@ -8,47 +8,56 @@ namespace DotLiquid.Tests
     [TestFixture]
     public class TemplateTests
     {
+
+        private System.Collections.Generic.List<string> TokenizeValidateBackwardCompat(string input)
```
daviburg, line 12: Nit nit: spell out words, avoid uncommon abbreviations. i.e. use Compatibility like elsewhere in the code. Yes full English words can get wordy, but modern IDEs offer auto-complete to cut down on the character typing.

src/DotLiquid/Context.cs
```diff
@@ -422,29 +436,43 @@ private object Resolve(string key, bool notifyNotFound = true)
...
+            }
+            else
+            {
+                foundVariable = TryEvaluateHashOrArrayLikeObject(scope, key, out foundValue);
             }
-            scope = scope ?? Environments.LastOrDefault() ?? Scopes.Last();
-            variable = variable ?? LookupAndEvaluate(scope, key);
 
-            variable = Liquidize(variable);
+
```
daviburg, line 469: nit: duplicate blank lines 468 and 469. Have only one blank line.
```suggestion
```

src/DotLiquid/Context.cs
```diff
@@ -633,12 +639,12 @@ private object LookupAndEvaluate(object obj, object key)
...
                 }
-                return newValue;
+                value = newValue;
             }
 
-            return value;
+            return true;
         }
-
+
```
daviburg, line 647: nit: possibly you have space characters now in this empty line.
```suggestion

```

src/DotLiquid/Hash.cs
```diff
@@ -135,31 +119,52 @@ public static Hash FromDictionary(IDictionary<string, object> dictionary)
...
+        /// Merges a dictionary object into the current <see cref="DotLiquid.Hash">Hash</see>.
+        /// </summary>
+        /// <param name="otherValues">The dictionary object to be merged into the Hash.</param>
         public void Merge(IDictionary<string, object> otherValues)
         {
             foreach (string key in otherValues.Keys)
                 _nestedDictionary[key] = otherValues[key];
         }
 
+
```
daviburg, line 162: nit: duplicate empty line

src/DotLiquid/Hash.cs
```diff
@@ -174,18 +179,26 @@ protected virtual object GetValue(string key)
             return null;
         }
 
+        /// <summary>
+        /// Provides strongly-typed access to each of the keys in the Hash
+        /// </summary>
+        /// <typeparam name="T"> A generic parameter that specifies the return type of the column.</typeparam>
```
daviburg, line 185: nit: extraneous white space at start of comment
```suggestion
        /// <typeparam name="T">A generic parameter that specifies the return type of the column.</typeparam>
```

src/DotLiquid/Hash.cs
```diff
@@ -207,21 +220,29 @@ IEnumerator IEnumerable.GetEnumerator()
...
 
+        /// <summary>
+        /// Determines whether the <see cref="DotLiquid.Hash">Hash</see> contains the specified key.
+        /// </summary>
+        /// <param name="key">The key to locate in the <see cref="DotLiquid.Hash">Hash</see></param>
+        /// <returns>true if the <see cref="DotLiquid.Hash">Hash</see> contains an element with the specified key or a default value; otherwise, false.</returns>
         public virtual bool Contains(object key)
         {
-            return ((IDictionary)_nestedDictionary).Contains(key);
+            return _lambda != null || _defaultValue != null || ((IDictionary) _nestedDictionary).Contains(key);
```
daviburg, line 236: nit: no space character after casting
```suggestion
            return _lambda != null || _defaultValue != null || ((IDictionary)_nestedDictionary).Contains(key);
```

src/DotLiquid/Tokenizer.cs
```diff
@@ -90,13 +110,66 @@ internal static List<string> Tokenize(string source)
...
+                                    break;
+                            }
+
+                            break;
+                        default:
+                            isComplete = ReadWordChar(nextVariable, markupEnumerator) && ReadToEndOfVariable(nextVariable, markupEnumerator);
+                            break;
+                    }
+
+                    if (!isComplete) //Somehow we reached the end without finding the end character(s)
```
daviburg, line 155:
```suggestion
                    if (!isComplete) // Somehow we reached the end without finding the end character(s)
```

src/DotLiquid/Tokenizer.cs
```diff
@@ -124,32 +194,100 @@ private static void ReadToEnd(StringBuilder sb, DotLiquid.Util.CharEnumerator ma
...
+                                }
+                                return true;
                             }
                             break;
                     }
                 }
             };
 
-            if (markupEnumerator.Remaining == 0) //Somehow we reached the end without finding the end character(s)
+            //Somehow we reached the end without finding the end character(s)
```
daviburg, line 227:
```suggestion
            // Somehow we reached the end without finding the end character(s)
```

src/DotLiquid/Tokenizer.cs
```diff
@@ -124,32 +194,100 @@ private static void ReadToEnd(StringBuilder sb, DotLiquid.Util.CharEnumerator ma
...
+            return false;
+        }
+
+        /// <summary>
+        /// Reads a token until, and inclusive of, the end character and advances the enumerator position
+        /// </summary>
+        /// <param name="sb">The StringBuilder to write to</param>
+        /// <param name="markupEnumerator">The string enumerator</param>
+        /// <param name="endChar">The character that indicates end of token</param>
+        /// <returns>True if reaches endChar, otherwise false</returns>
```
daviburg, line 237: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/documentation-comments#paramref
and
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/recommended-tags#see
```suggestion
        /// <returns><see langword="true"/> if reaches <paramref name="endChar"/>, otherwise <see langword="false"/></returns>
```

src/DotLiquid/Tokenizer.cs
```diff
@@ -124,32 +194,100 @@ private static void ReadToEnd(StringBuilder sb, DotLiquid.Util.CharEnumerator ma
...
             }
+            else if (!VariableSegmentRegex.IsMatch(nextChar.ToString()))
+            {
+                return false;
+            }
+
+            markupEnumerator.AppendNext(sb);
+            return true;
         }
+
```
daviburg, line 291:
```suggestion
```

daviburg (approved):

daviburg:
I'm trying to use inline suggestion for minor/cosmetic changes so you can just click on the accept button. Approved to merge with only these minor comments.

PR #443, opened 2021-07-12
Fix Iterating over Nested IDictionary objects

### LINKED ISSUE #320: Iterating over Nested Dictionary in For Loop

### LINKED ISSUE #414: how do I  loop  the Dictionary<string, object>
### Dotliquid version
2.0.366
### Expected behavior
image
please help me ,how can I loop the Data

### LINKED ISSUE #442: Duplicate item rendering of nested IDictionary objects
### Dotliquid version
2.1.469

### Expected behavior
Nested for loops should be possible for nested dictionary data

### Actual behavior
A duplicate entry is rendered for the nested dictionary, in the example below "Mike Doe (ID2)" is rendered twice for a single entry in the dictionary.
```
  Expected: "... Green (ID1), Mike Doe (ID2), \r\nMaths 2: Eric Schmidt (ID3..."
  But was:  "... Green (ID1), Mike Doe (ID2), Mike Doe (ID2), \r\nMaths 2: E..."
```

This appears to be due to `hashValue["itemName"] = key;` in For.BuildContext(), I can't find a reference to "itemName" in the Liquid documents, so a simple fix could be to remove this code - unless it is a known/used feature.

```
        [Test]
        public void TestNestedDictionaries()
        {
            var classes = new Dictionary<string, object> {{
                    "Classes", new Dictionary<string, object> {
                        { "C1", new Dictionary<string, object>{
                            { "Name", "English" },
                            { "Level", "1" },
                            { "Students", new Dictionary<string, object> {
                                { "ID1", new Dictionary<string, object>{ { "First", "Jane" }, { "Last", "Green" } } },
                                { "ID2", new Dictionary<string, object>{ { "First", "Mike" }, { "Last", "Doe" } } } } }
                            }
                        },
                        { "C2", new Dictionary<string, object>{
                            { "Name", "Maths" },
                            { "Level", "2" },
                            { "Students", new Dictionary<string, object> {
                                { "ID3", new Dictionary<string, object>{ { "First", "Eric" }, { "Last", "Schmidt" } } },
                                { "ID4", new Dictionary<string, object>{ { "First", "Bruce" }, { "Last", "Banner" } } } } }
                            }
                        }
                    }
            }};

            // Check for loops
            Helper.AssertTemplateResult(expected: @"English 1: Jane Green (ID1), Mike Doe (ID2), 
Maths 2: Eric Schmidt (ID3), Bruce Banner (ID4), 
",
                template: @"{% for class in Classes %}{{class.Name}} {{class.Level}}: {% for student in class.Students %}{{ student.First }} {{ student.Last }} ({{ student.itemName }}), {%endfor%}
{%endfor%}",
                localVariables: Hash.FromDictionary(classes));
        }
```

robin-parker:
- Added test case for multiple nested dictionary traversal in for loops.
- Removed `For.BuildContext` which is replaced by handling for nested KeyValuePair objects in Context.
- The undocumented `item.itemName` for loop property is now an alias of `KeyValuePair.Key`, and is only available when syntax is below 2.2 (resolves #442).
- Implicit access to nested dictionary values removed for syntax 2.2 and above. (resolves #320).

__Wiki updates to follow__
- [DotLiquid for Designers](https://github.com/dotliquid/dotliquid/wiki/DotLiquid-for-Designers) update on how to iterate nested for loops (resolves #414)

### DIFF
--- a/src/DotLiquid.Tests/Tags/StandardTagTests.cs
+++ b/src/DotLiquid.Tests/Tags/StandardTagTests.cs
@@ -102,11 +102,11 @@ namespace DotLiquid.Tests.Tags
         public void TestForWithNestedDictionary_DotLiquid22()
         {
             var dictionary = new Dictionary<string, object> { {
-            "People",
-            new Dictionary<string, object> {
-                    { "ID1", new Dictionary<string, object>{ { "First", "Jane" }, { "Last", "Green" } } },
-                    { "ID2", new Dictionary<string, object>{ { "First", "Mike" }, { "Last", "Doe" } } }
-                }
+                "People",
+                new Dictionary<string, object> {
+                        { "ID1", new Dictionary<string, object>{ { "First", "Jane" }, { "Last", "Green" } } },
+                        { "ID2", new Dictionary<string, object>{ { "First", "Mike" }, { "Last", "Doe" } } }
+                    }
             } };
 
             // Validate that:
@@ -645,7 +645,8 @@ namespace DotLiquid.Tests.Tags
             // - item.Key OR item[0] --> access the Key
             // - item.Value.property OR item[1].property --> access values of a nested Hash/IDictionary
             // - Pre 2.2 syntax (item.property> OR item.itemName) is ignored 
-            Helper.AssertTemplateResult(expected: @"English 1: Jane Green (ID1), Mike Doe (ID2), 
+            Helper.AssertTemplateResult(
+                expected: @"English 1: Jane Green (ID1), Mike Doe (ID2), 
 Maths 2: Eric Schmidt (ID3), Bruce Banner (ID4), 
 ",
                 template: @"{% for class in Classes %}{{class.Name}}{{class.Value.Name}} {{class.Value.Level}}: {% for student in class.Value.Students %}{{ student[1].First }} {{ student[1].Last }} ({{ student[0] }}), {%endfor%}
@@ -654,7 +655,8 @@ Maths 2: Eric Schmidt (ID3), Bruce Banner (ID4),
                 syntax: SyntaxCompatibility.DotLiquid22);
 
             // Check dot notation
-            Helper.AssertTemplateResult(expected: "Eric Schmidt",
+            Helper.AssertTemplateResult(
+                expected: "Eric Schmidt",
                 template: "{{ Classes.C2.Students.ID3.First }} {{ Classes.C2.Students.ID3.Last }}",
                 localVariables: Hash.FromDictionary(classes),
                 syntax: SyntaxCompatibility.DotLiquid22);
--- a/src/DotLiquid.Tests/Tags/StandardTagTests.cs
+++ b/src/DotLiquid.Tests/Tags/StandardTagTests.cs
@@ -88,10 +88,36 @@ namespace DotLiquid.Tests.Tags
                 }
             } };
 
-            Helper.AssertTemplateResult("JaneMike", "{% for item in People %}{{ item.First }}{%endfor%}",
-                Hash.FromDictionary(dictionary));
+            // Validate that:
+            // 1) Nested dictionary properties can be accessed with or without specifying Value.
+            // 2) itemName can still be used as an alias for Key.
+            Helper.AssertTemplateResult(
+                expected: "ID1:Jane Green,ID2:Mike Doe,",
+                template: "{% for item in People %}{{ item.itemName }}:{{ item.First }} {{ item.Value.Last }},{%endfor%}",
+                localVariables: Hash.FromDictionary(dictionary),
+                syntax: SyntaxCompatibility.DotLiquid20);
         }
 
+        [Test]
+        public void TestForWithNestedDictionary_DotLiquid22()
+        {
+            var dictionary = new Dictionary<string, object> { {
+            "People",
+            new Dictionary<string, object> {
+                    { "ID1", new Dictionary<string, object>{ { "First", "Jane" }, { "Last", "Green" } } },
+                    { "ID2", new Dictionary<string, object>{ { "First", "Mike" }, { "Last", "Doe" } } }
+                }
+            } };
+
+            // Validate that:
+            // 1) Nested dictionary properties can only be accessed when specifying Value.
+            // 2) itemName is no longer supported as an alias for Key.
+            Helper.AssertTemplateResult(
+                expected: "ID1:Jane,ID2:Mike,",
+                template: "{% for item in People %}{{ item.itemName }}{{ item.First }}{{ item.Key }}:{{ item.Value.First }},{%endfor%}",
+                localVariables: Hash.FromDictionary(dictionary),
+                syntax: SyntaxCompatibility.DotLiquid22);
+        }
 
         public class TestDictObject : Drop
         {
@@ -592,7 +618,7 @@ namespace DotLiquid.Tests.Tags
         }
 
         [Test]
-        public void TestNestedDictionaries()
+        public void TestNestedDictionaries_DotLiquid22()
         {
             var classes = new Dictionary<string, object> {{
                     "Classes", new Dictionary<string, object> {
@@ -615,21 +641,23 @@ namespace DotLiquid.Tests.Tags
                     }
             }};
 
-            // Check for loops, demonstrate values can be accessed using either:
-            // a: <item>.Value.<value-name> --> explicitly access values of a nested Hash/IDictionary
-            // b: <item>.<value-name> --> implicitly access values of a nested Hash/IDictionary (and retain backwards compatibility)
-            // c: <item>.Key --> access the Key, in case the template needs it (replacing the non-standard <item>.itemName) 
+            // Check for-loops, verify that:
+            // - item.Key OR item[0] --> access the Key
+            // - item.Value.property OR item[1].property --> access values of a nested Hash/IDictionary
+            // - Pre 2.2 syntax (item.property> OR item.itemName) is ignored 
             Helper.AssertTemplateResult(expected: @"English 1: Jane Green (ID1), Mike Doe (ID2), 
 Maths 2: Eric Schmidt (ID3), Bruce Banner (ID4), 
 ",
-                template: @"{% for class in Classes %}{{class.Name}} {{class.Value.Level}}: {% for student in class.Students %}{{ student.First }} {{ student.Value.Last }} ({{ student.Key }}), {%endfor%}
+                template: @"{% for class in Classes %}{{class.Name}}{{class.Value.Name}} {{class.Value.Level}}: {% for student in class.Value.Students %}{{ student[1].First }} {{ student[1].Last }} ({{ student[0] }}), {%endfor%}
 {%endfor%}",
-                localVariables: Hash.FromDictionary(classes));
+                localVariables: Hash.FromDictionary(classes),
+                syntax: SyntaxCompatibility.DotLiquid22);
 
             // Check dot notation
             Helper.AssertTemplateResult(expected: "Eric Schmidt",
                 template: "{{ Classes.C2.Students.ID3.First }} {{ Classes.C2.Students.ID3.Last }}",
-                localVariables: Hash.FromDictionary(classes));
+                localVariables: Hash.FromDictionary(classes),
+                syntax: SyntaxCompatibility.DotLiquid22);
         }
 
         [Test]
--- a/src/DotLiquid.Tests/Tags/StandardTagTests.cs
+++ b/src/DotLiquid.Tests/Tags/StandardTagTests.cs
@@ -81,7 +81,7 @@ namespace DotLiquid.Tests.Tags
         public void TestForWithNestedDictionary()
         {
             var dictionary = new Dictionary<string, object> { {
-            "People", 
+            "People",
             new Dictionary<string, object> {
                     { "ID1", new Dictionary<string, object>{ { "First", "Jane" }, { "Last", "Green" } } },
                     { "ID2", new Dictionary<string, object>{ { "First", "Mike" }, { "Last", "Doe" } } }
@@ -329,9 +329,9 @@ namespace DotLiquid.Tests.Tags
         [Test]
         public void TestContinueOutsideFor()
         {
-        var markup = "123{% continue %}456";
-        var expected = "123";
-        Helper.AssertTemplateResult(expected, markup);
+            var markup = "123{% continue %}456";
+            var expected = "123";
+            Helper.AssertTemplateResult(expected, markup);
         }
 
         [Test]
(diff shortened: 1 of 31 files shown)


microalps:
See #321 for discussion about whether we should even have this BuildContext method. It seems non-standard entirely.

robin-parker:
> See #321 for discussion about whether we should even have this BuildContext method. It seems non-standard entirely.

Interesting, I didn't notice that there was another ticket related to the same area. I've made a change to this PR to remove BuildContext whilst retaining backwards compatibility.

robin-parker:
@microalps - please could you review - this fixes #442 which is impacting me, and should also fix #320.

microalps:
I'm not convinced about the approach. I get we are trying to avoid a breaking change but we are creating a mess. In your C1 and C2 example, if I call .Name which name should be returned? Also, the right way will not return null when it is expected on Name (only c1 and c2 are valid)  I think we break this one case as it is invalid by all definitions of the word and make a notation in the readme. If @daviburg insists we create a 22a flag. But doing both is middle of the road - where you get run over!!

daviburg:
Help me understand the before and after, what is the current middle of the road which may look like a regression, what is the desired clean but more regression-like change?

robin-parker:
> Help me understand the before and after, what is the current middle of the road which may look like a regression, what is the desired clean but more regression-like change?
> That's the `item.itemName` was working but now no longer will?

@daviburg, the before is demonstrated in this [test case](https://github.com/dotliquid/dotliquid/blob/e18c34c619d946b9669da715fc5621dda7294c6d/src/DotLiquid.Tests/Tags/StandardTagTests.cs#L80-L93).
The `for` tag condenses a KeyValuePair where the Value is Dictionary, by moving the original Key to an entry in the nested dictionary with a new Key of `itemName` - [see here](https://github.com/dotliquid/dotliquid/blob/e18c34c619d946b9669da715fc5621dda7294c6d/src/DotLiquid/Tags/For.cs#L231). 
Unfortunately when iterating the nested dictionary that results in an extra loop (original values + itemName).

The after is to remove the `itemName` entry, and replace it with a `.Key` accessor, which fixes the issue reported in #442 - _this is potentially a breaking change, but I cannot find `itemName` documented anywhere in docs or tests_.
I have also addressed the issue raised in #320, by allowing nested dictionary values to be referenced with or without the `Value` qualifier, see [new test case in this PR](https://github.com/robin-parker/dotliquid/blob/5eb9d3e60410d9794083265d8bf022c428c0b057/src/DotLiquid.Tests/Tags/StandardTagTests.cs#L618-L621).

microalps:
> I have also addressed the issue raised in #320, by allowing nested dictionary values to be referenced with or without the Value qualifier, see new test case in this PR.

This can what I object to. If I am expecting to use .Key and .Value - why is .Prop not returning null and instead returning .Value.Prop ? Or the opposite if someone had programmed up skipping .Value and checks for .Key they will suddenly get a non null response. Middle of the road.

This is an issue which is clearly wrong in C# (and Ruby?) And just needs to be treated like the bug it is. It's not a non-standard behavior, but completely wrong. Like having a Capitalize filter that lower cases - the bug fix would be breaking but absolutely expected.

microalps:
To add some Ruby facts:
Source: https://stackoverflow.com/questions/8206869/iterate-over-hashes-in-liquid-templates
Test: https://replit.com/@microalps/JsonLiquidTemplate#main.rb

Bottom line - skipping key/value level is not valid for sure and returns blank. C# syntax makes sense to use .Key and .Value - but Ruby uses [0] for key and [1] for value.

robin-parker:
@microalps - so you're OK with everything except `item.Prop` being retained as an alias for `item.Value.Prop`, is that a fair summary?

I've put the code for backwards compatibility in Context behind a switch, so the PR strictly applies either the documented Liquid syntax (`item[0]` / `item[1]`) or C# syntax (`item.Key` / `item.Value`) from DotLiquid2.2 onwards.

Is that OK, or should I bump the version again to DotLiquid23?

microalps:
> @microalps - so you're OK with everything
In concept yes. In code, not really. I posted my counter proposal in gitter. See https://github.com/microalps/dotliquid/commit/3e15d7ca9650ffd2ec88827c76df903f4f2f0448

> Is that OK, or should I bump the version again to DotLiquid23?
DotLiquid 22a (221) would probably be safest as it isn't a major change. Was going to do 21a until we added parser versioning as well.

src/DotLiquid/Context.cs
```diff
@@ -522,67 +522,50 @@ private bool TryGetVariable(IEnumerator<string> partsEnumerator, out object vari
...
+                        @object = Liquidize(castCollection.LastOrDefault());
                     }
                 }
+                // No key was present with the desired value and it wasn't one of the directly supported
+                // keywords either. The only thing we got left is to return nil
+                else
+                {
+                    variable = null;
+                    return false;
+                }
```
microalps, line 568: Not sure what this code change has to do with the this PR, please revert or explain

src/DotLiquid/Context.cs
```diff
@@ -697,6 +680,12 @@ private static object Liquidize(object obj)
 
         private static bool IsKeyValuePair(object obj)
         {
+            return TryGetKeyValuePair(obj, out var key, out var value);
```
microalps, line 683: ```out _``` might be better ? Or just rewrite the only other reference to this function and remove it entirely?

src/DotLiquid/Context.cs
```diff
@@ -709,6 +698,8 @@ private static bool IsKeyValuePair(object obj)
                     Type baseType = valueType.GetGenericTypeDefinition();
                     if (baseType == typeof(KeyValuePair<,>))
                     {
+                        key = valueType.GetRuntimeProperty("Key").GetValue(obj);
+                        value = valueType.GetRuntimeProperty("Value").GetValue(obj);
```
microalps, line 702: As mentioned, I don't see any gain by assigning both by reflection when we need 1 or even none for that execution block. Will get back with actual perf comparisons

src/DotLiquid/Tags/For.cs
```diff
@@ -223,18 +221,43 @@ private static List<object> SliceCollectionUsingEach(Context context, IEnumerabl
...
+
+    /// <summary>
+    /// internal class to encapsulate pre DotLiquid 2.2 compatibility in for loop.
+    /// </summary>
+    /// <remarks>
+    /// DotLiquid 2.2 compatibility includes:
+    /// * An undocumented property `itemName` which is an implicit alias for KeyValuePair.Key
+    /// * Implicit access of properties within a nested IDictionary Value
+    /// </remarks>
+    class LegacyKeyValueDrop : Drop
```
microalps, line 234: should this be marked internal?

microalps (changes requested):

src/DotLiquid/Tags/For.cs
robin-parker: Isn't internal the default visibility for a class?

src/DotLiquid/Context.cs
robin-parker: Thanks, I haven't seen the discard feature before.

src/DotLiquid/Context.cs
robin-parker: These changes were made for pre 2.2 combability before implementing the LegacyKeyValueDrop.
I retained them because they:
- update misleading comments to reflect the current code behavior (several comments appeared to be cut/pasted in error).
- merged if statements originally on l535 and l547 into a single block to remove extra `is KeyValuePair`.
- remove a redundant else block (l543).
- remove unreachable branch (l572) caused by the if statement (l562 ensure values can only be size/first/last).
- remove unused variable assigns (l530, 538, 569, 574)

src/DotLiquid/Tags/For.cs
daviburg: Better make it explicit that leave the reader to guess.

src/DotLiquid.Tests/Tags/StandardTagTests.cs
```diff
@@ -88,10 +88,36 @@ public void TestForWithNestedDictionary()
...
+                expected: "ID1:Jane Green,ID2:Mike Doe,",
+                template: "{% for item in People %}{{ item.itemName }}:{{ item.First }} {{ item.Value.Last }},{%endfor%}",
+                localVariables: Hash.FromDictionary(dictionary),
+                syntax: SyntaxCompatibility.DotLiquid20);
         }
 
+        [Test]
+        public void TestForWithNestedDictionary_DotLiquid22()
+        {
+            var dictionary = new Dictionary<string, object> { {
```
daviburg, line 104: nit: alignment of the dictionary initializer doesn't look right

src/DotLiquid.Tests/Tags/StandardTagTests.cs
```diff
@@ -591,6 +617,49 @@ public void TestForReversed()
...
+                            }
+                        }
+                    }
+            }};
+
+            // Check for-loops, verify that:
+            // - item.Key OR item[0] --> access the Key
+            // - item.Value.property OR item[1].property --> access values of a nested Hash/IDictionary
+            // - Pre 2.2 syntax (item.property> OR item.itemName) is ignored 
+            Helper.AssertTemplateResult(expected: @"English 1: Jane Green (ID1), Mike Doe (ID2),
```
daviburg, line 648:
```suggestion
            Helper.AssertTemplateResult(
                expected: @"English 1: Jane Green (ID1), Mike Doe (ID2), 
```

src/DotLiquid.Tests/Tags/StandardTagTests.cs
```diff
@@ -591,6 +617,49 @@ public void TestForReversed()
...
+            Helper.AssertTemplateResult(expected: @"English 1: Jane Green (ID1), Mike Doe (ID2), 
+Maths 2: Eric Schmidt (ID3), Bruce Banner (ID4), 
+",
+                template: @"{% for class in Classes %}{{class.Name}}{{class.Value.Name}} {{class.Value.Level}}: {% for student in class.Value.Students %}{{ student[1].First }} {{ student[1].Last }} ({{ student[0] }}), {%endfor%}
+{%endfor%}",
+                localVariables: Hash.FromDictionary(classes),
+                syntax: SyntaxCompatibility.DotLiquid22);
+
+            // Check dot notation
+            Helper.AssertTemplateResult(expected: "Eric Schmidt",
```
daviburg, line 657:
```suggestion
            Helper.AssertTemplateResult(
                expected: "Eric Schmidt",

```

src/DotLiquid/Context.cs
microalps: I reviewed in a compare tool. Much easier to understand. Marking resolved

microalps: I reviewed in a compare tool. Much easier to understand. Marking resolved

microalps: I reviewed in a compare tool. Much easier to understand. Marking resolved

src/DotLiquid/Context.cs
daviburg: @microalps this is the last issue with this PR, right?
@robin-parker would it be possible to have more specialized helpers to avoid any performance impact from running reflection which may not be required by the caller?

robin-parker: I've reverted to `IsKeyValuePair` and added a new helper method `ObjectExtensionMethods.GetPropertyValue` to extract the Key/Values by reflection when needed.

src/DotLiquid/Util/ObjectExtensionMethods.cs
```diff
@@ -48,6 +55,25 @@ public static object Send(this object value, string member, object[] parameters
...
+        /// </summary>
+        /// <param name="object">an object to be interrogated</param>
+        /// <param name="propertyName">the name of the property to get</param>
+        /// <returns>The property value if it exists, otherwise null</returns>
+        public static object GetPropertyValue(this object @object, string propertyName)
+        {
+            if (@object is null)
+                return null;
+
+            Type type = @object.GetType();
```
daviburg, line 69:
```suggestion
            var type = @object.GetType();

```

src/DotLiquid/Util/ObjectExtensionMethods.cs
```diff
@@ -48,6 +55,25 @@ public static object Send(this object value, string member, object[] parameters
...
+            PropertyInfo propertyInfo = type.GetRuntimeProperty(propertyName);
```
daviburg, line 70:
```suggestion
            var propertyInfo = type.GetRuntimeProperty(propertyName);

```

src/DotLiquid/Util/ObjectExtensionMethods.cs
```diff
@@ -48,6 +55,25 @@ public static object Send(this object value, string member, object[] parameters
...
+            if (propertyInfo != null)
+                return propertyInfo.GetValue(@object, null);
```
daviburg, line 72: The whole thing can be replaced  by one-liner without if:
```
                return propertyInfo?.GetValue(@object, null);
```

daviburg: Actually...
```
                return @object.GetType().GetRuntimeProperty(propertyName)?.GetValue(@object, null);
```

daviburg (approved):

daviburg:
@microalps see if you can also sign-off this one.

src/DotLiquid/Context.cs
microalps: Thanks, I can also confirm this PR has no performance change. Signing off.

microalps (approved):

PR #444, opened 2021-07-12
Shopify filters

robin-parker:
Add some Shopify specific filters to complement the existing Liquid filters.
- [String filters](https://shopify.dev/api/liquid/filters/string-filters): md5, sha1, sha256, hmac_sha1, hmac_sha256

Added a new `addfilters` tag that can be used in templates to add additional/optional filters to the render context.


microalps:
I would not approve this PR on the basis of the added dependency. Many people prefer, or already use, Newtonsoft.Json, especially those not on CORE. We should discuss this in Gitter before spending too much time on it.

src/DotLiquid/Liquid.cs
```diff
@@ -55,6 +55,7 @@ static Liquid()
             Template.RegisterTag<Tags.Html.TableRow>("tablerow");
 
             Template.RegisterFilter(typeof(StandardFilters));
+            Template.RegisterFilter(typeof(ShopifyFilters));
```
microalps, line 58: As discussed on gitter, this should not be part of the library. MS will make a way to opt in outside of DLL.

src/DotLiquid.Tests/ShopifyFiltersTests.cs
```diff
@@ -0,0 +1,183 @@
...
+        [TestCase(123, "123")]
+        [TestCase(123.12, "123.12")]
+        [TestCase(-123.12, "-123.12")]
+        [TestCase(null, null)]
+        [TestCase("", "\"\"")]
+        public void TestJson(object value, string expected)
+        {
+            Assert.AreEqual(expected, ShopifyFilters.Json(value));
+        }
+#endif
```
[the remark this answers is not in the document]
robin-parker: Just a preference to protect against unintended insertion of new test case in between tests spanning a single #if block.

src/DotLiquid/Liquid.cs
robin-parker: How about an alternative, which would enable a Designer to add a non-standard filter class to the render context.

For example: `{%-register_filters 'DotLiquid.ShopifyFilters'-%}`
The `register_filters` tag would add the DotLiquid.ShopifyFilters class using Context.AddFilters during rendering.

src/DotLiquid.Tests/ShopifyFiltersTests.cs
```diff
@@ -0,0 +1,183 @@
...
+                localVariables: Hash.FromAnonymousObject(new { comment = new { email = " Joe.Bloggs@Shopify.com " } }));
+        }
+
+        [Test]
+        public void TestSha1()
+        {
+            Assert.AreEqual(null, ShopifyFilters.Sha1(null));
+            Assert.AreEqual("", ShopifyFilters.Sha1(""));
+            Assert.AreEqual(" ", ShopifyFilters.Sha1(" "));
+            Assert.AreEqual("c7322e3812d3da7bc621300ca1797517c34f63b6",
```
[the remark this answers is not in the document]
robin-parker: Fixed

src/DotLiquid/ShopifyFilters.cs
```diff
@@ -0,0 +1,112 @@
...
+        /// </summary>
+        /// <see href="https://shopify.dev/docs/themes/liquid/reference/filters/string-filters#hmac_sha1" />
+        /// <param name="input" />
+        /// <param name="secretKey" />
+        public static string HmacSha1(string input, string secretKey)
+        {
+            if (input.IsNullOrWhiteSpace() || secretKey.IsNullOrWhiteSpace())
+                return input;
+
+            var hmac = new HMACSHA1(Encoding.UTF8.GetBytes(secretKey));
```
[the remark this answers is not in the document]
robin-parker: I've changed the guard condition, secretKey should support the full range of UTF-8 characters.
Does this cover the case you were thinking of?

src/DotLiquid/ShopifyFilters.cs
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        /// <summary>
+        /// Generates a Hexadecimal hash of the provided input string.
+        /// </summary>
+        /// <param name="hashAlgorithm" />
+        /// <param name="input" />
+        private static string ComputeHash(HashAlgorithm hashAlgorithm, string input)
+        {
+            if (input.IsNullOrWhiteSpace())
```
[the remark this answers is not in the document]
robin-parker: Changed

src/DotLiquid/Tags/RegisterFilters.cs
```diff
@@ -0,0 +1,49 @@
...
+            Match syntaxMatch = Syntax.Match(markup);
+            if (syntaxMatch.Success)
+                className = syntaxMatch.Groups[1].Value.Trim(new[] { '\'', '"' });
+
+            // The class name must be non-empty
+            if (string.IsNullOrWhiteSpace(className))
+                throw new SyntaxException(Liquid.ResourceManager.GetString("SimpleTagSyntaxException"), new[] { tagName });
+
+            // Attempt to find the class type
+            classType = Type.GetType(className);
```
microalps, line 32: This is a security vulnerability. Potentially a template designer can access any (public?) class in .NET and call any method within that accepts one or more parameters. Without a dev specifying a whitelist of registerable filter classes (Shopify can be added by default to that whitelist) - and in that case we don't need reflection.

src/DotLiquid.Tests/Tags/RegisterFiltersTests.cs
```diff
@@ -0,0 +1,43 @@
...
+
+namespace DotLiquid.Tests.Tags
+{
+    [TestFixture]
+    public class RegisterFiltersTests
+    {
+        [Test]
+        public void TestInitialise()
+        {
+            new RegisterFilters().Initialize("register_filters", "DotLiquid.ShopifyFilters", null); // no quotes
```
microalps: Why should no quotes be accepted as if it was quoted? This stops the use of variables in the name - which if you think there is no purpose to then we don't need quoted versions. I vote supporting quoted versions only to be consistent with expected syntaxes in tags. Also, there is no explicit test here, just testing that no error is thrown?

src/DotLiquid/Liquid.cs
```diff
@@ -51,6 +51,7 @@ static Liquid()
             Template.RegisterTag<Tags.Literal>("literal");
             Template.RegisterTag<Tags.Unless>("unless");
             Template.RegisterTag<Tags.Raw>("raw");
+            Template.RegisterTag<Tags.RegisterFilters>("register_filters");
```
microalps, line 54: This would be the first snake case tag. I see IfChanged is not snake case - and that is a standard tag apparently. It doesn't seem like registerfilters should be snake case.

microalps (changes requested):

src/DotLiquid/Tags/RegisterFilters.cs
robin-parker: Good point a white list makes sense, I've added one by default containing ShopifyFilters.

If this PR is approved I'll update the wiki with a section describing how to use `addfilters` including a developer section on how to whitelist new filters.

src/DotLiquid/Liquid.cs
robin-parker: Renamed to `addfilters` to prevent conflation with Template.RegisterFilters()

src/DotLiquid.Tests/Tags/RegisterFiltersTests.cs
robin-parker: Modified to render markup using the Variable class.

src/DotLiquid.Tests/Tags/AddFiltersTests.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        [Test]
+        public void TestWhitelisted()
+        {
+            // addfilter is included, so ensure the value is sha1 hashed.
+            Helper.AssertTemplateResult(
+                expected: @"
+My encoded string is: c7322e3812d3da7bc621300ca1797517c34f63b6",
+                template: @"{%addfilters 'shopifyFilters'%}{% assign my_secret_string = ""ShopifyIsAwesome!"" | sha1 %}
```
microalps, line 38: why does this work with camelCase shopifyFilters and not ShopifyFilters? If it isn't case sensitive, we should have tests that validate that. If it is, not clear why it works.

src/DotLiquid.Tests/Tags/AddFiltersTests.cs
```diff
@@ -0,0 +1,52 @@
...
+My encoded string is: {{ my_secret_string }}");
+        }
+
+        [TestCase("'DotLiquid.ShopifyFilters'")]
+        [TestCase("'DotLiquid.Template'")]
+        [TestCase("'Template'")]
+        public void TestNotWhitelisted(string template)
```
microalps, line 45: Test cases are never used

src/DotLiquid/Tags/AddFilters.cs
```diff
@@ -0,0 +1,74 @@
...
+
+            alias = syntaxMatch.Groups[1].Value;
+
+            base.Initialize(tagName, markup, tokens);
+        }
+
+        /// <inheritdoc/>
+        public override void Render(Context context, TextWriter result)
+        {
+            var renderedAlias = getAliasValue(context);
```
microalps, line 54: What's wrong with context[alias] ? This is how include tag works - context handles quotes as literal and without as a variable. The difference, off the top of my head, is that new Variable will allow you to specify a filter as in 
```
{% addfilters unknown_variable | default: 'ShopifyFilters' %} 
```
which I don't think you were trying to do

src/DotLiquid/Tags/AddFilters.cs
```diff
@@ -0,0 +1,74 @@
...
+    /// {%- addfilters 'ShopifyFilters' -%}
+    /// </summary>
+    public class AddFilters : Tag
+    {
+        private static readonly Regex Syntax = R.B(R.Q(@"\s*([""'].+[""'])\s*"));
+
+        private static readonly IDictionary<string, Type> FILTER_WHITELIST = new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase);
+        static AddFilters()
+        {
+            WhitelistFilter(typeof(ShopifyFilters));
```
microalps, line 23: Would this be easier to find if it was in Liquid.cs together with StandardFilters?
```
Template.RegisterFilter(typeof(StandardFilters));
AddFilters.WhitelistFilter(typeof(ShopifyFilters)); //Don't register it but whitelist it for template authors
```
This way when my PR merges it can also add the ExtendedFilters to the whitelist in the same place.

src/DotLiquid/Tags/AddFilters.cs
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        /// <summary>
+        /// Add the provided class to the whitelist of classes that can be added by a Liquid Designer
+        /// </summary>
+        /// <param name="filterClassType">A class containing filter operations.</param>
+        /// <param name="alias">An alias for the class, if not provided the classes short name is used</param>
+        public static void WhitelistFilter(Type filterClassType, string alias = null)
```
microalps, line 31: AddFilters is plural because a class can contain multiple filters. So why is this method singular? Or even simply AddFilters.Whitelist ?

src/DotLiquid.Tests/Tags/AddFiltersTests.cs
robin-parker: Intended to check mixed-case, I've added cases for clarity.

src/DotLiquid.Tests/Tags/AddFiltersTests.cs
robin-parker: Fixed to pass template parameter into AddFilters.Initialize().

src/DotLiquid/Tags/AddFilters.cs
robin-parker: I based AddFilters on Assign, which uses new Variable. I've simplified it now to use context[alias].toString()

src/DotLiquid/Tags/AddFilters.cs
robin-parker: Moved

src/DotLiquid/Tags/AddFilters.cs
robin-parker: Changed to Whitelist.
I originally opted for `AddFilters.WhitelistFilter` over `Template.WhitelistFilter` as it is feels like better encapsulation and avoids adding more complexity in Template.

microalps:
I think I am OK with this PR as is. It is lacking some tests on the add filter tag (e.g. non quoted variables, alias use, double registration of same class, filter use before registration)

src/DotLiquid/ShopifyFilters.cs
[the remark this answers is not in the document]
microalps: That wasn't what I was thinking but more along how does I use a secret key that isn't a string? Like ascii 0-32 and 120s and up e.g. secretKey: 0x017F. But your's is a good catch none the less I'll give you credit. It is clearly out of scope of this PR and a fundamental gap in the liquid language.

robin-parker:
> I think I am OK with this PR as is. It is lacking some tests on the add filter tag (e.g. non quoted variables, alias use [codecov shows that line as yellow partially covered], double registration of same class, filter use before registration). When you are ready, mark this PR as ready for review and tag daviburg.
> 
> For record, I still don't like the idea of adding dependency for json within the standard library.

@microalps, I've added a few test cases to cover the examples you you listed.
Any suggestions on how else to deal with the dependency, would we be better off removing the `json` filter to avoid the dependency, or adding another tfm for a version of .net that includes `System.Text.Json`?

src/DotLiquid/DotLiquid.csproj
```diff
@@ -41,6 +41,10 @@
     </None>
   </ItemGroup>
 
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
+    <PackageReference Include="System.Text.Json" Version="4.7.2" />
```
daviburg, line 45: Create a separate assembly project to add new dependency to Json so client can opt-in. Move the new filter(s) depending on Json there.

daviburg: The new json filter will be much welcome by users of Logic App. However we need to introduce this via a separate assembly so we don't force it on other consumers.
Technically you may also need to generate a separate nuget package so you can declare the new nuget dependency separately as well.

daviburg:
I didn't do a code review, exceeded my time to spend on dot liquid for a while. Please ping me again in a few days.

src/DotLiquid/DotLiquid.csproj
microalps: I'm not familiar with NuGet policies, but will we simply be able to add a new target in appveyor for a new nuspec package and upload with the existing API key? Or will @tgjones need to do something first on NuGet to register the name / allow publish via API?

robin-parker: I've removed the dependency and the Json filter from the PR.

src/DotLiquid/DotLiquid.csproj
```diff
@@ -28,7 +28,7 @@
     <GenerateAssemblyInformationalVersionAttribute>false</GenerateAssemblyInformationalVersionAttribute>
     <DebugType>full</DebugType>
     <DebugSymbols>True</DebugSymbols>
-      <DocumentationFile>bin\Debug\$(TargetFramework)\DotLiquid.xml</DocumentationFile>
+    <DocumentationFile>bin\Debug\$(TargetFramework)\DotLiquid.xml</DocumentationFile>
```
microalps, line 31: nit: what's this change?

src/DotLiquid/Tags/AddFilters.cs
```diff
@@ -0,0 +1,59 @@
...
+    /// <summary>
+    /// Adds filters in a whitelisted class to the current context.
+    ///
+    /// {%- addfilters 'ShopifyFilters' -%}
+    /// </summary>
+    public class AddFilters : Tag
+    {
+        private static readonly Regex Syntax = R.B(R.Q(@"\s*([""'].+[""'])\s*"));
+
+        private static readonly IDictionary<string, Type> WHITELIST = new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase);
```
microalps, line 19: nit: non-standard case of whitelist

src/DotLiquid/DotLiquid.csproj
robin-parker: Format on save fixed the whitespace when I added, then removed the Json dependency.

src/DotLiquid/Tags/AddFilters.cs
robin-parker: Changed to Whitelist.

src/DotLiquid/Tags/AddFilters.cs
```diff
@@ -16,7 +16,7 @@ public class AddFilters : Tag
     {
         private static readonly Regex Syntax = R.B(R.Q(@"\s*([""'].+[""'])\s*"));
 
-        private static readonly IDictionary<string, Type> Whitelist = new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase);
+        private static readonly IDictionary<string, Type> _Whitelist = new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase);
```
daviburg, line 19: Why the extra '_'? I think I saw that on some of the code in this repo, but that's rather old school arcane flagging of variable names. And then, why is the also static readonly Syntax about not getting an underscore prefix?

If we want clarity when using the variables whether those are instance variable or static, we should prefix the call with `this.` or with the class name `AddFilters.`.

src/DotLiquid.Tests/Helper.cs
```diff
@@ -51,15 +52,20 @@ public static void AssertTemplateResult(string expected, string template, INamin
...
             var parameters = new RenderParameters(System.Globalization.CultureInfo.CurrentCulture)
             {
                 LocalVariables = localVariables,
-                SyntaxCompatibilityLevel = syntax
+                SyntaxCompatibilityLevel = syntax,
+                Filters = localFilters
             };
             Assert.AreEqual(expected, Template.Parse(template).Render(parameters));
         }
+        public static void AssertTemplateResult(string expected, string template, Hash localVariables, SyntaxCompatibility syntax = SyntaxCompatibility.DotLiquid20)
```
daviburg, line 65: Need a new line above
```suggestion

        public static void AssertTemplateResult(string expected, string template, Hash localVariables, SyntaxCompatibility syntax = SyntaxCompatibility.DotLiquid20)
```

src/DotLiquid/ShopifyFilters.cs
```diff
@@ -0,0 +1,97 @@
...
+        public static string Md5(string input)
+        {
+            return input.IsNullOrWhiteSpace() ? input : ComputeHash(MD5.Create(), input);
+        }
+
+        /// <summary> 
+        /// Converts a string into a SHA-1 hash.
+        /// </summary>
+        /// <see href="https://shopify.dev/docs/themes/liquid/reference/filters/string-filters#sha1" />
+        /// <param name="input"></param>
```
daviburg, line 27:
```suggestion
        /// <param name="input">The input to be encoded</param>
```

src/DotLiquid/ShopifyFilters.cs
```diff
@@ -0,0 +1,97 @@
...
+    /// Shopify filters implemented by DotLiquid
+    /// </summary>
+    /// <see href="https://shopify.dev/docs/themes/liquid/reference/filters"/>
+    public static class ShopifyFilters
+    {
+        /// <summary>
+        /// Converts a string into an MD5 hash.
+        /// </summary>
+        /// <see href="https://shopify.dev/docs/themes/liquid/reference/filters/string-filters#md5"/>
+        /// <param name="input"/>
```
daviburg, line 17:
```suggestion
        /// <param name="input">The input to be encoded</param>
```

src/DotLiquid/ShopifyFilters.cs
```diff
@@ -0,0 +1,97 @@
...
+        public static string Sha1(string input)
+        {
+            return input.IsNullOrWhiteSpace() ? input : ComputeHash(SHA1.Create(), input);
+        }
+
+        /// <summary>
+        /// Converts a string into a SHA-256 hash.
+        /// </summary>
+        /// <see href="https://shopify.dev/docs/themes/liquid/reference/filters/string-filters#sha256" />
+        /// <param name="input"></param>
```
daviburg, line 37:
```suggestion
        /// <param name="input">The input to be encoded</param>
```

src/DotLiquid/ShopifyFilters.cs
```diff
@@ -0,0 +1,97 @@
...
+        {
+            return input.IsNullOrWhiteSpace() ? input : ComputeHash(SHA256.Create(), input);
+        }
+
+        /// <summary>
+        /// Converts a string into a SHA-1 hash using a hash message authentication code (HMAC).
+        /// Pass the secret key for the message as a parameter to the filter.
+        /// </summary>
+        /// <see href="https://shopify.dev/docs/themes/liquid/reference/filters/string-filters#hmac_sha1" />
+        /// <param name="input" />
```
daviburg, line 48: /// <param name="input">The input to be encoded</param>

src/DotLiquid/ShopifyFilters.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// <param name="secretKey" />
```
daviburg, line 49:
```suggestion
        /// <param name="secretKey">The secret key</param>
```

src/DotLiquid/ShopifyFilters.cs
```diff
@@ -0,0 +1,97 @@
...
+        public static string HmacSha1(string input, string secretKey)
+        {
+            if (input.IsNullOrWhiteSpace() || secretKey is null)
+                return input;
+
+            var hmac = new HMACSHA1(Encoding.UTF8.GetBytes(secretKey));
```
daviburg, line 55: This class implements dispose interface. You need a using block everywhere you create an instance.

src/DotLiquid/ShopifyFilters.cs
```diff
@@ -0,0 +1,97 @@
...
+    public static class ShopifyFilters
+    {
+        /// <summary>
+        /// Converts a string into an MD5 hash.
+        /// </summary>
+        /// <see href="https://shopify.dev/docs/themes/liquid/reference/filters/string-filters#md5"/>
+        /// <param name="input"/>
+        public static string Md5(string input)
+        {
+            return input.IsNullOrWhiteSpace() ? input : ComputeHash(MD5.Create(), input);
```
daviburg, line 20: `HashAlgorithm : IDisposable, `
so you always need to dispose the classes that you pass in to this helper
At least for .NET Core
https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.hashalgorithm?view=net-5.0

src/DotLiquid/Tags/AddFilters.cs
```diff
@@ -0,0 +1,59 @@
...
+        {
+            _Whitelist[alias ?? filterClassType.Name] = filterClassType;
+        }
+
+        private string alias;
+
+        /// <inheritdoc/>
+        public override void Initialize(string tagName, string markup, List<string> tokens)
+        {
+            Match syntaxMatch = Syntax.Match(markup);
```
daviburg, line 36:
```suggestion
            var syntaxMatch = Syntax.Match(markup);
```

src/DotLiquid/Tags/AddFilters.cs
```diff
@@ -0,0 +1,59 @@
...
+            if (!syntaxMatch.Success)
+                throw new SyntaxException(Liquid.ResourceManager.GetString("AddFiltersTagSyntaxException"));
+
+            alias = syntaxMatch.Groups[1].Value;
```
daviburg, line 40:
```suggestion
            this.alias = syntaxMatch.Groups[1].Value;
```

src/DotLiquid/Tags/AddFilters.cs
```diff
@@ -0,0 +1,59 @@
...
+
+            base.Initialize(tagName, markup, tokens);
```
daviburg, line 42:
```suggestion
            base.Initialize(tagName: tagName, markup: markup, tokens: tokens);
```

src/DotLiquid/Tags/AddFilters.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        private static readonly Regex Syntax = R.B(R.Q(@"\s*([""'].+[""'])\s*"));
+
+        private static readonly IDictionary<string, Type> _Whitelist = new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase);
+
+        /// <summary>
+        /// Add the provided class to the whitelist of classes that can be added by a Liquid Designer
+        /// </summary>
+        /// <param name="filterClassType">A class containing filter operations.</param>
+        /// <param name="alias">An alias for the class, if not provided the classes short name is used</param>
```
daviburg, line 25:
```suggestion
        /// <param name="alias">An alias for the class, if not provided the class' short name is used</param>
```

daviburg (changes requested):

src/DotLiquid/Tags/AddFilters.cs
robin-parker: The static dictionary Whitelist clashes with the method Whitelist.
I agree the `_` is old school, so I've made a rename.

src/DotLiquid/ShopifyFilters.cs
robin-parker: Fixed

src/DotLiquid/ShopifyFilters.cs
robin-parker: Fixed.

src/DotLiquid/Tags/AddFilters.cs
```diff
@@ -0,0 +1,59 @@
...
+    /// <summary>
+    /// Adds filters in a whitelisted class to the current context.
+    ///
+    /// {%- addfilters 'ShopifyFilters' -%}
+    /// </summary>
+    public class AddFilters : Tag
+    {
+        private static readonly Regex Syntax = R.B(R.Q(@"\s*([""'].+[""'])\s*"));
+
+        private static readonly IDictionary<string, Type> Whitelisted = new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase);
```
daviburg, line 19: Oh, I just remembered, we should avoid this term altogether as it is offensive. Please use Allowed, Allowlist, Allowlisted, etc.

daviburg: Just this one change (albeit in all locations) and I am ready to sign-off.

daviburg (approved):

daviburg:
@microalps could you sign-off also? Your status still shows requesting changes.

src/DotLiquid/ShopifyFilters.cs
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        /// <summary>
+        /// Converts a string into a SHA-256 hash.
+        /// </summary>
+        /// <see href="https://shopify.dev/docs/themes/liquid/reference/filters/string-filters#sha256" />
+        /// <param name="input">The input to be encoded</param>
+        public static string Sha256(string input)
+        {
+            if (input.IsNullOrWhiteSpace())
```
microalps, line 52:
```suggestion
            if (input == null)
```

src/DotLiquid/ShopifyFilters.cs
```diff
@@ -0,0 +1,119 @@
...
+        /// <summary>
+        /// Converts a string into a SHA-1 hash using a hash message authentication code (HMAC).
+        /// Pass the secret key for the message as a parameter to the filter.
+        /// </summary>
+        /// <see href="https://shopify.dev/docs/themes/liquid/reference/filters/string-filters#hmac_sha1" />
+        /// <param name="input">The input to be encoded</param>
+        /// <param name="secretKey">The secret key</param>
+        public static string HmacSha1(string input, string secretKey)
+        {
+            if (input.IsNullOrWhiteSpace() || secretKey is null)
```
microalps, line 70:
```suggestion
            if (input == null || secretKey == null)
```
(just for consistency replaced is with ==)

src/DotLiquid/ShopifyFilters.cs
```diff
@@ -0,0 +1,119 @@
...
+        /// <summary>
+        /// Converts a string into a SHA-256 hash using a hash message authentication code (HMAC).
+        /// Pass the secret key for the message as a parameter to the filter.
+        /// </summary>
+        /// <see href="https://shopify.dev/docs/themes/liquid/reference/filters/string-filters#hmac_sha256" />
+        /// <param name="input">The input to be encoded</param>
+        /// <param name="secretKey">The secret key</param>
+        public static string HmacSha256(string input, string secretKey)
+        {
+            if (input.IsNullOrWhiteSpace() || secretKey is null)
```
microalps, line 89:
```suggestion
            if (input == null || secretKey is null)
```

src/DotLiquid.Tests/ShopifyFiltersTests.cs
```diff
@@ -0,0 +1,115 @@
...
+{
+    [TestFixture]
+    public class ShopifyFiltersTests
+    {
+        [Test]
+        public void TestMd5()
+        {
+            Assert.AreEqual(null, ShopifyFilters.Md5(null));
+            Assert.AreEqual("", ShopifyFilters.Md5(""));
+            Assert.AreEqual(" ", ShopifyFilters.Md5(" "));
```
microalps, line 14:
```suggestion
            Assert.AreEqual("d41d8cd98f00b204e9800998ecf8427e", ShopifyFilters.Md5(""));
```

src/DotLiquid.Tests/ShopifyFiltersTests.cs
```diff
@@ -0,0 +1,115 @@
...
+                localVariables: Hash.FromAnonymousObject(new { comment = new { email = " Joe.Bloggs@Shopify.com " } }),
+                localFilters: new[] { typeof(ShopifyFilters) });
+        }
+
+        [Test]
+        public void TestSha1()
+        {
+            Assert.AreEqual(null, ShopifyFilters.Sha1(null));
+            Assert.AreEqual("", ShopifyFilters.Sha1(""));
+            Assert.AreEqual(" ", ShopifyFilters.Sha1(" "));
```
microalps, line 31:
```suggestion
            Assert.AreEqual("da39a3ee5e6b4b0d3255bfef95601890afd80709", ShopifyFilters.Sha1(""));
```

src/DotLiquid.Tests/ShopifyFiltersTests.cs
```diff
@@ -0,0 +1,115 @@
...
+                localVariables: null,
+                localFilters: new[] { typeof(ShopifyFilters) });
+        }
+
+        [Test]
+        public void TestSha256()
+        {
+            Assert.AreEqual(null, ShopifyFilters.Sha256(null));
+            Assert.AreEqual("", ShopifyFilters.Sha256(""));
+            Assert.AreEqual(" ", ShopifyFilters.Sha256(" "));
```
microalps, line 50:
```suggestion
            Assert.AreEqual("e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855", ShopifyFilters.Sha256(""));
```

src/DotLiquid.Tests/ShopifyFiltersTests.cs
```diff
@@ -0,0 +1,115 @@
...
+
+        [Test]
+        public void TestHmacSha1()
+        {
+            Assert.AreEqual(null, ShopifyFilters.HmacSha1(null, null));
+            Assert.AreEqual("", ShopifyFilters.HmacSha1("", null));
+            Assert.AreEqual(" ", ShopifyFilters.HmacSha1(" ", null));
+            Assert.AreEqual(null, ShopifyFilters.HmacSha1(null, ""));
+            Assert.AreEqual("", ShopifyFilters.HmacSha1("", ""));
+            Assert.AreEqual("", ShopifyFilters.HmacSha1("", " "));
```
microalps, line 72:
```suggestion
            Assert.AreEqual(null, ShopifyFilters.HmacSha1(null, ""));
            Assert.AreEqual("fbdb1d1b18aa6c08324b7d64b71fb76370690e1d", ShopifyFilters.HmacSha1("", ""));
```

src/DotLiquid.Tests/ShopifyFiltersTests.cs
```diff
@@ -0,0 +1,115 @@
...
+
+        [Test]
+        public void TestHmacSha256()
+        {
+            Assert.AreEqual(null, ShopifyFilters.HmacSha256(null, null));
+            Assert.AreEqual("", ShopifyFilters.HmacSha256("", null));
+            Assert.AreEqual(" ", ShopifyFilters.HmacSha256(" ", null));
+            Assert.AreEqual(null, ShopifyFilters.HmacSha256(null, ""));
+            Assert.AreEqual("", ShopifyFilters.HmacSha256("", ""));
+            Assert.AreEqual("", ShopifyFilters.HmacSha256("", " "));
```
microalps, line 98:
```suggestion
            Assert.AreEqual(null, ShopifyFilters.HmacSha256(null, ""));
            Assert.AreEqual("b613679a0814d9ec772f95d778c35fc5ff1697c493715653c6c712144292c5ad", ShopifyFilters.HmacSha256("", ""));
```

src/DotLiquid/ShopifyFilters.cs
```diff
@@ -0,0 +1,119 @@
...
+        /// <summary>
+        /// Converts a string into a SHA-256 hash using a hash message authentication code (HMAC).
+        /// Pass the secret key for the message as a parameter to the filter.
+        /// </summary>
+        /// <see href="https://shopify.dev/docs/themes/liquid/reference/filters/string-filters#hmac_sha256" />
+        /// <param name="input">The input to be encoded</param>
+        /// <param name="secretKey">The secret key</param>
+        public static string HmacSha256(string input, string secretKey)
+        {
+            if (input == null || secretKey is null)
```
microalps, line 89:
```suggestion
            if (input == null || secretKey == null)
```

microalps (changes requested):
Updated tests for blank inputs. Removed unnecessary tests for space as input - if blank is accepted then certainly a space will be.

robin-parker:
> Updated tests for blank inputs. Removed unnecessary tests for space as input - if blank is accepted then certainly a space will be.

@microalps, thanks for updating the tests. I'm out on vacation, so didn't have access to make the changes myself.

microalps (approved):

PR #452, opened 2021-08-24
Strip HTML Ruby Behavior Sync

### LINKED ISSUE #342: strip_html leaves script, comment, and style HTML
### Dotliquid version
2.0.145.0

### Expected behavior
I expect this liquid code to render only whitespace.
```
{{ "<!-- <p>This comment is not handled correctly</p> -->" | strip_html }}
{{ "<script>var foo = 'This should have gotten stripped out.';</script>" | strip_html }}
```

### Actual behavior

Instead I get the following
```
This comment is not handled correctly -->
var foo = 'This should have gotten stripped out.';
```

### Details

Shopify seems to have regular expressions for handling `comments`, `script`, and `style` tags explicity: https://github.com/Shopify/liquid/blob/b3b63a683f6daaea7e70f1602ee829abdf970ef3/lib/liquid/standardfilters.rb#L14-L18

Aside: It's somewhat surprising to see regular expressions used to parse the HTML since conventional wisdom is to avoid doing so, e.g. https://stackoverflow.com/questions/1732348/regex-match-open-tags-except-xhtml-self-contained-tags. I wonder if something like [HTML Agility Pack](https://www.nuget.org/packages/HtmlAgilityPack/) could work better here. That said, "stripping" HTML is not the same as parsing HTML so maybe regexes can work in this use case.

microalps:
Resolves #342

### DIFF
--- a/src/DotLiquid.Tests/StandardFilterTests.cs
+++ b/src/DotLiquid.Tests/StandardFilterTests.cs
@@ -142,7 +142,14 @@ namespace DotLiquid.Tests
         {
             Assert.AreEqual("test", StandardFilters.StripHtml("<div>test</div>"));
             Assert.AreEqual("test", StandardFilters.StripHtml("<div id='test'>test</div>"));
+            Assert.AreEqual("", StandardFilters.StripHtml("<script type='text/javascript'>document.write('some stuff');</script>"));
+            Assert.AreEqual("", StandardFilters.StripHtml("<style type='text/css'>foo bar</style>"));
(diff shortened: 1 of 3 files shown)


microalps:
Note: Ruby version of tests expect a empty string if Strip is passed a null. This was not the behavior in DotLiquid previously and remains unchanged to avoid breaking changes. I believe this is consistent with other DotLiquid filters.

daviburg (approved):

PR #458, opened 2021-09-28
Unix epoch string support

### LINKED ISSUE #457: UNIX Epoch timestamp in strings ignored by Date filter.
### Dotliquid version
2.2.508

### Expected behavior
UNIX Epoch timestamps in a string should be read and formatted by the Date filter, see liquid test case: [`assert_equal("07/05/2006", @filters.date("1152098955", "%m/%d/%Y"))`](https://github.com/Shopify/liquid/blob/db3999a008e5b5ac999328bef55baccdd18e5089/test/integration/standard_filter_test.rb#L530)

Additionally there is only support for timestamps in seconds, values in milliseconds are not supported currently by either Liquid or DotLiquid.

### Actual behavior
String are ignored.

robin-parker:
Adds support for UNIX Epoch timestamps as strings - resolves #457.

Also adds support for millisecond timestamps when a timestamp value lies outside the range supported by `DateTimeOffset.FromUnixTimeSeconds` (which is 01-Jan-0000 through 31-Dec-9999)

### DIFF
--- a/src/DotLiquid/StandardFilters.cs
+++ b/src/DotLiquid/StandardFilters.cs
@@ -659,7 +659,17 @@ namespace DotLiquid
                 }
                 else if (!DateTimeOffset.TryParse(value, out dateTimeOffset))
                 {
-                    return value;
+                    // As a final role of the dice, check if the string holds an integer that can be treated as a UNIX timestamp
+                    if (!long.TryParse(value, NumberStyles.Integer, CultureInfo.InvariantCulture, out var timestamp))
+                    {
+                        return value;
+                    }
+
+                    // Integers from -62_135_596_800 (01-Jan-0000) through +62_135_596_800 (31-Dec-3938) are treated as seconds,
+                    // anything outside this range is treated as milliseconds
+                    dateTimeOffset = (Math.Abs(timestamp) < 62_135_596_800 ?
+                        DateTimeOffset.FromUnixTimeSeconds(timestamp) :
+                        DateTimeOffset.FromUnixTimeMilliseconds(timestamp)).ToLocalTime();
                 }
             }
 
--- a/src/DotLiquid/StandardFilters.cs
+++ b/src/DotLiquid/StandardFilters.cs
@@ -637,25 +637,21 @@ namespace DotLiquid
                 return input.ToString();
 
             DateTimeOffset dateTimeOffset;
+            long timestamp = 0L;
             if (input is DateTimeOffset inputOffset)
             {
                 dateTimeOffset = inputOffset;
             }
-            else if ((input is decimal) || (input is double) || (input is float) || (input is int) || (input is uint) || (input is long) || (input is ulong) || (input is short) || (input is ushort))
+            else if ((input is decimal) || (input is double) || (input is float) || (input is int) || (input is uint) || (input is long) || (input is ulong) || (input is short) || (input is ushort)
+                || (input is string stringInput && long.TryParse(stringInput, NumberStyles.Integer, CultureInfo.InvariantCulture, out timestamp)))
             {
-#if CORE
-                dateTimeOffset = DateTimeOffset.FromUnixTimeSeconds(Convert.ToInt64(input)).ToLocalTime();
-#else
-                dateTimeOffset = new DateTimeOffset(1970, 1, 1, 0, 0, 0, TimeSpan.Zero).AddSeconds(Convert.ToDouble(input)).ToLocalTime();
-#endif
-            }
-            else if (input is string stringInput && long.TryParse(stringInput, NumberStyles.Integer, CultureInfo.InvariantCulture, out var unixTimestamp))
-            {
-                // If the timestamp is outside the range of the DateTimeOffset.FromUnixTimeSeconds() method, treat it as seconds
-                if (unixTimestamp < -62_135_596_800 || unixTimestamp > 253_402_300_799)
-                    dateTimeOffset = DateTimeOffset.FromUnixTimeMilliseconds(unixTimestamp).ToLocalTime();
-                else
-                    dateTimeOffset = DateTimeOffset.FromUnixTimeSeconds(unixTimestamp).ToLocalTime();
+                // If input is a string timestamp we already have a long value, if not convert
+                timestamp = input is string ? timestamp : Convert.ToInt64(input);
+
+                // If the timestamp is outside the range supported by the DateTimeOffset.FromUnixTimeSeconds() method, treat it as milliseconds
+                dateTimeOffset = (timestamp < -62_135_596_800 || timestamp > 253_402_300_799 ?
+                    DateTimeOffset.FromUnixTimeMilliseconds(timestamp) :
+                    DateTimeOffset.FromUnixTimeSeconds(timestamp)).ToLocalTime();
             }
             else
             {
--- a/src/DotLiquid/StandardFilters.cs
+++ b/src/DotLiquid/StandardFilters.cs
@@ -651,10 +651,11 @@ namespace DotLiquid
             }
             else if (input is string stringInput && long.TryParse(stringInput, NumberStyles.Integer, CultureInfo.InvariantCulture, out var unixTimestamp))
             {
-                // If the timestamp is outside the range of the DateTimeOffset.FromUnixTimeSeconds() method, treat it as milliseconds
-                if (unixTimestamp < 62_135_596_800 || unixTimestamp > 253_402_300_799)
+                // If the timestamp is outside the range of the DateTimeOffset.FromUnixTimeSeconds() method, treat it as seconds
+                if (unixTimestamp < -62_135_596_800 || unixTimestamp > 253_402_300_799)
                     dateTimeOffset = DateTimeOffset.FromUnixTimeMilliseconds(unixTimestamp).ToLocalTime();
                 else
+                    dateTimeOffset = DateTimeOffset.FromUnixTimeSeconds(unixTimestamp).ToLocalTime();
             }
             else
             {
--- a/src/DotLiquid/StandardFilters.cs
+++ b/src/DotLiquid/StandardFilters.cs
@@ -665,9 +665,9 @@ namespace DotLiquid
                         return value;
                     }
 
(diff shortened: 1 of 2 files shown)


src/DotLiquid/StandardFilters.cs
```diff
@@ -640,16 +640,27 @@ public static string Date(Context context, object input, string format)
...
 
             DateTimeOffset dateTimeOffset;
+            long timestamp = 0L;
             if (input is DateTimeOffset inputOffset)
             {
                 dateTimeOffset = inputOffset;
             }
-            else if ((input is decimal) || (input is double) || (input is float) || (input is int) || (input is uint) || (input is long) || (input is ulong) || (input is short) || (input is ushort))
+            else if ((input is decimal) || (input is double) || (input is float) || (input is int) || (input is uint) || (input is long) || (input is ulong) || (input is short) || (input is ushort)
+                || (input is string stringInput && long.TryParse(stringInput, NumberStyles.Integer, CultureInfo.InvariantCulture, out timestamp)))
```
microalps, line 649:
```suggestion
                || (input is string stringInput && long.TryParse(stringInput, NumberStyles.Integer, CultureInfo.InvariantCulture, out var timestamp)))

```

src/DotLiquid/StandardFilters.cs
```diff
@@ -640,16 +640,27 @@ public static string Date(Context context, object input, string format)
                 return input.ToString();
 
             DateTimeOffset dateTimeOffset;
+            long timestamp = 0L;
```
microalps, line 643:
```suggestion
```

src/DotLiquid/StandardFilters.cs
```diff
@@ -640,16 +640,27 @@ public static string Date(Context context, object input, string format)
...
+            else if ((input is decimal) || (input is double) || (input is float) || (input is int) || (input is uint) || (input is long) || (input is ulong) || (input is short) || (input is ushort)
+                || (input is string stringInput && long.TryParse(stringInput, NumberStyles.Integer, CultureInfo.InvariantCulture, out timestamp)))
             {
-#if CORE
-                dateTimeOffset = DateTimeOffset.FromUnixTimeSeconds(Convert.ToInt64(input)).ToLocalTime();
+                // If input is a string timestamp we already have a long value, if not convert
+                timestamp = input is string ? timestamp : Convert.ToInt64(input);
+
+                // Check if the timestamp is outside the range supported by the DateTimeOffset.FromUnixTimeSeconds() method
+                var isMilliseconds = timestamp < -62_135_596_800 || timestamp > 253_402_300_799;
```
microalps, line 655: So it's impossible to specify a date prior to 1/11/1978 in milliseconds. It's sort of arbitrary IMHO. Would approve without this idea.

robin-parker: It's not quite arbitrary, the numbers are taken from the documented range supported by [DateTimeOffset.FromUnixTimeSeconds](https://docs.microsoft.com/en-us/dotnet/api/system.datetimeoffset.fromunixtimeseconds?view=net-5.0#exceptions).

My use case is to parse a UTC epoch milliseconds timestamp and convert to a specified timezone. So, perhaps a new extended filter (such as `MillisToDate`) similar to `Date` with a target date-format parameter, plus a timezone parameter would be a cleaner solution.

microalps: Agreed on a separate filter. We can't add to the library of you want to specify a timezone as that would require reliance on an external library like NodaTime. I believe there was a library that did just that need to find it.

src/DotLiquid/StandardFilters.cs
robin-parker: I tried this originally, but it doesn't compile: `The name 'timestamp' does not exist in the current context [DotLiquid]csharp(CS0103)`
It seems the compiler is not aware that `out var timestamp` will be initialised when `input is string`.

src/DotLiquid/StandardFilters.cs
robin-parker: Alternate filter ConvertTime added

microalps:
Personally I think this needs to be broken down to three 'extended' filters:

1. unix_ms - accepts numeric or string with numeric value - converts milliseconds to datetimeoffset
2. unix_sec - accepts numeric or string with numeric value - converts seconds to datetimeoffset
3. zone - accepts datetime (marked utc) , datetime (unspecified or local), date string, or datetimeoffset and converts to datetimeoffset in the specified zone.
a. numerics can be processed with unix_ms | zone: '<zone>' or unix_sec | zone '<zone>' 
b. now and today can be processed with date | zone: '<zone>'
c. formatting to a human readable string would chain zone: '<zone>' | date: 'yyyy/MM/dd'

The test cases would be greatly limited and the functionality understandable. Happy to debate naming of these filters.

src/DotLiquid/ExtendedFilters.cs
```diff
@@ -38,5 +40,83 @@ public static string UpcaseFirst(Context context, string input)
...
+        /// </summary>
+        /// <param name="context">The DotLiquid context</param>
+        /// <param name="input">DateTimeOffset, DateTime (unspecified, local or UTC) or a date string</param>
+        /// <param name="convertToTimezoneId">Windows timezone ID to convert to</param>
+        /// <seealso href="https://docs.microsoft.com/en-us/dotnet/api/system.timezoneinfo.findsystemtimezonebyid?view=netcore-3.1#remarks"/>
+        public static object TimeZone(Context context, object input, string convertToTimezoneId)
+        {
+            // Convert the input to a DateTimeOffset
+            if (input is DateTimeOffset dateTimeOffset ||
+                (input is string stringInput && DateTimeOffset.TryParse(stringInput, out dateTimeOffset)))
```
microalps, line 75: I don't agree with parsing in here, created extra overhead/testing.

src/DotLiquid/ExtendedFilters.cs
```diff
@@ -38,5 +40,83 @@ public static string UpcaseFirst(Context context, string input)
...
+        }
+
+        /// <summary>
+        /// Create a DateTimeOffset from the provided UNIX Epoch timestamp in milliseconds.
+        /// </summary>
+        /// <return cref="DateTimeOffset">A Date in UTC</return>
+        private static DateTimeOffset CreateDateTimeOffsetFromUnixMilliseconds(long milliseconds)
+        {
+#if NET45
+            return new DateTimeOffset(1970, 1, 1, 0, 0, 0, TimeSpan.Zero).AddMilliseconds(milliseconds);
```
microalps, line 116: should this date be statically declared?

src/DotLiquid/StandardFilters.cs
microalps: I see, you use timestamp when input is not string by assignment.

src/DotLiquid.Tests/ExtendedFilterTests.cs
```diff
@@ -47,5 +44,109 @@ public void TestUpcaseFirst()
...
+
+        [Test]
+        public void TestTimeZone()
+        {
+            Context context = new Context(CultureInfo.CurrentCulture);
+
+            // Test UTC specified DateTime
+            var utcDateTime = new DateTime(2020, 2, 29, 12, 10, 11, DateTimeKind.Utc);
+            Assert.AreEqual("2020-02-29T12:10:11+00:00", ToIso8601DateString(ExtendedFilters.TimeZone(context: context, input: utcDateTime, convertToTimezoneId: "UTC")));
+            Assert.AreEqual("2020-02-29T12:10:11+00:00", ToIso8601DateString(ExtendedFilters.TimeZone(context: context, input: utcDateTime, convertToTimezoneId: "Local")));
```
microalps, line 80: How will this pass on developer machines?

microalps (commented):
Looks pretty good to me. Local testing is an issue though.

src/DotLiquid/ExtendedFilters.cs
robin-parker: The date filter returns a string, so without string parsing time_zone cannot be chained after a date filter, which limits it's usability.
How about if I changed it to use TryParseExact with a fixed format of say ISO-8601 with explicit timezone specifier.
That would reduce the overhead and simplify testing to a narrower set of input formats.

src/DotLiquid.Tests/ExtendedFilterTests.cs
robin-parker: I've removed Local, thinking again I can't see an obvious use case for it.

src/DotLiquid/ExtendedFilters.cs
microalps: Good idea. I'd accept it if it was limited in scope as you describe.

src/DotLiquid.Tests/ExtendedFilterTests.cs
```diff
@@ -74,42 +74,30 @@ public void TestTimeZone()
         {
             Context context = new Context(CultureInfo.CurrentCulture);
 
-            // Test UTC specified DateTime
```
microalps: These tests between the type should result in a different value. On a machine NOT in UTC, converting a datetime that is EXPLICITLY local to UTC would change the hour, but a datetime that is EXPLICITLY UTC would not. This functionality is missing I guess.

microalps (changes requested):

src/DotLiquid/ExtendedFilters.cs
robin-parker: Changes made and docs drafted at https://github.com/robin-parker/dotliquid/wiki/time_zone

src/DotLiquid.Tests/ExtendedFilterTests.cs
robin-parker: @microalps, I'm not clear what functionality is missing. Could you advise/write a test case for the scenario you're thinking of?

microalps: Say you have a hash that includes a date - from DateTime.Now - and the local machine is not in UTC. Calling TimeZone with this parameter and 'UTC' as the destination should change the time. HOWEVER, the same with DateTime.UtcNow should not as it is already in UTC.

microalps:
Closed until author is ready for rebase.

Issue #456, opened 2021-09-24
Templates do not time out until blocking operation complete


ceilingfish:
### Dotliquid version
2.2.548
### Expected behavior

The template should throw a TimeoutException once the timeout has expired.

### Actual behavior

The rendering engine waits for the last blocking operation to complete before throwing a TimeoutException, meaning we can be blocking the calling thread for a much longer time than expected.

### Steps to reproduce the Problem (you can add files)

```csharp
//Linqpad / Roslynpad script
void Main()
{
	var template = Template.Parse("{% for i in (1..5) %}{{i}}. {{delay}}{% endfor %}");
	var parameters = Hash.FromAnonymousObject(new
	{
		delay = new Delay()
	});

	var cancellation = new CancellationTokenSource(50);
	var context = new RenderParameters(CultureInfo.InvariantCulture)
	{
		Timeout = 100,
		LocalVariables =parameters,
		ErrorsOutputMode = ErrorsOutputMode.Display
	};
	template.Render(context).Dump();//Blocks for 10 seconds
	
}

private class Delay : ILiquidizable
{
	public Context Context { get; set; }

	public object ToLiquid()
	{
		Thread.Sleep(10000);
		return "delay";
	}
}
```

ceilingfish:
I was wondering if we could expose a `CancellationToken` on `IContextAware.Context`.

microalps:
> I was wondering if we could expose a `CancellationToken` on `IContextAware.Context`.

There already is Context.CheckTimeout();

> Steps to reproduce the Problem
Your example includes a non-cancelable method - .NET 5.0 [has obsoleted Thread.Abort](https://docs.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/5.0/thread-abort-obsolete) so a real example of something that SHOULD work would help. 

> Add public cancellation token to Context 
Your PR uses a drop which has access to the context, while your example above uses ILiquidizable which does not. Seems like scope of the issue is yet to be fully defined.

microalps:
So then we need to convert the timeout to a cancellation token of its own and expose the linked cancellation token. This way cancelling either via timeout or cancellation token will bubble down to the async call.

microalps:
Double checked and it is not possible to have both a timeout and a cancellation token. So no need for a linked token. Also, cancellation token is not available to RenderParameters. If we do convert every timeout to a cancellationToken in addition to exposing it like you did we should be OK.

PR #459, opened 2021-10-02
Add public cancellation token to Context

ceilingfish:
Address #456 by adding a publicly accessible cancellation token. This can be used by subclasses of `IContextAware`, or `Drop`

### DIFF
--- a/src/DotLiquid.Tests/DropTests.cs
+++ b/src/DotLiquid.Tests/DropTests.cs
@@ -375,18 +375,9 @@ namespace DotLiquid.Tests
         {
             using (var cancellationTokenSource = new CancellationTokenSource())
             {
-                var cancellationToken = cancellationTokenSource.Token;
-                var context = new Context(
-                    environments: new List<Hash>(),
-                    outerScope: Hash.FromAnonymousObject(new { context = new CancellationDrop() }),
-                    registers: new Hash(),
-                    errorsOutputMode: ErrorsOutputMode.Display,
-                    maxIterations: 0,
-                    formatProvider: CultureInfo.InvariantCulture,
-                    cancellationToken: cancellationToken);
                 var renderParameters = new RenderParameters(CultureInfo.InvariantCulture)
                 {
-                    Context = context
+                    CancellationToken = cancellationTokenSource.Token
                 };
                 cancellationTokenSource.Cancel();
                 Assert.Throws<OperationCanceledException>(() => Template.Parse("{{ context.halt }}").Render(renderParameters));
@@ -399,7 +390,8 @@ namespace DotLiquid.Tests
             var cancellationDrop = new CancellationDrop();
             var renderParameters = new RenderParameters(CultureInfo.InvariantCulture)
             {
-                LocalVariables = Hash.FromAnonymousObject(new { context = cancellationDrop, dummy = new[] { 1, 2, 3 } })
+                LocalVariables = Hash.FromAnonymousObject(new { context = cancellationDrop, dummy = new[] { 1, 2, 3 } }),
+                CancellationToken = new CancellationToken(canceled: false)
             };
             cancellationDrop.Signal.Set();
             Assert.AreEqual("HaltHaltHalt", Template.Parse("{% for a in dummy %}{{ context.halt }}{% endfor %}").Render(renderParameters));
--- a/src/DotLiquid.Tests/DropTests.cs
+++ b/src/DotLiquid.Tests/DropTests.cs
@@ -376,13 +376,14 @@ namespace DotLiquid.Tests
             using (var cancellationTokenSource = new CancellationTokenSource())
             {
                 var cancellationToken = cancellationTokenSource.Token;
-                var context = new Context(new List<Hash>(),
-                    Hash.FromAnonymousObject(new { context = new CancellationDrop() }),
-                    new Hash(),
-                    ErrorsOutputMode.Display,
-                    0,
-                    CultureInfo.InvariantCulture,
-                    cancellationToken);
+                var context = new Context(
+                    environments: new List<Hash>(),
+                    outerScope: Hash.FromAnonymousObject(new { context = new CancellationDrop() }),
+                    registers: new Hash(),
+                    errorsOutputMode: ErrorsOutputMode.Display,
+                    maxIterations: 0,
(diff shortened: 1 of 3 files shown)


src/DotLiquid.Tests/DropTests.cs
```diff
@@ -49,6 +52,20 @@ public override object BeforeMethod(string method)
...
 
+        internal class CancellationDrop : Drop
+        {
+            public ManualResetEventSlim Signal { get; } = new ManualResetEventSlim();
+
+            public string Halt
+            {
+                get
+                {
+                    Signal.Wait(Context.CancellationToken);
```
daviburg, line 63: I'm not familiar with Signal.Wait
Could you point me to the documentation or implementation of that?

src/DotLiquid.Tests/DropTests.cs
```diff
@@ -353,6 +370,33 @@ public void TestAccessContextFromDrop()
             Assert.AreEqual("123", Template.Parse("{% for a in dummy %}{{ context.loop_pos }}{% endfor %}").Render(Hash.FromAnonymousObject(new { context = new ContextDrop(), dummy = new[] { 1, 2, 3 } })));
         }
 
+        [Test]
+        public void TestCancelBlockingOperationDrop()
+        {
+            var cancellationToken = new CancellationToken(true);
```
daviburg, line 376: Use a cancellation token source to get a cancellation token. see for instance https://docs.microsoft.com/en-us/dotnet/api/system.threading.cancellationtoken?view=net-5.0#examples

daviburg (changes requested):

src/DotLiquid.Tests/DropTests.cs
ceilingfish: I've switched to using a `CancellationTokenSource` now.

microalps: Still using the Slim, and cancellation is before the render. I think an ideal way to test this function is to have a context variable with number of seconds to sleep and see the framework cancel if the sleep > timeout. The same template will pass/fail depending on the number of seconds to sleep in the drop.

microalps: By the way, Context.CheckTimeout() is public, checks the cancellation token, and can already be accessed from the drop. Not seeing what making cancellationToken public is adding - as that is NOT checking timeout. Example:
```c#
        internal class CancellationDrop : Drop
        {
            public string Halt
            {
                get
                {
                    var timeToWait = (int)Context["sleep_time"];
                    while (timeToWait > 0)
                    {
                        Context.CheckTimeout();
                        var sleepTime = Math.Min(timeToWait, 50);
                        System.Threading.Thread.Sleep(sleepTime);
                        timeToWait -= sleepTime;
                    }

                    return "Here";
                }
            }
        }
```
The only missing piece is NATIVE evaluation of a drop property with a timeout. That would allow for a getter:
```c#
                    System.Threading.Thread.Sleep((int)Context["sleep_time"]);
                    return "Here";
```

src/DotLiquid.Tests/DropTests.cs
```diff
@@ -353,6 +370,40 @@ public void TestAccessContextFromDrop()
             Assert.AreEqual("123", Template.Parse("{% for a in dummy %}{{ context.loop_pos }}{% endfor %}").Render(Hash.FromAnonymousObject(new { context = new ContextDrop(), dummy = new[] { 1, 2, 3 } })));
         }
 
+        [Test]
+        public void TestCancelBlockingOperationDrop()
+        {
+            using (var cancellationTokenSource = new CancellationTokenSource())
+            {
+                var cancellationToken = cancellationTokenSource.Token;
+                var context = new Context(new List<Hash>(),
```
daviburg, line 379: Nit, on coding style: consider also putting the first argument on its on line when using multiple lines. And consider using named parameters when 1. using multiple lines, or 2. using constant input like this 0, or 3. using multiple parameters of same type hence their order will be ambiguous to human reader without their name.

daviburg (approved):

microalps (changes requested):
Waiting for rewrite to handle both types of cancellation of context, timeout and CancellationTokens. We should possibly expose the modern CancelationToken to the RenderParameters class as well to complement timeout [which is marked obsolete in the context class].

src/DotLiquid/Context.cs
```diff
@@ -42,6 +42,11 @@ public int MaxIterations
 
         private Strainer _strainer;
 
+        /// <summary>
+        /// A cancellation token which indicates if the template render has been cancelled
+        /// </summary>
+        public CancellationToken CancellationToken => _cancellationToken;
```
microalps, line 48: Without implementing a CancellationTokenSource for timeout based cancellations, this is very limited in real use. Adding either or to the RenderParameters doesn't negate that. @daviburg  the suggestion was to change Context to inherit IDisposable so he can use my code snippet safely, but author said it would be a breaking change. What do you suggest?

Another alternative is dropping timeout from Context entirely (it is marked obsolete) and converting the timeout to a CancellationToken in RenderParameters.

src/DotLiquid/RenderParameters.cs
```diff
@@ -107,15 +113,15 @@ internal void Evaluate(Template template, out Context context, out Hash register
                 environments.Add(LocalVariables);
             if (template.IsThreadSafe)
             {
-                context = new Context(environments, new Hash(), new Hash(), ErrorsOutputMode, MaxIterations, Timeout, FormatProvider)
+                context = new Context(environments, new Hash(), new Hash(), ErrorsOutputMode, MaxIterations, FormatProvider, Timeout, CancellationToken)
```
microalps, line 116: Not clear how this is either or. True this constructor is internal, but here you exposed the possibility of having both a timeout and a cancellation token. Respecting both, as your commit message says, convolutes things even more.

src/DotLiquid.Tests/DropTests.cs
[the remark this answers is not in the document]
microalps: Please see my test in this commit for what I was expecting:
https://github.com/microalps/dotliquid/blob/cancellation-token-disposable/src/DotLiquid.Tests/DropTests.cs

microalps (changes requested):

PR #466, opened 2021-10-29
Add Errors collection to render parameters

ceilingfish:
This PR adds an `Errors` collection to the `RenderParameters` object. It contains the same errors as `Context.Errors`. This allows clients of `template.Render(renderParameters)` to get hold of errors in a thread safe way.

### DIFF
--- a/src/DotLiquid/Context.cs
+++ b/src/DotLiquid/Context.cs
@@ -81,10 +81,8 @@ namespace DotLiquid
              , int maxIterations
              , int timeout
              , IFormatProvider formatProvider)
-            : this(environments, outerScope, registers, errorsOutputMode, maxIterations, formatProvider, CancellationToken.None)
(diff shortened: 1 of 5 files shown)


src/DotLiquid/Context.cs
```diff
@@ -105,6 +103,28 @@ public Context
...
+        {
+        }
+
+        internal Context
+            (List<Hash> environments
+             , Hash outerScope
+             , Hash registers
+             , ErrorsOutputMode errorsOutputMode
+             , int maxIterations
+             , int? timeout
```
microalps, line 124: all constructors with timeout are marked obsolete. See our other PR discussion about this. We should not be adding new constructors that include obsolete parameters. Let's hold until the other PR is worked out.

microalps:

Issue #205, opened 2016-10-10
Increment/Decrement tag not existing


davidpelayo:
As specified on the oficial Shopify Liquid documentation, there are two tags to iterate over variables' values, which aren't existing in dotliquid, which are:
- Increment
- Decrement

[See this doc](https://help.shopify.com/themes/liquid/tags/variable-tags).

It would be nice to have them.

laedit:
Sadly we don't have time to add this ourselves, but we are welcoming any PRs :smile:

Issue #331, opened 2018-12-28
Capture doesn't support assigning into all valid variables


danielcrabtree:
### Dotliquid version
Latest master - commit 8340562

### Expected behavior
{% capture var-2 %}{{ var }} foo {% endcapture %}{{ var-2 }}
Should assign evaluated value of "{{ var }} foo " to "var-2".

### Actual behavior
"var-2" is not assigned anything.

danielcrabtree:
I have fixed this in PR #321 by making capture assignment work in the same way as assign. So capture now supports assignment all variables.

PR #467, opened 2021-10-29
Add Increment & Decrement tag

ceilingfish:
Adding Increment / Decrement tags as discussed in #205

### DIFF
--- a/src/DotLiquid.Tests/Tags/StandardTagTests.cs
+++ b/src/DotLiquid.Tests/Tags/StandardTagTests.cs
@@ -724,6 +724,12 @@ Maths 2: Eric Schmidt (ID3), Bruce Banner (ID4),
         }
 
         [Test]
+        public void IncrementDetectsBadSyntax()
+        {
+            Assert.Throws<SyntaxException>(() => Template.Parse("{% increment %}"));
+        }
+
+        [Test]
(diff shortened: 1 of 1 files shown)


microalps:
Need to check what your regex matches. We had another commit that used same (which incidentally assigns does as well) but this differs from capture. Problem is the variable can have a dot in this version but cannot have a hyphen in the capture. Both are wrong IMHO.

src/DotLiquid/Tags/Decrement.cs
```diff
@@ -0,0 +1,61 @@
...
+using DotLiquid.Util;
+
+namespace DotLiquid.Tags
+{
+    /// <summary>
+    /// Decrements a computed value
+    /// </summary>
+    public class Decrement : Tag
+    {
+        private static readonly Regex Syntax = R.B(R.Q(@"(?<Variable>{0}+)\s*"), Liquid.VariableSignature);
```
microalps, line 14: Variable signature includes dotted and bracketed notations. This should be a simple variable name imho. See capture block, but keep in mind to fix the issue reported in #331 .

microalps (changes requested):

microalps:
This was included in a different PR with changes.

PR #473, opened 2022-01-12
Add slice array support

### LINKED ISSUE #455: Slice does not currently have array support
### Dotliquid version
2.2.541

### Expected behavior
https://shopify.github.io/liquid/filters/slice/
When using slice on an array it returns the array elements specified:
In the case of the documentation: PaulGeorge

### Actual behavior
Error occurs:
Object of type 'System.String[]' cannot be converted to type 'System.String'

{% assign beatles = "John, Paul, George, Ringo" | split: ", " %}
{{ beatles | slice: 1, 2 }}

robin-parker:
- Add handling of string arrays to Slice filter, fixes #455
- Compatibility change: V22a introduced to fix DotLiquid vs. Liquid inconsitencies.


src/DotLiquid/StandardFilters.cs
```diff
@@ -65,13 +65,34 @@ public static int Size(object input)
...
 
+        /// <summary>
+        /// Returns a substring of one character or series of array items beginning at the index specified by the first argument.
+        /// </summary>
+        /// <param name="input">The input to be sliced</param>
+        /// <param name="start">zero-based start position of string or array, negative values count back from the end of the string/array.</param>
+        /// <param name="len">An optional second argument specifies the length of the substring or number of array items to be returned</param>
+        public static object Slice(object input, int start, int len = 1)
+        {
+            if (input is IEnumerable<string> enumerableInput)
```
microalps, line 76: not clear why this needs to be limited to ienumerable of type string. All ienumerable, except for strings (which are ienumerable<char>)  should be supported if the spec doesn't say otherwise.

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -195,6 +195,24 @@ public void TestSlice()
             Assert.AreEqual("efg", StandardFilters.Slice("abcdefg", 4, 30));
             Assert.AreEqual("a", StandardFilters.Slice("abc", -4, 2));
             Assert.AreEqual("", StandardFilters.Slice("abcdefg", -10, 1));
+
+            // Test for arrays
+            var testArray = new string[] { "a", "b", "c", "d" };
```
microalps, line 200: this should probably be moved to another method. don't forget to create a test method for other types of arrays

microalps (changes requested):

src/DotLiquid/StandardFilters.cs
microalps: the ruby code says: 
if input.is_a?(Array)
I'm not a Ruby programmer, but I'm sure this can be tested further to validate. From what I see, any array is supported.

robin-parker: I'm not a ruby programmer either, so thanks for checking.

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -195,6 +195,24 @@ public void TestSlice()
             Assert.AreEqual("efg", StandardFilters.Slice("abcdefg", 4, 30));
             Assert.AreEqual("a", StandardFilters.Slice("abc", -4, 2));
             Assert.AreEqual("", StandardFilters.Slice("abcdefg", -10, 1));
+
+            // Test for arrays
+            var testArray = new string[] { "a", "b", "c", "d" };
+            Assert.AreEqual("a", StandardFilters.Slice(testArray, 0));
+            Assert.AreEqual("bc", StandardFilters.Slice(testArray, 1, 2));
+            Assert.AreEqual("c", StandardFilters.Slice(testArray, -2, 1));
+            Assert.AreEqual("d", StandardFilters.Slice(testArray, -1, 1));
```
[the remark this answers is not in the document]
robin-parker: @microalps, I've replicated the Ruby tests, which highlighted a small difference between the implementations.
DotLiquid returns null for an empty slice of a string input, whereas Liquid returns and empty string.

I've opted to go with the Liquid approach and return an empty string as this feels like a bug.
If you think this should be changed back so that DotLiquid return a null let me know.

microalps:
@daviburg how would you like to handle this change? Apply to 2.2+ only? Introduce a new 2.2a compatibility flag?

daviburg:
@microalps @robin-parker I agree that long term it is generally better / "correct" to match Shopify's Liquid behavior. Because it is difficult to say if a DotLiquid consumer has built a dependency on the current behavior, the safe road is indeed to use the compatibility flag / versioning we have built for this purpose.

daviburg:
Otherwise code looks good, thanks for having unit tests including examples from Shopify. The existing test that was changed should be wrapped in this legacy version test pattern we have used.

src/DotLiquid.Tests/StandardFilterTests.cs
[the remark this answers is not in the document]
robin-parker: Reinstated DotLiquid logic by default and introduced Liquid fix behind compatibility flag 2.2a

microalps:
First, I reread the case when null vs "" is returned. It seems these are invalid entries not the normal course of business. If so, I retract my ask for a flag. 

Second, tests for both versions should be identical. In my opinion we should separate the valid tests from invalid and test both contexts.

So TestSlice has input of context and all the tests in v22a now (except the null).
TestSliceV22 calls above with context22
TestSliceV22a calls above with context22 and expects same outcomes
TestSliceWithInvalidOffsetV22 has the tests that return ""
TestSliceWithInvalidOffsetV22a has the same tests that return null

microalps (approved):

src/DotLiquid/StandardFilters.cs
```diff
@@ -66,12 +67,52 @@ public static int Size(object input)
...
+            return (context.SyntaxCompatibilityLevel >= SyntaxCompatibility.DotLiquid22a && input == null) ? string.Empty : input;
+        }
+
         /// Return a Part of a String
         /// </summary>
-        /// <param name="input">Input to be transformed by this filter</param>
-        /// <param name="start">start position of string</param>
-        /// <param name="len">optional length of slice to be returned</param>
-        public static string Slice(string input, long start, long len = 1)
+        /// <param name="input"></param>
```
daviburg, line 111: Why were the input parameters descriptions removed here?

daviburg (approved):

src/DotLiquid/StandardFilters.cs
robin-parker: I've reinstated the comments.

daviburg (approved):

PR #477, opened 2022-04-07
Handling Nulls in Currency Filter

AnushaKaraka:
This PR is to add a quick null check for input in the Currency filter and added an associated testcase. 

Without this check a null reference exception is thrown whenever the input is null(which is a quite common scenario if we pass in a property of a model) and stopping the render of the entire template.

PS: Tried to create an issue first but couldn't for some reason. hence opened the PR directly. Please let me know if I have to follow a different process for the same. Here is my issue notes:

1. Create a customer object with a property named Amount.
2. Instantiate the object with a null value for Amount
3. Use DotLiquid syntax with currency filter `{{Customer.Amount | currency:'fr-CA'}}`

or 

 Use DotLiquid syntax with currency filter `{{null | currency:'fr-CA'}}`

### DIFF
--- a/src/DotLiquid.Tests/StandardFilterTests.cs
+++ b/src/DotLiquid.Tests/StandardFilterTests.cs
@@ -715,6 +715,21 @@ PaulGeorge",
         }
 
         [Test]
+        public void TestNullInputCurrency()
+        {
+            // Set the thread culture and test for backward compatibility
+            using (CultureHelper.SetCulture("en-US"))
+            {
+                Helper.AssertTemplateResult(
+                    expected: string.Empty,
+                    template: "{{ null | currency: 'de-DE' }}");
+            }
+
+            // _contextV20 is initialized with InvariantCulture
+            Assert.AreEqual(null, StandardFilters.Currency(context: _contextV20, input: null, languageTag: "de-DE"));
+        }
+
+        [Test]
         public void TestMalformedCurrency()
         {
             // Set the thread culture and test for backward compatibility
--- a/src/DotLiquid.Tests/StandardFilterTests.cs
+++ b/src/DotLiquid.Tests/StandardFilterTests.cs
@@ -715,19 +715,23 @@ PaulGeorge",
         }
 
         [Test]
-        public void TestNullInputCurrency()
+        [TestCase(null)]
+        [TestCase("")]
+        public void TestNullOrEmptyInputCurrency(string input)
         {
             // Set the thread culture and test for backward compatibility
             using (CultureHelper.SetCulture("en-US"))
             {
                 Helper.AssertTemplateResult(
                     expected: string.Empty,
-                    template: "{{ null | currency: 'de-DE' }}");
+                    template: "{{ input | currency: 'de-DE' }}");
             }
 
             // _contextV20 is initialized with InvariantCulture
-            Assert.AreEqual(null, StandardFilters.Currency(context: _contextV20, input: null, languageTag: "de-DE"));
-        }
+            Assert.AreEqual(
+                expected: input,
+                actual: StandardFilters.Currency(context: _contextV20, input: input, languageTag: "de-DE"));
+        }        
 
         [Test]
         public void TestMalformedCurrency()
--- a/src/DotLiquid.Tests/StandardFilterTests.cs
+++ b/src/DotLiquid.Tests/StandardFilterTests.cs
@@ -724,7 +724,8 @@ PaulGeorge",
             {
                 Helper.AssertTemplateResult(
                     expected: string.Empty,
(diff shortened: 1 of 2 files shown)


src/DotLiquid/StandardFilters.cs
```diff
@@ -359,6 +359,9 @@ public static string Rstrip(string input)
         /// <seealso href="https://shopify.dev/api/liquid/filters/money-filters#money">Shopify Money filter</seealso>
         public static string Currency(Context context, object input, string languageTag = null)
         {
+            // Check for null input, return null
+            if (input == null) return null;
```
microalps, line 363:
```suggestion
            if (string.IsNullOrEmpty(input))
                return input;
```

microalps (commented):
Would like to see a test and handling for empty string once we are at it.

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -714,6 +714,21 @@ public void TestEuroCurrencyFromString(object input, string expected, string lan
...
+            // Set the thread culture and test for backward compatibility
+            using (CultureHelper.SetCulture("en-US"))
+            {
+                Helper.AssertTemplateResult(
+                    expected: string.Empty,
+                    template: "{{ null | currency: 'de-DE' }}");
+            }
+
+            // _contextV20 is initialized with InvariantCulture
+            Assert.AreEqual(null, StandardFilters.Currency(context: _contextV20, input: null, languageTag: "de-DE"));
```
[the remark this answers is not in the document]
daviburg: Shouldn't there be also a template test with the V2 context?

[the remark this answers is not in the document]
microalps: There is no distinction, so it doesn't matter. This is standard pattern for most.

src/DotLiquid/StandardFilters.cs
AnushaKaraka: `string.IsNullOrEmpty(input)` is not working unless I explicitly cast the input(which is an object) to a string. and that explicit cast is not working with other datatypes and causing some existing testcases to fail.

But I updated the test case to handle the empty string scenario as well. the empty string scenario was working as expected even before this change.

src/DotLiquid.Tests/StandardFilterTests.cs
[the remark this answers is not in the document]
AnushaKaraka: added named params to the Assert.AreEqual call in the testcase.

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -715,19 +715,23 @@ public void TestEuroCurrencyFromString(object input, string expected, string lan
...
+        [TestCase("")]
+        public void TestNullOrEmptyInputCurrency(string input)
         {
             // Set the thread culture and test for backward compatibility
             using (CultureHelper.SetCulture("en-US"))
             {
                 Helper.AssertTemplateResult(
                     expected: string.Empty,
-                    template: "{{ null | currency: 'de-DE' }}");
+                    template: "{{ input | currency: 'de-DE' }}");
```
microalps, line 727: Is there a missing line here defining the hash table with variables? Because this is testing for input variable in scope of Liquid - and not passed from this method.

src/DotLiquid/StandardFilters.cs
microalps: Thanks for that.

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -714,6 +714,25 @@ public void TestEuroCurrencyFromString(object input, string expected, string lan
...
+        [TestCase(null)]
+        [TestCase("")]
+        public void TestNullOrEmptyInputCurrency(string input)
+        {
+            // Set the thread culture and test for backward compatibility
+            using (CultureHelper.SetCulture("en-US"))
+            {
+                Helper.AssertTemplateResult(
+                    expected: string.Empty,
+                    template: "{{ input | currency: 'de-DE' }}");
```
microalps, line 727:
```suggestion
                    expected: input,
                    template: "{{ input | currency: 'de-DE' }}",
                    localVariables: Hash.FromAnonymousObject(new { input = input });
```

AnushaKaraka: Fixed the testcase. the expected is always string.Empty for both Null and Empty inputs. just changed that part in the above suggestion

microalps: Of course, a template render doesn't return null. My mistake!

microalps (approved):

PR #478, opened 2022-04-14
Local/Embedded File System Naming Requirements

### LINKED ISSUE #472: Include doesn't allow nested template name with length of 1 symbol
I want to include nested template via `{% include 'B' %}`, I have `_B.liquid` file near with my caller template.

**Error**: Liquid error: Error - Illegal template name 'B'

If I change `B` to `BB`, it starts to work.

microalps:
Resolves #472

### DIFF
--- a/src/DotLiquid.Tests/FileSystemTests.cs
+++ b/src/DotLiquid.Tests/FileSystemTests.cs
@@ -54,7 +54,9 @@ namespace DotLiquid.Tests
         {
             LocalFileSystem fileSystem = new LocalFileSystem(@"D:\Some\Path");
             foreach (var validPath in validPaths)
-                Assert.AreEqual(System.IO.Path.Combine(@"D:\Some\Path", validPath.Value.Replace('\\', Path.DirectorySeparatorChar)), fileSystem.FullPath(validPath.Key));
+                Assert.AreEqual(
+                    expected: Path.Combine(@"D:\Some\Path", validPath.Value.Replace('\\', Path.DirectorySeparatorChar)),
+                    actual: fileSystem.FullPath(validPath.Key));
 
             foreach (var invalidPath in invalidPaths)
                 Assert.Throws<FileSystemException>(() => fileSystem.FullPath(invalidPath));
@@ -76,7 +78,9 @@ namespace DotLiquid.Tests
             var assembly = typeof(FileSystemTests).GetTypeInfo().Assembly;
             EmbeddedFileSystem fileSystem = new EmbeddedFileSystem(assembly, "DotLiquid.Tests.Embedded");
             foreach (var validPath in validPaths)
-                Assert.AreEqual("DotLiquid.Tests.Embedded." + fileSystem.DirectorySeparators.Replace(validPath.Value, "."), fileSystem.FullPath(validPath.Key));
+                Assert.AreEqual(
+                    expected: "DotLiquid.Tests.Embedded." + fileSystem.DirectorySeparators.Replace(validPath.Value, "."),
+                    actual: fileSystem.FullPath(validPath.Key));
 
             foreach (var invalidPath in invalidPaths)
                 Assert.Throws<FileSystemException>(() => fileSystem.FullPath(invalidPath));
--- a/src/DotLiquid.Tests/FileSystemTests.cs
+++ b/src/DotLiquid.Tests/FileSystemTests.cs
@@ -79,7 +79,7 @@ namespace DotLiquid.Tests
             EmbeddedFileSystem fileSystem = new EmbeddedFileSystem(assembly, "DotLiquid.Tests.Embedded");
             foreach (var validPath in validPaths)
                 Assert.AreEqual(
-                    expected: "DotLiquid.Tests.Embedded." + fileSystem.DirectorySeparators.Replace(validPath.Value, "."),
+                    expected: "DotLiquid.Tests.Embedded." + Liquid.DirectorySeparatorsRegex.Replace(validPath.Value, "."),
                     actual: fileSystem.FullPath(validPath.Key));
 
             foreach (var invalidPath in invalidPaths)
--- a/src/DotLiquid.Tests/FileSystemTests.cs
+++ b/src/DotLiquid.Tests/FileSystemTests.cs
@@ -5,11 +5,42 @@ using NUnit.Framework;
 
 namespace DotLiquid.Tests
 {
+    using System.Collections.Generic;
+    using System.IO;
     using System.Reflection;
 
     [TestFixture]
     public class FileSystemTests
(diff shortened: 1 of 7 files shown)


src/DotLiquid.Tests/RegexpTests.cs
```diff
@@ -18,20 +18,11 @@ public void TestAllRegexesAreCompiled()
...
                     if (t.FieldType == typeof(Regex))
                     {
-                        if (t.IsStatic)
-                        {
-                            Assert.AreNotEqual(0, RegexOptions.Compiled & ((Regex) t.GetValue(null)).Options);
-                        }
-                        else
-                        {
-                            Assert.AreNotEqual(0, RegexOptions.Compiled & ((Regex)t.GetValue(parent)).Options);
-                        }
```
microalps, line 32: Instance variables need an instance to be created. parent is a Type not an instance of it, and without knowing the constructor parameters this is not possible. Killed non-working code entirely to allow this test to pass.

src/DotLiquid.Tests/FileSystemTests.cs
```diff
@@ -22,22 +53,20 @@ public void TestDefault()
         public void TestLocal()
         {
             LocalFileSystem fileSystem = new LocalFileSystem(@"D:\Some\Path");
-            Assert.AreEqual(@"D:\Some\Path\_mypartial.liquid", fileSystem.FullPath("mypartial"));
-            Assert.AreEqual(@"D:\Some\Path\dir\_mypartial.liquid", fileSystem.FullPath("dir/mypartial"));
+            foreach (var validPath in validPaths)
+                Assert.AreEqual(System.IO.Path.Combine(@"D:\Some\Path", validPath.Value.Replace('\\', Path.DirectorySeparatorChar)), fileSystem.FullPath(validPath.Key));
```
daviburg, line 57: System.IO is redundant, or do we need to disambiguate because we defined another Path class conflicting with the system?

daviburg: Also preferable to use named parameters with Assert.Equal to avoid non-obvious out of order parameters.

src/DotLiquid/FileSystems/EmbeddedFileSystem.cs
```diff
@@ -47,17 +51,17 @@ public string ReadTemplateFile(Context context, string templateName)
...
         {
-            if (templatePath == null || !Regex.IsMatch(templatePath, @"^[^.\/][a-zA-Z0-9_\/]+$"))
+            if (templatePath == null || !Regex.IsMatch(templatePath, @"^(?![\\\/\.])(?:[^<>:;,?""*|\x00-\x1F\/\\]+|[\/\\](?!\.))+(?<!\/)$"))
                 throw new FileSystemException(
                     Liquid.ResourceManager.GetString("LocalFileSystemIllegalTemplateNameException"), templatePath);
 
-            var basePath = templatePath.Contains("/")
-                ? Path.Combine(Root, Path.GetDirectoryName(templatePath))
+            var basePath = DirectorySeparators.IsMatch(templatePath)
+                ? DirectorySeparators.Replace(Path.Combine(Root, Path.GetDirectoryName(templatePath)), ".")
```
daviburg, line 59: nit: consider named parameters when 2+ arguments share a common type.

src/DotLiquid/FileSystems/LocalFileSystem.cs
```diff
@@ -19,13 +19,16 @@ namespace DotLiquid.FileSystems
     /// </summary>
     public class LocalFileSystem : IFileSystem
     {
+        internal readonly Regex DirectorySeparators = new Regex(@"[\\/]", RegexOptions.Compiled);
```
daviburg, line 22: could this be declared just once in a common file?

src/DotLiquid/FileSystems/LocalFileSystem.cs
```diff
@@ -37,14 +40,13 @@ public string ReadTemplateFile(Context context, string templateName)
 
         public string FullPath(string templatePath)
         {
-            if (templatePath == null || !Regex.IsMatch(templatePath, @"^[^.\/][a-zA-Z0-9_\/]+$"))
+            if (templatePath == null || !Regex.IsMatch(templatePath, @"^(?![\\\/\.])(?:[^<>:;,?""*|\x00-\x1F\/\\]+|[\/\\](?!\.))+(?<!\/)$"))
```
daviburg, line 43: code should be refactored to avoid copies.

daviburg (approved):

src/DotLiquid/FileSystems/LocalFileSystem.cs
microalps: if we make it static, yes, and put in DotLiquid/Liquid.cs but then anyone not using include (or using an alternate storage) would have it load for nothing.

src/DotLiquid/FileSystems/LocalFileSystem.cs
microalps: Generally, we have all regex strings declared in Liquid.cs, but probably better to address by declaring the regex somewhere common. The rest of the fullpath implementations differ due to path separators being different.

src/DotLiquid/FileSystems/LocalFileSystem.cs
daviburg: Could we initialize just in time with C# Lazy?

src/DotLiquid/Liquid.cs
```diff
@@ -30,6 +33,15 @@ public static class Liquid
...
         public static bool UseRubyDateFormat = false;
 
+        internal static readonly string DirectorySeparators = @"[\\/]";
+        internal static readonly string LimitRelativePath = @"^(?![\\\/\.])(?:[^<>:;,?""*|\x00-\x1F\/\\]+|[\/\\](?!\.))+(?<!\/)$"; /* Blocks hidden files in linux and directory traversal .. */
+        private static readonly Lazy<Regex> LazyDirectorySeparatorsRegex = new Lazy<Regex>(() => R.C(DirectorySeparators), LazyThreadSafetyMode.ExecutionAndPublication);
+        private static readonly Lazy<Regex> LazyLimitRelativePathRegex = new Lazy<Regex>(() => R.C(LimitRelativePath), LazyThreadSafetyMode.ExecutionAndPublication);
+
+        internal static Regex DirectorySeparatorsRegex => LazyDirectorySeparatorsRegex.Value;
+        internal static Regex LimitRelativePathRegex => LazyLimitRelativePathRegex.Value;
+
```
daviburg, line 43: nit nit: duplicate empty line

daviburg (approved):

PR #484, opened 2022-05-09
Enable default security scan with GitHub CodeQL

daviburg:
While looking at newly accessible project settings, I noticed that code scanning for security vulnerabilities was not enabled. There is a default scan available from GitHub called CodeQL, so giving it a try.


daviburg:
As there is no C# code change, I assume the code coverage decrease report is some incorrect analysis.

daviburg:
Merging after chatting with Microalps.

PR #488, opened 2022-05-09
Plus, minus and times crossing 32 to 64 bit boundary

### LINKED ISSUE #483: Math Operations Don't Fully Handle Big Numbers
### Dotliquid version
2.2.614

### Expected behavior
Math functions around the integer boundary should work, for example adding 1 to int.MaxValue

### Actual behavior
Liquid error: Arithmetic operation resulted in an overflow.

``` 
    public void TestPlusNumericBoundary()
    {
        Helper.AssertTemplateResult(
            expected: "2147483648",
            template: "{{ big | plus: 1 }}",
            localVariables: Hash.FromAnonymousObject(new { big = int.MaxValue }));

        Helper.AssertTemplateResult(
            expected: "79228162514264337593543950336",
            template: "{{ big | plus: 1 }}",
            localVariables: Hash.FromAnonymousObject(new { big = decimal.MaxValue }));
    }
```

daviburg:
Handle integer numbers beyond the range of 32 bit integer but still within 64 bit integer range.

Fixes #483

### DIFF
--- a/src/DotLiquid.Tests/StandardFilterTests.cs
+++ b/src/DotLiquid.Tests/StandardFilterTests.cs
@@ -1206,6 +1206,31 @@ PaulGeorge",
         }
 
         [Test]
+        public void TestPlusNumericBoundary()
+        {
+            TestPlusNumericBoundaryByContext(_contextV20);
+            TestPlusNumericBoundaryByContext(_contextV21);
+            Helper.AssertTemplateResult(expected: "2147483648", template: "{{ '2147483647' | plus: 1 }}", syntax: _contextV21.SyntaxCompatibilityLevel);
+        }
+
+        private void TestPlusNumericBoundaryByContext(Context context)
+        {
+            // NOTE(David Burg): Tests promotion from 32-bit to 64-bit integer.
+            Helper.AssertTemplateResult(
+                expected: "2147483648",
+                template: "{{ big | plus: 1 }}",
+                localVariables: Hash.FromAnonymousObject(new { big = int.MaxValue }),
+                syntax: context.SyntaxCompatibilityLevel);
+
+            // NOTE(David Burg): We don't have support for BigInteger as this time.
+            Helper.AssertTemplateResult(
+                expected: "Liquid error: Arithmetic operation resulted in an overflow.",
+                template: "{{ big | plus: 1 }}",
+                localVariables: Hash.FromAnonymousObject(new { big = long.MaxValue }),
+                syntax: context.SyntaxCompatibilityLevel);
+        }
+
+        [Test]
         public void TestMinus()
         {
             TestMinus(_contextV20);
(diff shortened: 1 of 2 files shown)


src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -1205,6 +1205,31 @@ public void TestPlusStringV21()
             TestPlus(context);
         }
 
+        [Test]
+        public void TestPlusNumericBoundary()
+        {
+            TestPlusNumericBoundaryByContext(_contextV20);
+            TestPlusNumericBoundaryByContext(_contextV21);
+            Helper.AssertTemplateResult(expected: "2147483648", template: "{{ '2147483647' | plus: 1 }}", syntax: _contextV21.SyntaxCompatibilityLevel);
```
microalps, line 1213: Should we comment that this tests a string which is supported in Syntax 2.1? Should we also test a string as a local variable?

src/DotLiquid/StandardFilters.cs
```diff
@@ -778,12 +792,19 @@ public static object Minus(Context context, object input, object operand)
...
-                : DoMathsOperation(context, input, operand, Expression.MultiplyChecked);
+                return input is string && (operand is int || operand is long)
+                    ? Enumerable.Repeat((string)input, Convert.ToInt32(operand))
+                    : DoMathsOperation(context, input, operand, Expression.MultiplyChecked);
+            }
+            catch (OverflowException) when (input is int)
+            {
+                return DoMathsOperation(context, Convert.ToInt64(input), operand, Expression.MultiplyChecked);
+            }
         }
```
microalps, line 808: what about divide by fraction? e.g. 2147483647 / 0.5 might result in an overflow.

src/DotLiquid/StandardFilters.cs
```diff
@@ -751,12 +751,19 @@ public static object Last(IEnumerable array)
...
+                    return DoMathsOperation(context, input, operand, Expression.AddChecked);
 
-            return input is string
-                ? string.Concat(input, operand)
-                : DoMathsOperation(context, input, operand, Expression.AddChecked);
+                return input is string
+                    ? string.Concat(input, operand)
+                    : DoMathsOperation(context, input, operand, Expression.AddChecked);
+            }
+            catch (OverflowException) when (input is int)
```
microalps, line 763: wouldn't uint also have a boundary? and smaller integer values, like short? Additionally, Ruby implementation supports adding 1 to decimal max without losing precision.

microalps (changes requested):

src/DotLiquid/StandardFilters.cs
daviburg: I tried bytes and short, they worked fine without modification, I can add a few tests to demonstrate it. I didn't try uint.

daviburg: BigInteger is out of scope for a bug fix, that support would be a feature in itself.

src/DotLiquid/StandardFilters.cs
daviburg: Should convert to a floating point hence range is different.

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -1205,6 +1205,31 @@ public void TestPlusStringV21()
...
+        {
+            // NOTE(David Burg): Tests promotion from 32-bit to 64-bit integer.
+            Helper.AssertTemplateResult(
+                expected: "2147483648",
+                template: "{{ big | plus: 1 }}",
+                localVariables: Hash.FromAnonymousObject(new { big = int.MaxValue }),
+                syntax: context.SyntaxCompatibilityLevel);
+
+            // NOTE(David Burg): We don't have support for BigInteger as this time.
+            Helper.AssertTemplateResult(
```
daviburg, line 1226: todo: also add tests demonstrating that byte max, short max work already, but double range limit cannot be extended beyond current overflow.

RodneyRichardson:
is this superceded by #571? Can this be closed?

PR #491, opened 2022-05-11
LiquidTypeAttribute Support for Type Reserved Keyword

### LINKED ISSUE #479: Liquid error: An item with the same key has already been added. Key: get_type
### Dotliquid version
2.2.614

### Expected behavior

Code looks like this:
```csharp
var template = Template.Parse(CLASS_TEMPLETE_FULL);
var hash = Hash.FromAnonymousObject(new { data = tplData });
var content = template.Render(hash);
```

Hash is correct
image

Data structure:
```csharp
[LiquidType("*")]
public class TemplateField
{
    public string type { get; set; }
    public string name { get; set; }
    public string desc { get; set; }
    public bool isArray { get; set; }
}

[LiquidType("*")]
public class TemplateData
{
    public string excelFile { get; set; }
    public string tableName { get; set; }
    public string tableDesc { get; set; }
    public string entityName { get; set; }
    
    public List<string> namespaces { get; set; }
    public List<TemplateField> fields { get; set; }
}
```

Template:
```liquid
// Code generated by Excel2Unity. DO NOT EDIT.
// source file: {{data.excel_file}}
// source sheet: {{data.table_name}}

{% for ns in data.namespaces -%}
using {{ns}};
{% endfor -%}

namespace Config
{
        /// <summary>
	/// {{data.entity_name}}
        ///
	/// {{data.table_desc}}
	/// </summary>
	[Serializable]
	public partial struct {{data.entity_name}}
	{
		{% for field in data.fields -%}
		/// <summary>
		/// {{field.desc}}
		/// </summary>
		public {{field.type}} {{field.name}};

		{% endfor -%}
        }
}
```

### Actual behavior

But output error:
```csharp
System.ArgumentException: An item with the same key has already been added. Key: get_type
   at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at System.Collections.Generic.Dictionary`2.Add(TKey key, TValue value)
   at System.Linq.Enumerable.ToDictionary[TSource,TKey](IEnumerable`1 source, Func`2 keySelector, IEqualityComparer`1 comparer)
   at DotLiquid.TypeResolution.GetMemberDictionary[T](IEnumerable`1 members, Func`2 filterMemberCallback)
   at DotLiquid.TypeResolution..ctor(Type type, Func`2 filterMemberCallback)
   at DotLiquid.DropProxy.CreateTypeResolution(Type type)
   at DotLiquid.DropBase.get_TypeResolution()
   at DotLiquid.DropBase.InvokeDrop(Object name)
   at DotLiquid.DropBase.get_Item(Object method)
   at DotLiquid.Context.TryEvaluateHashOrArrayLikeObject(Object obj, Object key, Object& value)
   at DotLiquid.Context.TryGetVariable(IEnumerator`1 partsEnumerator, Object& variable)
   at DotLiquid.Context.Variable(String markup, Boolean notifyNotFound)
   at DotLiquid.Context.Resolve(String key, Boolean notifyNotFound)
   at DotLiquid.Context.get_Item(String key, Boolean notifyNotFound)
   at DotLiquid.Variable.RenderInternal(Context context)
   at DotLiquid.Variable.Render(Context context, TextWriter result)
   at DotLiquid.Block.RenderAll(List`1 list, Context context, TextWriter result)
```

microalps:
Resolves #479

### DIFF
--- a/src/DotLiquid.Tests/DropTests.cs
+++ b/src/DotLiquid.Tests/DropTests.cs
@@ -419,9 +419,9 @@ namespace DotLiquid.Tests
             {
                 var type = typeof(ConflictingChildDrop);
                 var resolver = new TypeResolution(type, mi => true);
-                Assert.IsTrue(resolver.CachedMethods.ContainsKey("get_class_name"));
+                CollectionAssert.Contains(resolver.CachedMethods.Keys, "get_class_name");
                 Assert.IsTrue(resolver.CachedMethods["get_class_name"].DeclaringType == type);
-                Assert.IsTrue(resolver.CachedProperties.ContainsKey("name"));
+                CollectionAssert.Contains(resolver.CachedProperties.Keys, "name");
                 Assert.IsTrue(resolver.CachedProperties["name"].DeclaringType == type);
 
                 Helper.AssertTemplateResult(
--- a/src/DotLiquid.Tests/DropTests.cs
+++ b/src/DotLiquid.Tests/DropTests.cs
@@ -116,6 +116,20 @@ namespace DotLiquid.Tests
             }
         }
 
+        internal class ConflictingParentDrop : Drop
+        {
+            public string Name { get; set; } = nameof(ConflictingParentDrop);
+
+            public string GetClassName() => nameof(ConflictingParentDrop);
+        }
+
+        internal class ConflictingChildDrop : ConflictingParentDrop
+        {
+            public new string Name { get; set; } = nameof(ConflictingChildDrop);
+
+            public new string GetClassName() => nameof(ConflictingChildDrop);
+        }
+
         internal class EnumerableDrop : Drop, IEnumerable
         {
             public int Size
@@ -396,6 +410,25 @@ namespace DotLiquid.Tests
                 template: "{{ value.ProductID }}",
                 anonymousObject: new { value = new CamelCaseDrop() },
                 namingConvention: new RubyNamingConvention());
+        }
+
+        [Test]
+        public void TestTypeResolutionDuplicateNames()
+        {
+            Helper.LockTemplateStaticVars(new RubyNamingConvention(), () =>
+            {
+                var type = typeof(ConflictingChildDrop);
+                var resolver = new TypeResolution(type, mi => true);
+                Assert.IsTrue(resolver.CachedMethods.ContainsKey("get_class_name"));
+                Assert.IsTrue(resolver.CachedMethods["get_class_name"].DeclaringType == type);
+                Assert.IsTrue(resolver.CachedProperties.ContainsKey("name"));
+                Assert.IsTrue(resolver.CachedProperties["name"].DeclaringType == type);
+
+                Helper.AssertTemplateResult(
+                    expected: "ConflictingChildDrop|ConflictingChildDrop",
+                    template: "{{ value.name }}|{{ value.get_class_name }}",
+                    localVariables: Hash.FromAnonymousObject(new { value = new ConflictingChildDrop() }));
+            });
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/DotLiquid/Drop.cs
```diff
@@ -70,13 +70,13 @@ private static IEnumerable<MethodInfo> GetMethodsWithoutDuplicateNames(Type type
             IList<MemberInfo> methods = predicate != null
                                             ? type
                                                   .GetRuntimeMethods()
-                                                  .Where(m => m.IsPublic && !m.IsStatic)
+                                                  .Where(m => m.IsPublic && !m.IsStatic && m.DeclaringType != typeof(object))
```
daviburg, line 73: I assume that code coverage feels that only half the conditions are covered in the unit tests. Is there some negative tests that can enable the other half of the conditions' coverage?

daviburg (approved):

src/DotLiquid/Drop.cs
microalps: predicate is never null. That's why I think codecoverage is not liking it. Will try to push something today to get coverage.

src/DotLiquid.Tests/LiquidTypeAttributeTests.cs
```diff
@@ -43,6 +43,14 @@ public class MyLiquidTypeWithReservedKeyword
             public string Type { get; set; }
         }
 
+        [LiquidType("*")]
+        public class MyLiquidTypeWithConflictingGetter
+        {
+            public string Name { get; set; }
+
+            public string GetName() => Name;
```
daviburg, line 51: Isn't it that by having the GetName explicit method return the same Name value, we can't distinguish in the test if the response came from that method or from the generated Name property getter?

src/DotLiquid.Tests/DropTests.cs
```diff
@@ -397,5 +411,24 @@ public void TestRubyNamingConventionPrintsHelpfulErrorIfMissingPropertyWouldMatc
...
         }
+
+        [Test]
+        public void TestTypeResolutionDuplicateNames()
+        {
+            Helper.LockTemplateStaticVars(new RubyNamingConvention(), () =>
+            {
+                var type = typeof(ConflictingChildDrop);
+                var resolver = new TypeResolution(type, mi => true);
+                Assert.IsTrue(resolver.CachedMethods.ContainsKey("get_class_name"));
```
daviburg, line 422: Maybe CollectionAssert.Contains on resolver.CachedMethods.Keys. Should give detailed error of the collection not containing the expected key when test fails.

https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.testtools.unittesting.collectionassert.contains?view=visualstudiosdk-2022

src/DotLiquid.Tests/LiquidTypeAttributeTests.cs
microalps: True, I can adjust that to be unique. But previously it clashed and threw an error so this is already progress.

src/DotLiquid.Tests/DropTests.cs
daviburg: I'm not fully understanding your response to the suggestion. I thought I was just suggesting a refactoring of the assert implementation so it validates the same condition but outputs a more detailed message if the condition fails. Did my suggestion change the asserted condition?

microalps: CollectionAssert.IsSubsetOf would work, but CollectionAssert.Contains returns this:

    Expected: some item equal to < "get_class_name" >
      But was:  < "get_class_name", "to_liquid", "to_string", "get_hash_code", "get_type" >

daviburg: It looks like we have some kind of format massaging missing, as we do see from the contains output that the collection does contain the item as far as we humans can tell.

microalps: My am I tired. My mistake, it's resolved now.

Issue #480, opened 2022-04-23
Documentation: Render with filters


pekspro:
On this page: https://github.com/dotliquid/dotliquid/wiki/DotLiquid-for-Developers it says that filters could be passed to the `Render` method. Is this correct? I do not find any overload like that.

microalps:
See https://github.com/dotliquid/dotliquid/blob/master/src/DotLiquid.Tests/Helper.cs#L61 - it is passed as a renderparameter object. Thanks for the catch!

pekspro:
Thank @microalps. I have updated the Wiki :-)

microalps:
@pekspro can you commit changes to the DotLiquid website "learn" page as well? https://github.com/dotliquid/dotliquid/tree/master/src/DotLiquid.Website/Views/Docs

pekspro:
@microalps, sure! I will give it a try in a few days

PR #492, opened 2022-05-16
Update Extending.cshtml

pekspro:
Updates sample code.

This is for issue #480


microalps:
@pekspro any reason you changed the example to not use Textilize filter? Does using RegexReplace as example change anything?

microalps:
This was incorporated into PR #554

PR #504, opened 2022-10-24
Added xml escaping as a function with replace function

Joostluijben:
We found out the dotliquid does not support the xml escaping yet like in [ruby ]([url](https://www.rubydoc.info/gems/spreefinery_themes/Liquid%2FFilters:xml_escape)). Therefore I implemented the function with only replace functions like shown [here](https://weblogs.sqlteam.com/mladenp/2008/10/21/different-ways-how-to-escape-an-xml-string-in-c/). I used the replace function because of the discussion found on this stackoverflow [post](https://stackoverflow.com/questions/1132494/string-escape-into-xml)

### DIFF
--- a/src/DotLiquid/RubyFilters.cs
+++ b/src/DotLiquid/RubyFilters.cs
@@ -1,0 +1,19 @@
+﻿namespace DotLiquid
+{
+    public class RubyFilters
+    {
+        public static string Xml_escape(string input)
+        {
+            if (string.IsNullOrEmpty(input))
+                return input;
+            try
+            {
+                return input.Replace("&", "&amp;").Replace("<", "&lt;").Replace(">", "&gt;").Replace("&#34;", "&quot;").Replace("'", "&apos;");
+            }
+            catch
+            {
+                return input;
+            }
+        }
+    }
+}


microalps:
1. Please include tests to validate this new functionality. If this is in Ruby Liquid, needs to have a c# of all tests included from Ruby.
2. Makes sense to place in ExtendedFilters or ShopifyFilters and not in it's own RubyFilters

microalps:
@Joostluijben We appreciate contributions, but please see my comments above. Also, codecov indicates that testing decreased, so please add whatever tests are needed to validate the functionality works as expected.

microalps (changes requested):
Kindly see above.

microalps:
This PR is being abandoned for the following reasons:
* This library only intends to replicate the functionality of Ruby Liquid and not Liquid for Shopify Templates 
* I could not find documentation of xml_escape as a feature or Liquid or Shopify's
* xml_escape is not in source or the version history of [Liquid](https://github.com/shopify/liquid) and actually doesn't render when tested on replit
* Author has not responded in more than a year

PR #508, opened 2022-12-07
Move Legacy Regex-based Replace to an extended filter

### LINKED ISSUE #507: Enhancement: Ability to utilize the Legacy Regex-based Replace filters
### Dotliquid version
DotLiquid22

### Expected behavior
Support Regex-based input strings to Replace filter without switching to DotLiquid20

microalps:
Resolves #507

### DIFF
--- a/src/DotLiquid/StandardFilters.cs
+++ b/src/DotLiquid/StandardFilters.cs
@@ -539,7 +539,7 @@ namespace DotLiquid
             if (context.SyntaxCompatibilityLevel >= SyntaxCompatibility.DotLiquid21)
                 return input.Replace(@string, replacement);
 
-            return ExtendedFilters.RegexReplace(input, @string, replacement);
+            return ExtendedFilters.RegexReplace(input: input, pattern: @string, replacement: replacement);
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


microalps:
Note to reviewer: Reason ReplaceFirst didn't get the same love is because replacement string cannot contain capture groups (and would be a breaking change to the old behavior if it did). I'm open to suggestions though on how this might work.

src/DotLiquid/StandardFilters.cs
```diff
@@ -539,7 +539,7 @@ public static string Replace(Context context, string input, string @string, stri
             if (context.SyntaxCompatibilityLevel >= SyntaxCompatibility.DotLiquid21)
                 return input.Replace(@string, replacement);
 
-            return Regex.Replace(input, @string, replacement, RegexOptions.None, Template.RegexTimeOut);
+            return ExtendedFilters.RegexReplace(input, @string, replacement);
```
daviburg, line 542: nit: because input string and replacement are all of the same type (string), the arguments could be out of order (they aren't here). Hence it could reduce error risk by using named parameters, here and in the actual Regex Replace method call.

daviburg (approved):

Issue #493, opened 2022-05-31
Expando Objects and mapped categories


AnEvilPenguin:
### Dotliquid version
2.2.656

### Expected behavior
All mapped categories are displayed

### Actual behavior
No categories are returned

### Steps to reproduce the Problem (you can add files)
This shows the actual (first) and expected behavior
https://dotnetfiddle.net/aiA1JM

But loosely I'm reading in JSON using Newtonsoft.JSON to an ExpandoObject and then attempting to use FromDictionary.

If my understanding of the problem is correct I believe that the FromDictionary method of Hash is only going down the one level and so is probably getting tripped up on the deeper ExpandoObjects.

I've looked over previous issues and from the ones that I found, ExpandoObjects aren't fully supported.

I have managed to resolve the issue myself using:

```

public static Hash FromNestedDictionary (IDictionary<string, object> dictionary)
{
  var hash = new Hash();

  foreach (string key in dictionary.Keys)
  {
	  IDictionary<string,object> dictionaryValue = dictionary[key] as IDictionary<string, object>;
	  if (dictionaryValue != null)
	  {
		  hash[key] = FromNestedDictionary(dictionaryValue);
		  continue;
	  }

	  List<object> listValue = dictionary[key] as List<object>;
	  if (listValue != null)
	  {
		  hash[key] = FromListOfObjects(listValue);
		  continue;
	  }
		  
	  hash[key] = dictionary[key];
  }

  return hash;
}

private static List<object> FromListOfObjects(List<object> list)
{
  List<object> output = new List<object>();

  foreach (var item in list)
  {
	  IDictionary<string, object> dictionary = item as IDictionary<string, object>;

	  if (dictionary != null)
	  {
		  output.Add(FromNestedDictionary(dictionary));
		  continue;
	  }

	  output.Add(item);
  }

  return output;
}

```

But I'm a novice programmer, so my initial questions would be: 
* Am I missing something that I could do today to get around this issue on my side?
* Do you know of anything daft that I've done with the workaround that I've not considered (I appreciate that you're not a code review service so no worries if not).
* If there is value in this, would it be something you would consider a PR for?

microalps:
Please provide a code snippet - object and template - to reproduce the issue without your fix.

microalps:
[...]

Issue #495, opened 2022-06-01
where filter for ExpandoObjects


AnEvilPenguin:
### Dotliquid version
2.2.656

### Expected behavior

```

where creates an array including only the objects with a given property value, or any truthy value by default.

In this example, assume you have a list of products and you want to show your kitchen products separately. Using where, you can create an array containing only the products that have a "type" of "kitchen":

All products:

- Vacuum

- Spatula

- Television

- Garlic press

- Coffee mug

- Limited edition sneakers

- Boring sneakers

- Hawaiian print sweater vest




Kitchen products:

- Spatula

- Garlic press


Say instead you have a list of products and you only want to show those that are available to buy. You can where with a property name but no target value to include all products with a truthy "available" value:

All products:

- Vacuum

- Spatula

- Television

- Garlic press

- Coffee mug

- Limited edition sneakers

- Boring sneakers

- Hawaiian print sweater vest




Available products:

- Coffee mug

- Boring sneakers



The where filter can also be used to find a single object in an array when combined with the first filter. For example, say you want to show off the shirt in your new fall collection:



Featured product: Hawaiian print sweater vest

```

### Actual behavior

```
where creates an array including only the objects with a given property value, or any truthy value by default.

In this example, assume you have a list of products and you want to show your kitchen products separately. Using where, you can create an array containing only the products that have a "type" of "kitchen":

All products:

- Vacuum

- Spatula

- Television

- Garlic press

- Coffee mug

- Limited edition sneakers

- Boring sneakers

- Hawaiian print sweater vest




Kitchen products:


Say instead you have a list of products and you only want to show those that are available to buy. You can where with a property name but no target value to include all products with a truthy "available" value:

All products:

- Vacuum

- Spatula

- Television

- Garlic press

- Coffee mug

- Limited edition sneakers

- Boring sneakers

- Hawaiian print sweater vest




Available products:



The where filter can also be used to find a single object in an array when combined with the first filter. For example, say you want to show off the shirt in your new fall collection:



Featured product: 
```

### Steps to reproduce the Problem (you can add files)

```

var template = Template.Parse(@"where creates an array including only the objects with a given property value, or any truthy value by default.

In this example, assume you have a list of products and you want to show your kitchen products separately. Using where, you can create an array containing only the products that have a ""type"" of ""kitchen"":

All products:
{% for product in products %}
- {{ product.title }}
{% endfor %}

{% assign kitchen_products = products | where: ""type"", ""kitchen"" %}

[...]

Issue #494, opened 2022-06-01
Sort filter for ExpandoObjects


AnEvilPenguin:
### Dotliquid version

2.2.656

### Expected behavior

```

An optional argument specifies which property of the array’s items to use for sorting:

  <h4>juice</ h4 >
  <h4>ham</ h4 >
  <h4>bacon</ h4 >
  <h4>beans</ h4 >
  <h4>microwave</ h4 >

```

### Actual behavior

```

An optional argument specifies which property of the array’s items to use for sorting:

  <h4></ h4 >
  <h4>beans</ h4 >
  <h4></ h4 >
  <h4>ham</ h4 >
  <h4></ h4 >
  <h4>bacon</ h4 >
  <h4></ h4 >
  <h4>microwave</ h4 >
  <h4></ h4 >
  <h4>juice</ h4 >

```

### Steps to reproduce the Problem (you can add files)

```

var template = Template.Parse(@"sort sorts items in an array in case-sensitive order:

{% assign my_array = ""zebra, octopus, giraffe, Sally Snake"" | split: "", "" %}

{{ my_array | sort | join: "", "" }}

An optional argument specifies which property of the array’s items to use for sorting:

{% assign products_by_price = collection.products | sort: ""price"" -%}
{% for product in products_by_price -%}
  <h4>{{ product.title }}</ h4 >
{% endfor %}");

var json = @"{
	""collection"": {
		""products"": [
			{
				""price"": 5.00,
				""title"": ""beans""
			},
			{
				""price"": 2.50,
				""title"": ""ham""
			},
			{
				""price"": 4.50,
				""title"": ""bacon""
			},
			{
				""price"": 160.00,
				""title"": ""microwave""
			},
			{
				""price"": 2.00,
				""title"": ""juice""
			}
		]
	}
}";

var parsedJson = JsonConvert.DeserializeObject<ExpandoObject>(json, new ExpandoObjectConverter());
var hashFromDictionary = Hash.FromDictionary(parsedJson);
Console.Write(template.Render(hashFromDictionary));

```

https://dotnetfiddle.net/yVLJsL

Loosely recreating the docs from https://shopify.github.io/liquid/filters/sort/

PR #509, opened 2022-12-11
where, map, sort filter should respect IIndexable and ExpandoObject

hieuvuvan:


microalps:
Thank you for your multiple contributions. I will review this week hopefully. On this PR does it address #495? Do we need to implement any of the other conditions present in Context.TryEvaluateHashOrArrayLikeObject?

hieuvuvan:
i don't think it address #495 - it about ExpandoObjects support. For other conditions check, should i use current checking behavior of  `StandardFilters#Map` for that? somthing like this:

```patch
-            return listedInput
-                .Select(selector: element =>
-                {
-                    if (element == null)
-                        return null;
-
-                    var indexable = element as IIndexable;
-                    if (indexable == null)
-                    {
-                        var type = element.GetType();
-                        var safeTypeTransformer = Template.GetSafeTypeTransformer(type);
-                        if (safeTypeTransformer != null)
-                            indexable = safeTypeTransformer(element) as DropBase;
-                        else
-                        {
-                            var liquidTypeAttribute = type
-                                .GetTypeInfo()
-                                .GetCustomAttributes(attributeType: typeof(LiquidTypeAttribute), inherit: false)
-                                .FirstOrDefault() as LiquidTypeAttribute;
-                            if (liquidTypeAttribute != null)
-                            {
-                                indexable = new DropProxy(element, liquidTypeAttribute.AllowedMembers);
-                            }
-                            else if (TypeUtility.IsAnonymousType(type))
-                            {
-                                return element.RespondTo(property) ? element.Send(property) : element;
-                            }
-                        }
-                    }
-
-                    return (indexable?.ContainsKey(property) ?? false) ? indexable[property] : null;
-                });
+            return listedInput.Select(element => ResolveObjectPropertyValue(element, property));
         }
 
         /// <summary>
@@ -1048,26 +1017,45 @@ public static IEnumerable Where(IEnumerable input, string propertyName, object t
         /// <param name="targetValue">target property value</param>
         private static bool HasMatchingProperty(this object any, string propertyName, object targetValue)
         {
-            // Check if the 'any' object has a propertyName
-            object propertyValue = null;
-            if (any is IDictionary dictionary && dictionary.Contains(key: propertyName))
-            {
-                propertyValue = dictionary[propertyName];
-            }
-            else if (any is IIndexable indexer && indexer.ContainsKey(propertyName))
-            {
-                propertyValue = indexer[propertyName];
-            }
-            else if (any != null && any.RespondTo(propertyName))
-            {
-                propertyValue = any.Send(propertyName);
-            }
-
+            var propertyValue = ResolveObjectPropertyValue(any, propertyName);
             return targetValue == null || propertyValue == null
                 ? propertyValue.IsTruthy()
                 : propertyValue.SafeTypeInsensitiveEqual(targetValue);
         }
 
+        private static object ResolveObjectPropertyValue(this object obj, string propertyName)
+        {
+            if (obj == null)
+                return null;
+            if (obj is IDictionary dictionary && dictionary.Contains(key: propertyName))
+                return dictionary[propertyName];
+            var indexable = obj as IIndexable;
+            if (indexable == null)
+            {
+                var type = obj.GetType();
+                var safeTypeTransformer = Template.GetSafeTypeTransformer(type);
+                if (safeTypeTransformer != null)
+                    indexable = safeTypeTransformer(obj) as DropBase;
+                else
+                {
+                    var liquidTypeAttribute = type
+                        .GetTypeInfo()
+                        .GetCustomAttributes(attributeType: typeof(LiquidTypeAttribute), inherit: false)
+                        .FirstOrDefault() as LiquidTypeAttribute;
+                    if (liquidTypeAttribute != null)
+                    {
+                        indexable = new DropProxy(obj, liquidTypeAttribute.AllowedMembers);
+                    }
+                    else if (TypeUtility.IsAnonymousType(type) && obj.RespondTo(propertyName))
+                    {
+                        return obj.Send(propertyName);
+                    }
+                }
+            }
+
+            return (indexable?.ContainsKey(propertyName) ?? false) ? indexable[propertyName] : null;
+        }
```

microalps:
In theory, but we might need to add more for ExpandoObject. See Context.TryEvaluateHashOrArrayLikeObject where we have:
```c#
            // Resolve #350/#417, add support for rendering of a nested  ExpandoObject
            else if (obj is IDictionary<string, object> dictionaryObject && dictionaryObject.ContainsKey(key.ToString()))
                value = dictionaryObject[key.ToString()];
```

hieuvuvan:
@microalps i update, it resolve #493 #494 #495

microalps:
@AnEvilPenguin does this resolve the linked issues to your satisfaction? I will review this shortly and wondering if it covers your use case

AnEvilPenguin:
Sorry for the delay @microalps.  Yes this change does seem to fix all of the issues I brought up!

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -438,7 +478,7 @@ public void TestMap()
                         Hash.FromAnonymousObject(new { foo = Hash.FromAnonymousObject(new { bar = "c" }) })
                     }
                     }));
-            CollectionAssert.AreEqual(new[] { new { a = 1 }, new { a = 2 }, new { a = 3 }, new { a = 4 } },
+            CollectionAssert.AreEqual(new object[] { null, null, null, null },
```
microalps, line 481: This is a breaking change, albeit understandable. More on the actual line that causes this change.

src/DotLiquid/StandardFilters.cs
```diff
@@ -1048,22 +1014,47 @@ public static IEnumerable Where(IEnumerable input, string propertyName, object t
...
+                        .GetTypeInfo()
+                        .GetCustomAttributes(attributeType: typeof(LiquidTypeAttribute), inherit: false)
+                        .FirstOrDefault() as LiquidTypeAttribute;
+                    if (liquidTypeAttribute != null)
+                    {
+                        indexable = new DropProxy(obj, liquidTypeAttribute.AllowedMembers);
+                    }
+                    else if (TypeUtility.IsAnonymousType(type) && obj.GetType().GetRuntimeProperty(propertyName) != null)
+                    {
+                        return type.GetRuntimeProperty(propertyName).GetValue(obj, null);
```
microalps, line 1050: It definitely looks more efficient to not use RespondTo, but in the case the property is undefined, it returns null. This is different than before. Is there a reason for this change?
```
return element.RespondTo(property) ? element.Send(property) : element;
```

microalps (changes requested):

src/DotLiquid/StandardFilters.cs
microalps: For this task to continue, we will need to change this behavior so it doesn't cause a breaking change, or a rationale and get @daviburg to agree to it. Without running this, it seems like a bug fix, but need developer to confirm that.

microalps: In case this helps the case to prove it's a bug. https://replit.com/@microalps/LiquidWhereTests

src/DotLiquid.Tests/StandardFilterTests.cs
daviburg: There is this explicit change and the removed tests above. Regardless of whether the new behavior is right or wrong, this is indeed an explicit breaking change. As such, at minimum the change must be behind an opt-in version compatibility flag, so we don't risk breaking consuming projects.

src/DotLiquid/StandardFilters.cs
```diff
@@ -455,13 +451,14 @@ private static IEnumerable SortInternal(StringComparer stringComparer, object in
...
-            {
-                ary.Sort((a, b) => comparer.Compare(((IDictionary)a)[property], ((IDictionary)b)[property]));
-            }
-            else if (ary.All(o => o.RespondTo(property)))
+            else
             {
-                ary.Sort((a, b) => comparer.Compare(a.Send(property), b.Send(property)));
+                ary.Sort((a, b) =>
+                {
+                    var aPropertyValue = ResolveObjectPropertyValue(a, property);
```
daviburg, line 458: For readability reason, when using static method, I suggest putting the class name prefix, e.g. StandardFilters.ResolveObjectPropertyValue(a, property);

src/DotLiquid.Tests/FilterTests.cs
```diff
@@ -407,12 +407,10 @@ public void TestSort()
             _context["value"] = 3;
             _context["numbers"] = new[] { 2, 1, 4, 3 };
             _context["words"] = new[] { "expected", "as", "alphabetic" };
-            _context["arrays"] = new[] { new[] { "flattened" }, new[] { "are" } };
 
             CollectionAssert.AreEqual(new[] { 1, 2, 3, 4 }, new Variable("numbers | sort").Render(_context) as IEnumerable);
             CollectionAssert.AreEqual(new[] { "alphabetic", "as", "expected" }, new Variable("words | sort").Render(_context) as IEnumerable);
             CollectionAssert.AreEqual(new[] { 3 }, new Variable("value | sort").Render(_context) as IEnumerable);
-            CollectionAssert.AreEqual(new[] { "are", "flattened" }, new Variable("arrays | sort").Render(_context) as IEnumerable);
         }
```
daviburg, line 414: Why is sorting nested arrays no longer supported?

src/DotLiquid.Tests/StandardFilterTests.cs
daviburg: This test was supposed to map property "b" from the array of objects, but these objects don't have property b. The previous result is pretty bizarre. I struggle to find the value in that past behavior.

daviburg: While normally I do say that we have to provide compatibility with explicitly tested behavior, behaviors we have knowledge a consuming project is relying on, or behavior one could reasonably even if incorrectly expect, I don't understand how the past behavior of preserving the array made sense. Could it be that some consumers applied multiple maps in a chain, expecting the first matching property to be rendered??

src/DotLiquid.Tests/FilterTests.cs
microalps: Checked reference Ruby (same https://replit.com/@microalps/LiquidWhereTests) and it does indeed flatten nested arrays. This is a breaking change.

src/DotLiquid/StandardFilters.cs
```diff
@@ -433,14 +433,10 @@ public static IEnumerable SortNatural(object input, string property = null)
         private static IEnumerable SortInternal(StringComparer stringComparer, object input, string property = null)
         {
             List<object> ary;
-            if (input is IEnumerable<Hash> enumerableHash && !string.IsNullOrEmpty(property))
-                ary = enumerableHash.Cast<object>().ToList();
-            else if (input is IEnumerable enumerableInput)
-                ary = enumerableInput.Flatten().Cast<object>().ToList();
+            if (input is IEnumerable enumerable)
+                ary = enumerable.Cast<object>().ToList();
```
microalps, line 437: I guess these are the lines that need to be reverted to fix @daviburg issue with flattening.

microalps:
My changes that will resolve the flatten issue is in https://github.com/hieuvuvan/dotliquid/pull/1 - if the author accepts it we can keep this PR, otherwise I will create a new PR.

microalps:
Merged with PR #521

PR #510, opened 2022-12-11
behavior change: blank object condition

hieuvuvan:
implement shopify `empty`, `blank` condition check. The ability to easily check a_string, array is blank, for example:
```liquid
# before
{% if a_string != nil and a_string != '' %}
// do something
{% endif %}

# after
{% if a_string != blank %}
// do something
{% endif %}
```
ref:
- https://shopify.dev/api/liquid/basics#string
- https://shopify.dev/api/liquid/basics#empty

### DIFF
--- a/src/DotLiquid.Tests/ConditionTests.cs
+++ b/src/DotLiquid.Tests/ConditionTests.cs
@@ -43,6 +43,10 @@ namespace DotLiquid.Tests
                 return other == this.ToString();
             }
         }
+
+        public class DummyDrop : Drop
+        {
+        }
         #endregion
 
         // NOTE(David Burg): This forces sequential execution of tests, risk side effect resulting in non deterministic behavior.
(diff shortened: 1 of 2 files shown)


microalps:
It looks like the code for resolving the special names `empty` and `blank` is in Context.Resolve method. I would expect a change to simply enhance the code there to differentiate between `empty` and `blank` - currently they are the same.

FYI all `empty` tests were successful without any code change.

src/DotLiquid/Condition.cs
```diff
@@ -128,6 +144,36 @@ private static bool EqualVariables(object left, object right)
...
+            }
+        }
+
+        private static bool EmptyVariable(object obj)
+        {
+            if (obj == self)
+                return false;
+            switch (obj)
+            {
+                case string s when s == string.Empty:
```
microalps, line 169: string is an ienumerable<char> and covered by next line

src/DotLiquid/Condition.cs
```diff
@@ -128,6 +144,36 @@ private static bool EqualVariables(object left, object right)
             return left.SafeTypeInsensitiveEqual(right);
         }
 
+        private static bool BlankVariable(object obj)
+        {
+            if (obj == self)
+                return false;
+            switch (obj)
```
microalps, line 151: this can simply be placed in Context.Resolve method, e.g
```
                case "blank":
                    return new Symbol(o => {
                        switch (o)
                        {
                            case null:
                            case bool b when b == false:
                            case string s when string.IsNullOrWhiteSpace(s):
                            case IEnumerable e when !e.GetEnumerator().MoveNext():
                                return true;
                            default:
                                return false;
                        }
                    });
```

microalps (changes requested):

src/DotLiquid.Tests/ConditionTests.cs
```diff
@@ -695,6 +699,118 @@ public void TestEqualOperatorsWorksOnEnum()
...
+            _context["dict"] = new Dictionary<string, string> { { "abc", "xyz"} };
+            _context["empty_dict"] = new Dictionary<string, string> { };
+            _context["list"] = new List<string> { "abc" };
+            _context["empty_list"] = new List<string> { };
+            _context["arr"] = new string[] { "foo" };
+            _context["empty_arr"] = new string[] {};
+            _context["a_drop"] = new DummyDrop();
+
+            // self check
+            AssertEvaluatesFalse("blank", "==", "blank");
```
microalps, line 715: Daviburg is strict about using named parameters when two or more share the same type.
```suggestion
            AssertEvaluatesFalse(left: "blank", op: "==", right: "blank");
```

microalps:
Closed and created #522 to implement correctly

PR #513, opened 2022-12-11
fix condition check null string throw null pointer exception

hieuvuvan:


microalps:
@hieuvuvan couple of details needed here:
1. What's the expected outcome in the reverse - if null is the left and string is the right?
2. Can you post an example on replit that reproduces the same scenarios on ruby liquid and shows we have a difference in implementation?

microalps:
Replaced by #579

PR #514, opened 2022-12-11
fix default filter

hieuvuvan:
- fix default filter for `nil`, `false`, `empty array`
- support allow_false param: https://shopify.github.io/liquid/filters/default/

### DIFF
--- a/src/DotLiquid.Tests/StandardFilterTests.cs
+++ b/src/DotLiquid.Tests/StandardFilterTests.cs
@@ -1495,6 +1495,7 @@ PaulGeorge",
             Helper.AssertTemplateResult("foobar", "{{ '' | default: 'foobar' }}", assigns);
             Helper.AssertTemplateResult("foobar", "{{ empty_array | default: 'foobar' }}", assigns);
             Helper.AssertTemplateResult("12", "{{ array | default: 'foobar' }}", assigns);
+            Helper.AssertTemplateResult("1", "{{ 1 | default: 100 }}", assigns);
 
             // allow false parameter
             Helper.AssertTemplateResult("foobar", "{{ false | default: 'foobar', false }}", assigns);


src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -1487,10 +1487,19 @@ public void TestUrldecode()
...
             Helper.AssertTemplateResult("bar", "{{ var2 | default: 'foobar' }}", assigns);
             Helper.AssertTemplateResult("foobar", "{{ unknownvariable | default: 'foobar' }}", assigns);
+            Helper.AssertTemplateResult("foobar", "{{ false | default: 'foobar' }}", assigns);
+            Helper.AssertTemplateResult("foobar", "{{ '' | default: 'foobar' }}", assigns);
+            Helper.AssertTemplateResult("foobar", "{{ empty_array | default: 'foobar' }}", assigns);
+            Helper.AssertTemplateResult("12", "{{ array | default: 'foobar' }}", assigns);
+            Helper.AssertTemplateResult("1", "{{ 1 | default: 100 }}", assigns);
+
+            // allow false parameter
+            Helper.AssertTemplateResult("foobar", "{{ false | default: 'foobar', false }}", assigns);
```
microalps, line 1501: documentation shows that the correct syntax is with explicit named parameter, e.g. `, allow_false: true`

src/DotLiquid/StandardFilters.cs
```diff
@@ -861,9 +861,18 @@ public static object Modulo(Context context, object input, object operand)
...
         /// <param name="input">Input to be transformed by this filter</param>
         /// <param name="defaultValue">value to apply if input is nil, false or empty.</param>
-        public static string Default(string input, string @defaultValue)
+        /// <param name="allowFalse">if this value is true and input is false value. return false instead of defaultValue</param>
+        public static object Default(object input, object defaultValue, bool allowFalse = false)
         {
-            return !string.IsNullOrWhiteSpace(input) ? input : defaultValue;
+            if (input == null)
+                return defaultValue;
+            if (input is string s)
```
microalps, line 869: this test should be covered below with IEnumerable

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -1487,10 +1487,19 @@ public void TestUrldecode()
         [Test]
         public void TestDefault()
         {
-            Hash assigns = Hash.FromAnonymousObject(new { var1 = "foo", var2 = "bar" });
+            Hash assigns = Hash.FromAnonymousObject(new { var1 = "foo", var2 = "bar", array = new []{"1", "2"}, empty_array = new string[]{} });
             Helper.AssertTemplateResult("foo", "{{ var1 | default: 'foobar' }}", assigns);
             Helper.AssertTemplateResult("bar", "{{ var2 | default: 'foobar' }}", assigns);
             Helper.AssertTemplateResult("foobar", "{{ unknownvariable | default: 'foobar' }}", assigns);
+            Helper.AssertTemplateResult("foobar", "{{ false | default: 'foobar' }}", assigns);
```
microalps, line 1494: might want to use an explicit variable to be clear this is actually a boolean and not an unknown variable named 'false'

microalps (changes requested):

src/DotLiquid.Tests/StandardFilterTests.cs
microalps: See https://replit.com/@microalps/LiquidUnitTests#main.rb (test_default_ignores_two_args) for proof that a second argument is ignored and not treated the same as a named argument. So, this doesn't just not match the documentation, it is actually a false assertion.

PR #515, opened 2022-12-15
include tag should not loop if string variable

hieuvuvan:


src/DotLiquid.Tests/Tags/IncludeTagTests.cs
```diff
@@ -139,6 +154,19 @@ public void TestIncludeTagWithLocalVariables()
...
+        [Test]
+        public void TestIncludeTagWithStringVariable()
+        {
+            WithFileSystem(new DictionaryFileSystem(new Dictionary<string, string>
+            {
+                ["product"] = "Product: {{ product }}"
+            }), () =>
+            {
+                Assert.AreEqual("Product: foo", Template.Parse("{% include 'product' with 'foo' %}").Render());
+                Assert.AreEqual("Product: foo", Template.Parse("{% include 'product' for 'foo' %}").Render());
```
microalps, line 166: I would suggest another test where string is not hard-coded and is a variable from hash.

src/DotLiquid.Tests/Tags/IncludeTagTests.cs
```diff
@@ -224,5 +252,19 @@ public void TestIncludeFromTemplateFileSystem()
             }
             Assert.AreEqual(fileSystem.CacheHitTimes, 1);
         }
+
+        public void WithFileSystem(IFileSystem fs, Action action)
```
microalps, line 256: If we need this we should probably move to the `Helper` class and do a proper lock like there is in `LockTemplateStaticVars`

microalps (changes requested):

microalps:

PR #516, opened 2022-12-15
Feature partial cache

hieuvuvan:
- Partial cache template in include tag that request `ITemplateFileSystem#GetTemplate`, for example:
```liquid
{% include 'foo' %}
// do something
{% include 'foo' %} # Second call should be cached in local context
```
- introduce new break change in include tag. Current the `templateName` parameter passed into `ITemplateFileSystem` from include tag is actually template name expression - implementation of `ITemplateFileSystem` must call `context[templateName]` (`templateName = "'foo'"` for this example) to get final name, this change evaluate the expression before pass into `ITemplateFileSystem` (implementation of `ITemplateFileSystem` will be received`templateName = "foo"` for this example ).
- i know current we could implement this behavior by implementing ITemplateFileSystem, but it is nice if it is natural in lib and for render tag in feature

### DIFF
--- a/src/DotLiquid.Tests/Tags/IncludeTagTests.cs
+++ b/src/DotLiquid.Tests/Tags/IncludeTagTests.cs
@@ -111,6 +111,22 @@ namespace DotLiquid.Tests.Tags
             }
         }
 
+        private class CountingFileSystem : IFileSystem
+        {
+            public int Count { get; private set; }
+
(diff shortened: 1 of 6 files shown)


microalps:
Due to change in NUnit requirements, classic asserts cannot be merged after Build v2.3.0. Please rebase if you want this considered or this PR will be closed in 1 week.

microalps:
Closed until author is ready for rebase.

PR #517, opened 2022-12-15
Feature render tag

hieuvuvan:
- port render tag from ruby implementation. Render tag like include tag but more strict in term scope. for more detail here https://shopify.github.io/liquid/tags/template#render
- this PR include extra feature: the ability to disable tags. for example:
```csharp
var context = new Context(CultureInfo.InvariantCulture);
context.WithDisabledTags(new [] { "include" }, () =>
{
    Assert.AreEqual("include usage is not allowed in this context", Template.Parse("{% include 'foo' %}").Render(RenderParameters.FromContext(context, context.FormatProvider)))
});
```

### DIFF
--- a/src/DotLiquid.Tests/ContextTests.cs
+++ b/src/DotLiquid.Tests/ContextTests.cs
@@ -1089,5 +1089,88 @@ namespace DotLiquid.Tests
                 Assert.False(context.IsTagDisabled("bar"));
             });
         }
+
+        [Test]
+        public void TestStaticEnvironmentsAreReadWithLowerPriorityThanEnvironments()
+        {
+            var context = new Context(CultureInfo.InvariantCulture);
+            context.StaticEnvironments[0]["shadowed"] = "static";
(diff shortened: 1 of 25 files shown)


microalps:
1. Looks like there are several previous tasks that are dependencies before this can be merged. Can you describe them and let's chat on gitter about how to get them included?
2. Please make sure all ruby liquid tests are included in new features. e.g. https://github.com/Shopify/liquid/blob/master/test/integration/tags/render_tag_test.rb

microalps:

RodneyRichardson:
The render tag would be a very useful inclusion. Is this pull request dead?

microalps:
Yes, it's tagged as rebase required and initial comments were not addressed. Feel free to submit a new PR that addresses this issue. Also note, that the Golden Liquid tests relating to render should all pass.

RodneyRichardson:
I've created #567 to continue this.

PR #522, opened 2023-03-06
behavior change: blank object condition (Revised)

### LINKED ISSUE #523: Blank Condition Test
### Dotliquid version
2.2.685

### Expected behavior
Be able to test if a string is non-empty without checking null and empty. 

### Actual behavior
Blank and empty are synonymous.

https://replit.com/@microalps/LiquidBlankTests#main.rb

microalps:
Resolves #523

### DIFF
--- a/src/DotLiquid.Tests/ConditionTests.cs
+++ b/src/DotLiquid.Tests/ConditionTests.cs
@@ -43,6 +43,10 @@ namespace DotLiquid.Tests
                 return other == this.ToString();
             }
         }
+
+        public class DummyDrop : Drop
+        {
+        }
         #endregion
 
         // NOTE(David Burg): This forces sequential execution of tests, risk side effect resulting in non deterministic behavior.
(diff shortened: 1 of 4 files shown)


src/DotLiquid.Tests/ConditionTests.cs
```diff
@@ -695,6 +699,80 @@ public void TestEqualOperatorsWorksOnEnum()
...
+            _context["dict"] = new Dictionary<string, string> { { "abc", "xyz" } };
+            _context["empty_dict"] = new Dictionary<string, string> { };
+            _context["list"] = new List<string> { "abc" };
+            _context["empty_list"] = new List<string> { };
+            _context["arr"] = new string[] { "foo" };
+            _context["empty_arr"] = new string[] { };
+            _context["a_drop"] = new DummyDrop();
+
+            // self check
+            AssertEvaluatesFalse(left: "blank", op: "==", right: "blank");
```
daviburg, line 715: Why is the blank string not equal to itself?

daviburg:
Is there any Shopify Liquid reference documentation or Shopify open source test case to anchor our blank object condition to?

src/DotLiquid.Tests/ConditionTests.cs
```diff
@@ -695,6 +699,80 @@ public void TestEqualOperatorsWorksOnEnum()
             AssertEvaluatesFalse("enum", "!=", "'Yes'");
         }
 
+        [Test]
+        public void TestBlankObject()
+        {
+            _context = new Context(CultureInfo.InvariantCulture);
+            _context["dict"] = new Dictionary<string, string> { { "abc", "xyz" } };
```
daviburg, line 706: might benefit from using full English words for readability - i.e., dictionary and array.

src/DotLiquid/Context.cs
```diff
@@ -389,6 +389,18 @@ private object Resolve(string key, bool notifyNotFound = true)
                 case "false":
                     return false;
                 case "blank":
+                    return new Symbol(o => {
```
daviburg, line 392: Try to use full English words when possible for variable names instead of single letter o, b, s, etc.

src/DotLiquid/Context.cs
```diff
@@ -389,6 +389,18 @@ private object Resolve(string key, bool notifyNotFound = true)
                 case "false":
                     return false;
                 case "blank":
+                    return new Symbol(o => {
+                        switch (o)
+                        {
+                            case null:
+                            case bool b when b == false:
+                            case string s when string.IsNullOrWhiteSpace(s):
+                            case IEnumerable e when !e.GetEnumerator().MoveNext():
```
daviburg, line 398: Is that meant to be the same test as existing below `(o is IEnumerable enumerableO) && !enumerableO.Cast<object>().Any()`?

daviburg (commented):
Uncertain at this point. Waiting to learn more about what reference we may follow for this behavior.

src/DotLiquid/Context.cs
microalps: Yes, the original PR actually had this line for both. You pick which one you're more comfortable with.

src/DotLiquid.Tests/ConditionTests.cs
microalps: Added to my ruby tests - blank is not blank and empty is not empty.

https://replit.com/@microalps/LiquidBlankTests#main.rb

microalps: Also if you're looking for documentation about this, it simply doesn't exist. The entire test ruby script fails miserably without the line `require 'active_support/all'` because ruby and liquid doesn't define blank or empty. These are left to individual classes to define - and in this case Rails extends string and other common classes to define these methods for you. You'll find some more explanation on [stackoverflow](https://stackoverflow.com/questions/11029256/difference-between-nil-blank-and-empty)

daviburg: 🤯

daviburg: As the blank concept is inherited from Ruby on Rails, it appears logical to mimic that source's behavior.

daviburg: Consider capturing some explanation as an inline code comment for future maintainers.

src/DotLiquid/Context.cs
daviburg: While `.Any()` is elegant, `.Cast<object>()` to get to access IEnumerable<T> extensions is outright ugly. I don't want to explore the possible unexpected effects of casting an unknown quantity to object (... boxing types?).

Funny thing is it appears that the enumerator needs disposing.

https://stackoverflow.com/questions/52600634/c-sharp-interface-ienumerable-any-without-specifying-generic-types

and

https://stackoverflow.com/questions/13459447/do-i-need-to-consider-disposing-of-any-ienumerablet-i-use

daviburg: I suggested using the IEnumerable extension of the first link, without touching IEnumerable<T>.

microalps: Applied suggestion

src/DotLiquid/Util/EnumerableExtensionMethods.cs
```diff
@@ -33,5 +33,26 @@ public static void EachWithIndex(this IEnumerable<object> array, Action<object,
                 ++index;
             }
         }
+
+        /// <summary>
+        /// Determines whether a sequence contains any elements.
+        /// </summary>
+        /// <param name="source">The IEnumerable to check for emptiness.</param>
+        /// <returns>true if the source sequence contains any elements; otherwise, false.</returns>
```
daviburg, line 41:
```suggestion
        /// <returns><see langword="true"/> if the source sequence contains any elements; otherwise, <see langword="false"/>.</returns>
```

daviburg (approved):

daviburg (approved):

PR #526, opened 2023-03-23
Fix Param Whitespace Throws Exception

### LINKED ISSUE #496: Param Whitespace Throws Exception
### Dotliquid version
2.2.656

### Expected behavior
``` {% param date_format='ruby' %} ```
Behave the same as without the trailing whitespace

### Actual behavior
Liquid syntax error: Variable ''ruby' ' was not properly terminated
or (DotLiquid 2.0 Syntax)
Liquid error: Object reference not set to an instance of an object.

microalps:
Resolves #496

### DIFF
--- a/src/DotLiquid.Tests/Tags/ParamTest.cs
+++ b/src/DotLiquid.Tests/Tags/ParamTest.cs
@@ -35,13 +35,28 @@ namespace DotLiquid.Tests.Tags
             Assert.Throws<SyntaxException>(() => tag.Render(context, new StringWriter()));
         }
 
-        [Test]
-        public void TestSyntaxCompatibility()
+        [TestCase("param Syntax='DotLiquid21'")]
+        [TestCase(" param Syntax= 'DotLiquid21'")]
+        [TestCase("param Syntax= 'DotLiquid21'")]
+        [TestCase("param Syntax= 'DotLiquid21' ")]
+        [TestCase("param Syntax = 'DotLiquid21' ")]
+        public void TestSyntaxCompatibility(string markup)
         {
             // Initialize as DotLiquid20, then assert that the DotLiquid21 rules for Capitalize are followed.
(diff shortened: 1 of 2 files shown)


microalps:
The solution is a bit hackish because the following is silently ignored:
``` {% param Syntax = 'DotLiquid21' | replace: '21', '10' %} ```
I don't think we want to support filters here. Simple trim will pass entire string with filters to context, which will obviously fail but without any error handling or messaging that filters are not supported. Suggestions?

daviburg:
> The solution is a bit hackish because the following filter is silently ignored: `{% param Syntax = 'DotLiquid21' | replace: '21', '10' %}` I don't think we want to support filters here. Simple trim will pass entire string with filters to context, which will obviously fail but without any error handling or messaging that filters are not supported. Suggestions?

With simple string trim `{% param Syntax = 'DotLiquid21' | replace: '21', '10' %}` will turn into `{% param Syntax = 'DotLiquid21' | replace: '21', '10'%}`, yes?
What error does user get with this unsupported filter?

Are there scenarios where string trim doesn't fix a whitespace issue but the ().Name hack does fix?

microalps:
@daviburg 
> What error does user get with this unsupported filter?

Liquid syntax error: Invalid syntax option: 'DotLiquid21' | r...
> Are there scenarios where string trim doesn't fix a whitespace issue but the ().Name hack does fix?

Not that I know of. Reran the tests, they all passed.

daviburg:
> @daviburg
> 
> > What error does user get with this unsupported filter?
> 
> Liquid syntax error: Invalid syntax option: 'DotLiquid21' | r...
> 
> > Are there scenarios where string trim doesn't fix a whitespace issue but the ().Name hack does fix?
> 
> Not that I know of. Reran the tests, they all passed.

Then string trim seems a pretty clean option.

daviburg (approved):

PR #540, opened 2024-03-26
RegisterSafeType concurrent access error fixes

ozlemkzn:


src/DotLiquid/Template.cs
```diff
@@ -56,8 +56,8 @@ public class Template
         /// </summary>
         public static TimeSpan RegexTimeOut { get; set; }
 
-        private static readonly Dictionary<Type, Func<object, object>> SafeTypeTransformers;
-        private static readonly Dictionary<Type, Func<object, object>> ValueTypeTransformers;
+        private static readonly ConcurrentDictionary<Type, Func<object, object>> SafeTypeTransformers;
```
daviburg, line 59: Wouldn't this need to be accompanied by changes to adopt the thread safe dictionary methods per https://learn.microsoft.com/en-us/dotnet/standard/collections/thread-safe/how-to-add-and-remove-items
?

daviburg (changes requested):

src/DotLiquid/Template.cs
microalps: And would need accompanying tests to prove it works it resolved the issue.

microalps:
@saikrishna35 Thank you for correcting the code. While it looks good, would be nice to add a test that can verify this (a) works now (b) doesn't break down the road.

saikrishna35:
> @saikrishna35 Thank you for correcting the code. While it looks good, would be nice to add a test that can verify this (a) works now (b) doesn't break down the road.

Hi @microalps , I have raised one PR : https://github.com/ozlemkzn/dotliquid/pull/2..

src/DotLiquid/Template.cs
saikrishna35: Hi @daviburg, Updated the changes. Please check once.

daviburg: I see that you are using the proper method to update the concurrent dictionary now, but I do not yet understand how the conflict resolution logic avoids prior race conditions.

src/DotLiquid/Template.cs
```diff
@@ -174,7 +174,7 @@ public static void RegisterSafeType(Type type, string[] allowedMembers, Func<obj
         /// <param name="func">Function that converts the specified type into a Liquid Drop-compatible object (eg, implements ILiquidizable)</param>
         public static void RegisterSafeType(Type type, Func<object, object> func)
         {
-            SafeTypeTransformers[type] = func;
+            SafeTypeTransformers.AddOrUpdate(type, func, (k, oldFunction) => func);
```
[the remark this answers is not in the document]
saikrishna35: AddOrUpdate resolves concurrency issues by ensuring atomicity in the add or update operations. It guarantees that:

- If multiple threads attempt to add a new entry for the same key, only one will succeed, and the others will move to the update logic.
-  If multiple threads attempt to update an existing entry, AddOrUpdate ensures that each update is applied on the latest value, one after the other, 
preventing any thread from overwriting changes made by another thread without first seeing those changes.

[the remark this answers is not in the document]
daviburg: And you are not taking benefit from the update logic because your provided delegate always overwrites the old function without actually taking the old function into consideration.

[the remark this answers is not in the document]
daviburg: We are still back to square 1: demonstrate the issue through a unit test, leverage that unit test to prove that your change resolves the race condition through an orderly conflict (update) resolution.

PR #546, opened 2024-05-31
Fix floating numbers parse and respect culture

aeleftheriadis:
Fix StandardFilters methods that parses object to string, that fixes floating point numbers with current that has different decimal separator such as comma.

### DIFF
--- a/src/DotLiquid.Tests/StandardFilterTests.cs
+++ b/src/DotLiquid.Tests/StandardFilterTests.cs
@@ -1735,9 +1735,6 @@ PaulGeorge",
             Helper.AssertTemplateResult(
                 expected: "4",
                 template: "{{ 4 | abs }}");
-            Helper.AssertTemplateResult(
-                expected: "19,86",
-                template: "{{ '-19,86' | abs }}");
         }
 
         [Test]
--- a/src/DotLiquid.Tests/StandardFilterTests.cs
+++ b/src/DotLiquid.Tests/StandardFilterTests.cs
@@ -1407,7 +1407,7 @@ PaulGeorge",
 
                 Assert.Null(StandardFilters.Ceil(_contextV20, ""));
                 Assert.Null(StandardFilters.Ceil(_contextV20, "two"));
-                Assert.AreEqual(4, StandardFilters.Ceil(_contextGreekCultureV20, 3.5));
+                Assert.AreEqual(4, StandardFilters.Ceil(_contextGreekCultureV20, "3,5"));
             }
         }
 
@@ -1438,7 +1438,7 @@ PaulGeorge",
 
                 Assert.Null(StandardFilters.Floor(_contextV20, ""));
                 Assert.Null(StandardFilters.Floor(_contextV20, "two"));
-                Assert.AreEqual(3, StandardFilters.Floor(_contextGreekCultureV20, 3.5));
+                Assert.AreEqual(3, StandardFilters.Floor(_contextGreekCultureV20, "3,5"));
             }
         }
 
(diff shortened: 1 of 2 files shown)


microalps:
Please create corresponding tests for all the changed methods.

All the best,

Moshe Alfih
microalps web services

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -1675,6 +1713,30 @@ public void TestAbs()
...
+
+            Helper.AssertTemplateResult(
+                expected: "17",
+                template: "{{ -17 | abs }}");
+            Helper.AssertTemplateResult(
+                expected: "17",
+                template: "{{ 17 | abs }}");
+            Helper.AssertTemplateResult(
+                expected: "4",
+                template: "{{ 4 | abs }}");
```
microalps, line 1737: missing culture, what do these tests validate? And no example of fraction in this set of tests.

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -1699,19 +1761,42 @@ public void TestAtLeast()
         }
 
         [Test]
-        public void TestAtMost()
```
microalps, line 1702: do not remove existing tests unless there is a good justification for that.

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -1389,6 +1394,23 @@ public void TestCeil()
...
+            using (CultureHelper.SetCulture("el-GR"))
+            {
+                Helper.AssertTemplateResult("2", "{{ 1,2 | ceil }}");
+                Helper.AssertTemplateResult("2", "{{ 2,0 | ceil }}");
+                Helper.AssertTemplateResult("184", "{{ 183,357 | ceil }}");
+                Helper.AssertTemplateResult("4", "{{ \"3,5\" | ceil }}");
+
+                Assert.Null(StandardFilters.Ceil(_contextV20, ""));
+                Assert.Null(StandardFilters.Ceil(_contextV20, "two"));
+                Assert.AreEqual(4, StandardFilters.Ceil(_contextGreekCultureV20, 3.5));
```
microalps, line 1410: should this also have a test for passing a string as "3,5"?

microalps (changes requested):

aeleftheriadis:
@microalps  I have updated the tests as reviewed

aeleftheriadis (commented):
Reviewed as requested

microalps:
I will look into this shortly. Sorry I missed this!

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -1389,6 +1394,23 @@ public void TestCeil()
...
+
+        [Test]
+        public void TestCeil_DifferentCultureDecimalSeparator()
+        {
+            using (CultureHelper.SetCulture("el-GR"))
+            {
+                Helper.AssertTemplateResult("2", "{{ 1,2 | ceil }}");
+                Helper.AssertTemplateResult("2", "{{ 2,0 | ceil }}");
+                Helper.AssertTemplateResult("184", "{{ 183,357 | ceil }}");
+                Helper.AssertTemplateResult("4", "{{ \"3,5\" | ceil }}");
```
microalps, line 1406: I would suggest an additional test for thousands separators for both cultures. e.g. 1.234.567,89 should parse for el-GR and 1,234,567.89 for en.

microalps (changes requested):

microalps:
TestFloor_DifferentCultureDecimalSeparator and TestCeil_DifferentCultureDecimalSeparator do not fail without your fixes. There is missing test for intended scenario(s).

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -1389,6 +1394,23 @@ public void TestCeil()
...
+
+                Assert.Null(StandardFilters.Ceil(_contextV20, ""));
+                Assert.Null(StandardFilters.Ceil(_contextV20, "two"));
+                Assert.AreEqual(4, StandardFilters.Ceil(_contextGreekCultureV20, "3,5"));
```
microalps, line 1410: this is within a using and is the reason the test doesn't fail it seems

microalps (changes requested):
I think this covers the tests required

microalps (changes requested):
I think these are the changes needed...

microalps:
Another note: `Currency` method also calls ToString and presumably is also affected

microalps:
The more I think about this, it seems that the library incorrectly expects that each individual filter handles the culture setting. In my opinion, it would be more consistent if we simply changed the threads culture in `RenderInternal` if different than the current setting. The only side-effect would be the removal of the proprietary `{% param culture='en-US' %}` or change its behavior to set during initialization and not during render.

microalps:
Did some more research into this - a fix would probably not be backwards compatible. In testing, I found that `TestEuroCurrencyFromString` has various cultures with decimal separated fractions where those cultures should be comma separated. I assume that the same is true if anyone is relying on passing American numbers as strings with foreign cultures in the context/renderparameters. I'm baffled how to continue here.  I created #555 to track my work thus far on this issue.

microalps:
Closing this as #555 incorporates this patch and some.

daviburg:
> Did some more research into this - a fix would probably not be backwards compatible. In testing, I found that `TestEuroCurrencyFromString` has various cultures with decimal separated fractions where those cultures should be comma separated. I assume that the same is true if anyone is relying on passing American numbers as strings with foreign cultures in the context/renderparameters. I'm baffled how to continue here. I created #555 to track my work thus far on this issue.

Correct, backward compatibility would be a big issue for such type of change. We could only do this if we find a way to have it opt-in so current consumers don't suddenly get different outputs upon updating the version with no code change on their end.

PR #550, opened 2024-07-21
Migrated from .NETCore 2.0to .NET8. Updated all nugets packages to la…

freelearncoding:
Migrated Code to .NET8
Revise Test cases as per latest Nuget Pacakges.
Updated UseMVC method related code as per latest .NET8 version.


freelearncoding:
wrong account, will raise again

microalps:
Please note that changing the minimum requirements would be a breaking
change. Especially pulling support for .net framework (either net standard
2.0 or net4xx)

Issue #410, opened 2021-01-17
AppVeyor: NuGet API key expired


fubar-coder:
### Dotliquid version

n/a

### Expected behavior

IT seems that all commits to the master branch are automatically compiled and uploaded to NuGet. It's expected that the NuGet upload succeeds.

### Actual behavior

NuGet upload fails due to an expired API key.

### Steps to reproduce the Problem (you can add files)

Just take a look at the AppVeyor build logs

microalps:
Can @daviburg or @tgjones Close this? Resolved.

microalps:
This is happening again, @daviburg or @tgjones have access to fix?

daviburg:
I don't fully remember but I think only @tgjones had access to the AppVeyor account last time.

PR #552, opened 2024-11-18
Change AppVeyor CI account

microalps:


microalps:
@daviburg The intent here is to resume building automatically to nuget. Our previous setup was managed by @tgjones who has not updated the nuget API keys in a while. Resolved #410

daviburg (approved):

PR #553, opened 2024-11-19
Adjust regular expressions to allow for newlines within tags

### LINKED ISSUE #541: Line breaks in literal strings fail to parse
### Dotliquid version
[2.2.692](https://www.nuget.org/packages/DotLiquid/2.2.692)

### Expected behavior
The following template should parse:

**mytemplate.liquid**
```liquid
{% capture ATLANTIS_GH_TOKEN %}
--- begin ---
aaa
--- end ---
{% endcapture %}

{{ ATLANTIS_GH_TOKEN | replace: "
", "x" }}
```

### Actual behavior

```
MethodInvocationException: Exception calling "Parse" with "1" argument(s): "Variable '{{ ATLANTIS_GH_TOKEN | replace: "
", "x" }}' was not properly terminated with regexp: (?-mix:\}\})"
```

The following PowerShell sample fails to parse due to the literal line breaks:

```liquid
[DotLiquid.Template]::Parse(@"
{% capture ATLANTIS_GH_TOKEN %}
--- begin ---
aaa
--- end ---
{% endcapture %}
{{ ATLANTIS_GH_TOKEN | replace: "
", "x" }}
"@)

MethodInvocationException: Exception calling "Parse" with "1" argument(s): "Variable '{{ ATLANTIS_GH_TOKEN | replace: "
", "x" }}' was not properly terminated with regexp: (?-mix:\}\})"
```

whereas this one works fine without the literal line break:

```liquid
[DotLiquid.Template]::Parse(@"
{% capture ATLANTIS_GH_TOKEN %}
--- begin ---
aaa
--- end ---
{% endcapture %}
{{ ATLANTIS_GH_TOKEN | replace: "", "x" }}
"@)
```

### Additional Info

It seems that Liquid doesn't have any escape sequences for strings (e.g. "\n" for new line) so they have to be represented verbatim in strings - for line breaks this means actually putting a line break in the source code per above. Other parsers appear to support this - see:

* https://geekplayers.com/run-liquild-online.html
* https://jumpseller.co.uk/support/liquid-sandbox/

And this stackoverflow answer suggests the same thing:

* https://stackoverflow.com/questions/27694610/how-can-i-split-a-string-by-newline-in-shopify

microalps:
Resolves #541

### DIFF
--- a/src/DotLiquid.Tests/ParsingQuirksTests.cs
+++ b/src/DotLiquid.Tests/ParsingQuirksTests.cs
@@ -130,5 +130,36 @@ namespace DotLiquid.Tests
             Assert.AreEqual("}", Template.Parse("{{{}}}", SyntaxCompatibility.DotLiquid22).Render());
             Assert.AreEqual("{##}", Template.Parse("{##}", SyntaxCompatibility.DotLiquid22).Render());
         }
+
+        [TestCase("\r")]
+        [TestCase("\n")]
+        [TestCase("\r\n")]
+        public void TestBlockWithNewlines(string newLineChars)
+        {
+            var testString = "foo,bar".Replace(",", newLineChars);
+            Helper.AssertTemplateResult(testString, "{% assign test = \"" + testString + "\" %}{{ test }}");
+            Helper.AssertTemplateResult(testString, "{% assign test = '" + testString + "' %}{{ test }}");
+        }
+
+        [TestCase("\r")]
+        [TestCase("\n")]
+        [TestCase("\r\n")]
+        public void TestVariableWithNewlines(string newLineChars)
+        {
+            var testString = "foo,bar".Replace(",", newLineChars);
+            Helper.AssertTemplateResult(testString, "{{ \"" + testString + "\" }}");
+            Helper.AssertTemplateResult(testString, "{{ '" + testString + "' }}");
+        }
+
+        [TestCase("\r")]
+        [TestCase("\n")]
+        [TestCase("\r\n")]
+        public void TestFilterWithNewlines(string newLineChars)
+        {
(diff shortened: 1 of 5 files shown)


daviburg:
> @daviburg what are your thoughts here? I did not test against reference liquid library, but I noticed [they](https://github.com/Shopify/liquid/blob/main/lib/liquid/variable.rb) have a 'strict parsing' and 'lax parsing' mode which ours is neither (I assume we were lax at some point).
> 
> I further don't know how they handle the new `liquid` tag which intentionally has new lines, and whether this code will further complicate adding that feature in the future. `return @source.split("\n") if @for_liquid_tag` seems to indicate ruby tokenizes using new lines which would not work with variables containing new lines.

I agree that the Dot Liquid variable parsing behavior appears to be modeled after the lax parsing of Shopify's Ruby implementation, based on both sharing the RegEx based approach.

I suppose that for the liquid tag implementation, one would replace each line start and end within the liquid tag block with opening {% and closing %}, then only process the reformatted content with either the lax or strict parsing. Would that avoid the issue?

I admit I'm not fully grasping this complex problem.

microalps:
I'm moving this to draft as I'm not ready to have this proceed. My thought is that the impact on performance due to changed regular expressions will affect more people than the fix will handle. I can think of the following workarounds:
* Use newline_to_br filter
* Use RegexReplace filter with finer controls for \r or \n. 
* Use Capture tags to capture a new line and replace using said variable. Note that this is least recommended as it would depend on the newline used in the file itself. Your editor's newlines can be replaced silently by Git, FTP software (text vs ascii upload), etc, and have unexpected results.

microalps:
Abandoning in favor of new approach

Issue #535, opened 2023-11-26
Unsupported type DateOnly in .Net 7.0+


heggi:
### Dotliquid version

2.2.692

### Expected behavior

Render `DateOnly` as string and it needs to be compatible with `date` filter

### Actual behavior

Got error `Liquid syntax error: Object '30.11.2023' is invalid because it is neither a built-in type nor implements ILiquidizable`

### Steps to reproduce the Problem (you can add files)

```csharp
var rendered = DotLiquid.Template
    .Parse("{{ MyDate }}")
    .Render(DotLiquid.Hash.FromAnonymousObject(new
    {
        MyDate = new DateOnly(2023, 11, 30),
    }), CultureInfo.GetCultureInfo("ru-RU"));

Console.WriteLine(rendered);
```

microalps:
This requires that we start compiling .NET 6.0 specific builds. It's a work in progress - see https://github.com/microalps/dotliquid/tree/build-net6. Once this is done, it should be easy to whitelist these types.

PR #554, opened 2024-11-21
.NET 6.0 Build & NuGet Metadata

microalps:
Related to PR #550 that was abandoned.

Visually see how NuGet will display this at https://int.nugettest.org/packages/DotLiquid/2.2.724-dev

Other NuGet Metadata Changes: 
* Migrated to new license that matches the gist of LICENSE.md
* Removed dotliquidmarkup.org as project URL and added explicit source repository

This PR is a prerequisite for #535


microalps:
@daviburg couple of questions on how to proceed here:
1. Seems like .NET 6 has breaking changes to culture formatting. This is causing several tests to fail in VS.
2. Do we need to ensure testing .NET6 as part of CI? If yes, why didn't we have core 1 and 2 tested?

microalps:
@daviburg I added a commit to address the questions above in hopefully an acceptable way. Also waiting in the wings is commit for DateOnly support once this is merged. https://github.com/microalps/dotliquid/commit/e7856483652e017bb7c2db2eeef41d8ef24303f5

microalps:
@robin-parker maybe you want to chime in on the currency filter behavior?

src/DotLiquid.Tests/BlockTests.cs
```diff
@@ -78,7 +78,7 @@ public void TestWithCustomTagFactory()
             Template.RegisterTagFactory(new CustomTagFactory());
             Template result = null;
             Assert.DoesNotThrow(() => result = Template.Parse("{% custom %}"));
-            Assert.AreEqual("I am a custom tag"+Environment.NewLine, result.Render());
+            Assert.That(result.Render(), Is.EqualTo("I am a custom tag" +Environment.NewLine));
```
daviburg, line 81:
```suggestion
            Assert.That(result.Render(), Is.EqualTo("I am a custom tag" + Environment.NewLine));
```

daviburg (approved):

daviburg:
> @daviburg couple of questions on how to proceed here:
> 
> 1. Seems like .NET 6 has breaking changes to culture formatting. This is causing several tests to fail in VS.
>    
>    * Euro decimal separator was NBSP now it is NNBSP (narrow non-breaking space)
>    * 2-character culture names "es" and "fr" do not emit the € symbol rather ¤
> 2. Do we need to ensure testing .NET6 as part of CI? If yes, why didn't we have core 1 and 2 tested?

Yes, I experienced the same issue of breaking changes with .NET 6 culture formatting in other projects, not specifically the two you are listing, but in general changes in formatting for culture that ends up in differences in the serialized output.

I do believe we need some validation of the different targets in CI, as they are not all behaving the same hence we can overlook problems by not testing each and everyone.

I do not know why previous core versions were not tested.

daviburg:
@microalps try to use specific cultures (e.g., es-ES instead of es) to ensure complete localization data, see if that solves the issue of no currency symbol defined. It is likely due to implementing neutral cultures, e.g. es, without the defaults of a specific region, e.g. without the currency and other default of es-ES from the European zone.

src/DotLiquid.Tests/BlockTests.cs
microalps: I let VS reformat somewhat and am committing just space changes.

microalps:
> @microalps try to use specific cultures (e.g., es-ES instead of es) to ensure complete localization data, see if that solves the issue of no currency symbol defined. It is likely due to implementing neutral cultures, e.g. es, without the defaults of a specific region, e.g. without the currency and other default of es-ES from the European zone.

I found the csproj setting of `<RuntimeHostConfigurationOption Include="System.Globalization.UseNls" Value="true" />` would let me leave the neutral cultures in place, even though they are wrong. It would technically fail on Linux which doesn't have Nls, but in my opinion the tests are testing DotNet and not DotLiquid anyways in this case.

PR #555, opened 2024-11-26
Culture filter handling of numeric types

microalps:
Similar to #546

### DIFF
--- a/src/DotLiquid/StandardFilters.cs
+++ b/src/DotLiquid/StandardFilters.cs
@@ -346,6 +346,34 @@ namespace DotLiquid
             return input?.TrimEnd();
         }
 
+        private static bool TryParseDecimal(CultureInfo culture, object input, out decimal output)
+        {
+            if (IsReal(input))
+            {
+                try
+                {
+                    output = input is decimal dcmInput ? dcmInput : Convert.ToDecimal(input);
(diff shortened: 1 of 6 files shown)


src/DotLiquid/StandardFilters.cs
```diff
@@ -365,12 +393,7 @@ public static string Currency(Context context, object input, string languageTag
             var culture = languageTag == null ? context.CurrentCulture : new CultureInfo(languageTag.Trim());
 
             // Attempt to convert to a currency using the context current culture.
-            if (IsReal(input))
-                return Convert.ToDecimal(input).ToString("C", culture);
-            if (decimal.TryParse(input.ToString(), NumberStyles.Currency, context.CurrentCulture, out decimal amount))
-                return amount.ToString("C", culture);
-
-            return input.ToString();
+            return TryParseDecimal(context.CurrentCulture, input, out var d) ? d.ToString("C", culture) : input.ToString();
```
microalps, line 396: Note to reviewer: this doesn't make sense to me but is kept as-is to avoid breaking tests.

microalps: @robin-parker maybe you want to chime in on the currency filter behavior?

PR #559, opened 2024-12-05
Support Net6 DateOnly and TimeOnly 

### LINKED ISSUE #535: Unsupported type DateOnly in .Net 7.0+

microalps:
Resolves #535

### DIFF
--- a/src/DotLiquid.Tests/ContextTests.cs
+++ b/src/DotLiquid.Tests/ContextTests.cs
@@ -233,6 +233,19 @@ namespace DotLiquid.Tests
             Assert.That(_context["nil"], Is.EqualTo(null));
         }
 
+#if NET6_0_OR_GREATER
+        [Test]
+        public void TestVariables_NET60()
+        {
+            var dateOnly = new DateOnly(2013, 9, 10);
+            _context["dateonly"] = dateOnly;
+            Assert.That(_context["dateOnly"], Is.EqualTo(dateOnly));
+            var timeOnly = new TimeOnly(0, 10, 32);
+            _context["timeonly"] = timeOnly;
+            Assert.That(_context["timeonly"], Is.EqualTo(timeOnly));
+        }
+#endif
+
         private enum TestEnum { Yes, No }
 
         [Test]
--- a/src/DotLiquid.Tests/ContextTests.cs
+++ b/src/DotLiquid.Tests/ContextTests.cs
@@ -237,10 +237,10 @@ namespace DotLiquid.Tests
         [Test]
         public void TestVariables_NET60()
         {
-            var dateOnly = new DateOnly(2013, 9, 10);
+            var dateOnly = new DateOnly(year: 2013, month: 9, day: 10);
             _context["dateonly"] = dateOnly;
             Assert.That(_context["dateOnly"], Is.EqualTo(dateOnly));
-            var timeOnly = new TimeOnly(0, 10, 32);
+            var timeOnly = new TimeOnly(hour: 0, minute: 10, second: 32);
             _context["timeonly"] = timeOnly;
             Assert.That(_context["timeonly"], Is.EqualTo(timeOnly));
         }
--- a/src/DotLiquid/Util/StrFTime.cs
+++ b/src/DotLiquid/Util/StrFTime.cs
@@ -107,9 +107,7 @@ namespace DotLiquid.Util
         /// <param name="culture">the CurrentCulture to be used when formatting</param>
         /// <returns>a string version of date-time matching pattern.</returns>
         public static string ToStrFTime(this DateOnly dateOnly, string format, CultureInfo culture)
-        {
-            return ObjectToStrFTime(dateOnly, format, culture);
-        }
+            => ObjectToStrFTime(dateOnly, format, culture);
 
         /// <summary>
         /// Applies formatting consistent to the rules specified by the Ruby Time.strftime function.
@@ -125,9 +123,7 @@ namespace DotLiquid.Util
         /// <param name="culture">the CurrentCulture to be used when formatting</param>
         /// <returns>a string version of date-time matching pattern.</returns>
         public static string ToStrFTime(this TimeOnly timeOnly, string format, CultureInfo culture)
-        {
-            return ObjectToStrFTime(timeOnly, format, culture);
-        }
+            => ObjectToStrFTime(timeOnly, format, culture);
 #endif
 
         /// <summary>
@@ -144,9 +140,7 @@ namespace DotLiquid.Util
         /// <param name="culture">the CurrentCulture to be used when formatting</param>
         /// <returns>a string version of date-time matching pattern.</returns>
         public static string ToStrFTime(this DateTime dateTime, string format, CultureInfo culture)
-        {
-            return ObjectToStrFTime(dateTime, format, culture);
-        }
+            => ObjectToStrFTime(dateTime, format, culture);
 
         /// <summary>
         /// Formats a date using a ruby date format string
@@ -156,9 +150,7 @@ namespace DotLiquid.Util
         /// <param name="culture">the CurrentCulture to be used when formatting</param>
         /// <returns>a string version of date-time matching pattern.</returns>
         public static string ToStrFTime(this DateTimeOffset dateTimeOffset, string format, CultureInfo culture)
-        {
-            return ObjectToStrFTime(dateTimeOffset, format, culture);
-        }
+            => ObjectToStrFTime(dateTimeOffset, format, culture);
 
         private static string ObjectToStrFTime(object source, string format, CultureInfo culture)
         {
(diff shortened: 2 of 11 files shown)


microalps:
> Attention: Patch coverage is `93.97590%` with `5 lines` in your changes missing coverage. Please review.

The report is based on the .NET Framework build which doesn't have a path to this exception. However, the FormatException is tested in .NET6 builds.

daviburg:
> > Attention: Patch coverage is `93.97590%` with `5 lines` in your changes missing coverage. Please review.
> 
> The report is based on the .NET Framework build which doesn't have a path to this exception. However, the FormatException is tested in .NET6 builds.

In another thread there was a discussion about adding test passes in other .NET flavors. Where did we land on this topic?

src/DotLiquid.Tests/ContextTests.cs
```diff
@@ -233,6 +233,19 @@ public void TestVariables()
             Assert.That(_context["nil"], Is.EqualTo(null));
         }
 
+#if NET6_0_OR_GREATER
+        [Test]
+        public void TestVariables_NET60()
+        {
+            var dateOnly = new DateOnly(2013, 9, 10);
```
daviburg, line 240: nit: could use named parameters to disambiguate what integer is for what. It is easy to mix up the order as all are the same type. Although the year is blatant, the month and date could be either.

src/DotLiquid.Tests/ContextTests.cs
```diff
@@ -233,6 +233,19 @@ public void TestVariables()
...
         }
 
+#if NET6_0_OR_GREATER
+        [Test]
+        public void TestVariables_NET60()
+        {
+            var dateOnly = new DateOnly(2013, 9, 10);
+            _context["dateonly"] = dateOnly;
+            Assert.That(_context["dateOnly"], Is.EqualTo(dateOnly));
+            var timeOnly = new TimeOnly(0, 10, 32);
```
daviburg, line 243: same ambiguity - which is hour, minutes, seconds - named parameters would disambiguate.

src/DotLiquid/Util/StrFTime.cs
```diff
@@ -16,79 +16,116 @@ public static class StrFTime
...
-            { "a", (dateTime, culture) => dateTime.ToString("ddd", culture) },
-            { "A", (dateTime, culture) => dateTime.ToString("dddd", culture) },
-            { "b", (dateTime, culture) => dateTime.ToString("MMM", culture) },
-            { "B", (dateTime, culture) => dateTime.ToString("MMMM", culture) },
-            { "c", (dateTime, culture) => dateTime.ToString("ddd MMM dd HH:mm:ss yyyy", culture) },
-            { "C", (dateTime, culture) => ((int)Math.Floor(Convert.ToDouble(dateTime.ToString("yyyy", culture))/100)).ToString(culture) },
-            { "d", (dateTime, culture) => dateTime.ToString("dd", culture) },
-            { "D", (dateTime, culture) => dateTime.ToString("MM/dd/yy", culture) },
-            { "e", (dateTime, culture) => dateTime.ToString("%d", culture).PadLeft(2, ' ') },
+            { "a", (dateTime, culture) => string.Format(culture, "{0:ddd}", dateTime) },
```
daviburg, line 24: Is it me or is there no matching test for "a" format? As we are changing the line and the format from `ddd` to `{0:ddd}` I was wondering if the output remains indeed the same (should be).

Looks like reformatting to adhere to a code style, removing ambiguity of extension methods and avoiding any possible null deref.

src/DotLiquid/Util/StrFTime.cs
```diff
@@ -124,17 +150,28 @@ public static string ToStrFTime(this DateTime dateTime, string format, CultureIn
...
+                    evaluator: specifier => SpecifierEvaluator(
+                        specifier: specifier.Groups[0].Value,
+                        flags: specifier.Groups[GROUP_FLAGS].Captures.Cast<Capture>().Select(capture => capture.Value).ToList(),
+                        width: specifier.Groups[GROUP_WIDTH].Captures.Cast<Capture>().Select(capture => (int?)Convert.ToInt32(capture.Value)).FirstOrDefault(),
+                        directive: specifier.Groups[GROUP_DIRECTIVE].Captures.Cast<Capture>().Select(capture => capture.Value).FirstOrDefault(),
+                        source: source,
+                        culture: culture
+                        ));
+            }
+            catch (FormatException ex) when (ex.Message == "Only")
```
daviburg, line 171: Exception should by design by localized, such as programmatic string equality checks are dangerous. Is there any other programmatic check we could rely on for this filtering?

src/DotLiquid/Util/StrFTime.cs
```diff
@@ -124,17 +150,28 @@ public static string ToStrFTime(this DateTime dateTime, string format, CultureIn
...
+            {
+                throw new FormatException($"String '{format}' contains parts which are not specific to the {source.GetType().Name}.");
```
daviburg, line 173: Likely not implemented localization in the repo, so just noting but likely not actionable in this PR, exception messages should be localized. They should come from a resource file (i.e. *.resx) and then translations can be provided for each culture.

src/DotLiquid/Util/StrFTime.cs
microalps: This error is thrown in two places in Liquid not by .NET itself. We could make a custom internal exception, but seems like a lot of boilerplate code for something internal.

src/DotLiquid/Util/StrFTime.cs
microalps: Noted. There is localization in the project with Italian as well, but in recent PR we haven't been attentive to properly address.

src/DotLiquid/Util/StrFTime.cs
microalps: Reformatting is to allow for various date types to utilize the same codebase rather than have a version for each primitive type. Test is one of the test cases listed on TestToStrFTime_DateFormat, specifically, `[TestCase("%a", ExpectedResult = "Sun")] // Abbreviated weekday name (Mon)`

src/DotLiquid/Util/StrFTime.cs
daviburg: If this is a local exception then absolutely we should be using our own exception definition. It is a common best practice to define own exception types and avoid throwing generic system exceptions. But, it is also a breaking change if that exception was not only internal but also sometimes thrown to the caller of our APIs. So proceed with caution.

src/DotLiquid/Util/StrFTime.cs
```diff
@@ -16,79 +16,116 @@ public static class StrFTime
...
         };
 
-        private static readonly Dictionary<string, DateTimeOffsetDelegate> OffsetFormats = new Dictionary<string, DateTimeOffsetDelegate>
+        private static readonly Dictionary<string, DateTimeDelegate> DateFormats = new Dictionary<string, DateTimeDelegate>
         {
-            { "s", (dateTimeOffset, culture) => ((long)(dateTimeOffset - new DateTimeOffset(1970, 1, 1, 0,0,0, TimeSpan.Zero)).TotalSeconds).ToString(culture) },
-            { "z", (dateTimeOffset, culture) => dateTimeOffset.ToString("%K", culture).Replace(":", string.Empty) },
-            { ":z", (dateTimeOffset, culture) => dateTimeOffset.ToString("%K", culture) },
-            { "Z", (dateTimeOffset, culture) => dateTimeOffset.ToString("zzz", culture) }
+            { "g", (dateTime, culture) => dateTime.GetIso8601WeekOfYear("g", culture) },
```
microalps, line 85: Note to self - there is no reason why DateOnly is excluded from this functionality. TimeOnly though is another story. Existing test also is not asserting anything...

src/DotLiquid/Util/StrFTime.cs
microalps: It is thrown on https://github.com/dotliquid/dotliquid/pull/559/files#diff-b7bfc6e3a7d66bf9e9f2047c1695a75add8ea64a5c7afd95630ea7b651c762e9R204 (strftime.cs line 66 and 204). I'm not clear what the expectation is - in those areas we don't have scope to format the message so we throw an exception, catch where we do have more info and properly format the message.

daviburg: Just noting for future reference that this comment was addressed with the addition of DotLiquid.Exceptions.DateFormatInvalidException custom internal exception, which removed the need to rely on exception message string matching test.

daviburg (approved):

PR #561, opened 2024-12-09
Reduce Dependence on Regex (2024)

### LINKED ISSUE #541: Line breaks in literal strings fail to parse

microalps:
Resolve #541 

Speeds up parsing significantly and rendering (Ruby naming convention specifically) by a bit based on Benchmark.NET.

```
|              Method |        Job | BuildConfiguration | syntaxCompatibilityLevel |      Mean |     Error |   StdDev | Ratio | RatioSD |   Gen 0 |  Gen 1 | Gen 2 | Allocated |
|-------------------- |----------- |------------------- |------------------------- |----------:|----------:|---------:|------:|--------:|--------:|-------:|------:|----------:|
|               Parse | Job-IETHXO |              NuGet |              DotLiquid22 |  21.61 us |  1.584 us | 0.411 us |  1.00 |    0.00 |  2.6245 | 0.0305 |     - |     16 KB |
|               Parse | Job-DUZQPM |            Default |              DotLiquid22 |  13.80 us |  0.411 us | 0.107 us |  0.64 |    0.01 |  1.9836 | 0.0305 |     - |     12 KB |
|                     |            |                    |                          |           |           |          |       |         |         |        |       |           |
|            ParseBig | Job-IETHXO |              NuGet |              DotLiquid22 |  82.36 us |  4.462 us | 1.159 us |  1.00 |    0.00 | 15.3809 | 1.0986 |     - |     95 KB |
|            ParseBig | Job-DUZQPM |            Default |              DotLiquid22 |  42.84 us |  2.915 us | 0.757 us |  0.52 |    0.01 | 12.2070 | 0.8545 |     - |     75 KB |
|                     |            |                    |                          |           |           |          |       |         |         |        |       |           |
|    RenderNestedHash | Job-IETHXO |              NuGet |              DotLiquid22 | 292.46 us | 43.190 us | 6.684 us |  1.00 |    0.00 | 47.3633 | 7.3242 |     - |    293 KB |
|    RenderNestedHash | Job-DUZQPM |            Default |              DotLiquid22 | 246.08 us | 20.337 us | 3.147 us |  0.84 |    0.03 | 49.8047 | 7.3242 |     - |    307 KB |
|                     |            |                    |                          |           |           |          |       |         |         |        |       |           |
| RenderILiquidizable | Job-IETHXO |              NuGet |              DotLiquid22 | 335.61 us | 22.986 us | 3.557 us |  1.00 |    0.00 | 50.7813 | 5.8594 |     - |    314 KB |
| RenderILiquidizable | Job-DUZQPM |            Default |              DotLiquid22 | 302.96 us | 35.106 us | 9.117 us |  0.90 |    0.03 | 53.2227 | 6.3477 |     - |    328 KB |

````

### CODE REVIEW

File: src/DotLiquid.Tests/Tags/ParamTest.cs [hunk 1]
```diff
@@ -29,11 +29,12 @@ public void TestInitialize_SyntaxValidation()
         [TestCase("Syntax = 'DotLiquid21' | replace: '21', '10'")] // Filters is not valid in this tag
         public void TestInvalidOptions(string markup)
         {
-            var tag = new Param();
-            tag.Initialize(tagName: "param", markup: markup, tokens: null);
-
-            var context = new Context(new CultureInfo("en-US"));
-            Assert.Throws<SyntaxException>(() => tag.Render(context, new StringWriter()));
+            Assert.Throws<SyntaxException>(() => {
```
daviburg, line 32: Why this refactoring? The previous code was setting up the text context and only asserting on the specific call that is expected to throw.

src/DotLiquid/Block.cs
```diff
@@ -13,11 +13,8 @@ namespace DotLiquid
     /// </summary>
     public class Block : Tag
     {
-        private static readonly Regex IsTag = R.B(@"^{0}", Liquid.TagStart);
-        private static readonly Regex IsVariable = R.B(@"^{0}", Liquid.VariableStart);
-        private static readonly Regex ContentOfVariable = R.B(@"^{0}(.*){1}$", Liquid.VariableStart, Liquid.VariableEnd);
 
-        internal static readonly Regex FullToken = R.B(@"^{0}\s*(\w+)\s*(.*)?{1}$", Liquid.TagStart, Liquid.TagEnd);
+        internal static readonly Regex FullToken = R.B(@"^{0}\s*(\w+)\s*((?s:.*)?){1}$", Liquid.TagStart, Liquid.TagEnd);
```
daviburg, line 17: Oh boy, which regex syntax is that? The {0} and {1} are those C# string interpolations or actual regex pieces?

src/DotLiquid/Block.cs
```diff
@@ -31,8 +28,7 @@ protected override void Parse(List<string> tokens)
             string token;
             while ((token = tokens.Shift()) != null)
             {
-                Match isTagMatch = IsTag.Match(token);
-                if (isTagMatch.Success)
+                if (token.Length > 4 && token.Substring(0, 2) == Tokenizer.TagStart)
```
daviburg, line 31: Generally C# code analyzers frown upon the use of string equality operator and want instead to use the explicit string.Equal with a culture specifier such as ordinal case sensitive (or not)

daviburg (commented):
Not done reviewing but sharing early simple comments.

Comparing equivalency of Regex to new code will be challenging.

src/DotLiquid/Block.cs
microalps: yep, it's combining regex for start, finish, and anything in between into one. Not my idea...

src/DotLiquid.Tests/Tags/ParamTest.cs
microalps: Good catch. I refactored because the exception is now thrown earlier in the process.

src/DotLiquid/Block.cs
microalps: might be even better to just compare token[0] and token[1] as chars. 
```
if (token.Length > 4 && token[0] == Tokenizer.CharCurlyBraceStart && token[1] == Tokenizer.CharPercent)
```

microalps:
Merged Golden Liquid tests #566. This resolves 19 issues in total:
18 are fixed by the new parsing engine, of which 4 were failing due to differences in new lines as well.
1 is fixed due to the less sensitive end-of-line comparison, and not related to parsing engine.

src/DotLiquid/Tokenizer.cs
```diff
@@ -12,14 +12,29 @@ namespace DotLiquid
...
     {
+        internal const string TagStart = "{%";
+        internal const char CharPipe = '|';
+        internal const char CharColon = ':';
+        internal const char CharComma = ',';
+        internal const char CharEquals = '=';
+        internal const char CharSingleQuote = '\'';
+        internal const char CharDoubleQuote = '"';
+        internal const char CharCurlyBraceStart = '{';
+		internal const char CharCurlyBraceEnd = '}';
```
RodneyRichardson, line 23: Formatting - Too many spaces (tabs?) at the start

src/DotLiquid/Tokenizer.cs
```diff
@@ -169,7 +184,7 @@ internal static IEnumerator<string> GetVariableEnumerator(string source)
         /// </summary>
         /// <param name="markupEnumerator">The string enumerator</param>
         /// <param name="markupLength">The number of characters to read</param>
-        private static string ReadChars(CharEnumerator markupEnumerator, int markupLength)
+        internal static string ReadChars(CharEnumerator markupEnumerator, int markupLength)
```
RodneyRichardson, line 187: I'm always wary when a private method needs to be made accessible. This is usually a smell that we're tying classes together for convenience that may not have a real connection if designed from scratch. Same for `ReadToEndofTag`.
Also, why is this internal and the other is public (on an internal class)?

RodneyRichardson (commented):
Although this does simplify things a little, and move logic away froma central location into the specific Tags, I find it a little harder to understand. Maybe each tag just needs a comment to descibe exactly what it's looking for in the text input.

RodneyRichardson:
Should this go to develop branch and check overlap with #582?

Issue #566, opened 2025-01-27
Liquid Golden Test Suite


microalps:
### Issue
Initial testing of the [Golden Liquid tests ](https://github.com/jg-rp/golden-liquid) indicate that 46 tests are failing on the latest version. Note that in-depth debugging was not performed and there might be a coding mistake that led to some tests failing.

### Expectation
DotLiquid is more consistent with Ruby Liquid and passes most if not all the tests. 

### Examples - Object Reference Errors
```
 1)   liquid.golden.abs_filter - undefined left value
  Expected: "0"
  But was:  "Liquid error: Object reference not set to an instance of an o..."
 7)   liquid.golden.at_least_filter - undefined argument
  Expected: "5"
  But was:  "Liquid error: Object reference not set to an instance of an o..."
 10)   liquid.golden.at_most_filter - undefined argument
  Expected: "0"
  But was:  "Liquid error: Object reference not set to an instance of an o..."
 15)   liquid.golden.ceil_filter - undefined left value
  Expected: "0"
  But was:  "Liquid error: Object reference not set to an instance of an o..."
 20)   liquid.golden.cycle_tag - multiple undefined variable names
  Expected: "123"
  But was:  "Liquid error: Object reference not set to an instance of an o..."
```

### Examples - Date Filter Mishandling
```
 25)   liquid.golden.date_filter - seconds since epoch format directive
  Expected: "1457913600"
  But was:  "1457928000"

 26)   liquid.golden.date_filter - timestamp string
  Expected: "07/05/2006"
  But was:  "1152098955"
```

RodneyRichardson:
See #570 and #572

microalps:
I should have mentioned I started working on it in https://github.com/microalps/dotliquid/tree/golden-test

microalps:
There is a class of errors related to undefined arguments. I'm not sure what the rule is, but it seems that several factors need to be considered:

* Is the argument optional? If yes, and it was omitted completely, that's OK
* Is the argument a string? If yes, null/undefined variables are OK
* Is the argument an integer or similar primitive type? 
** Slice - offset is integer without a default. Tests indicate that's not OK
** Slice - length is integer with default of 1. Tests indicate that's OK
** Truncatewords - words is int with default of 15. Tests indicate that's NOT OK.

Much more research is needed though @jg-rp do you know when liquid does/doesn't accept null arguments?

jg-rp:
> do you know when liquid does/doesn't accept null arguments?

[...]

PR #569, opened 2025-02-20
Add Golden Liquid test suite/project

RodneyRichardson:
Added a test project to execute the Golden Liquid test suite, as per #566. 

I wasn't exactly sure about a few things:
- Am I correct passing the "partials" field as the "Registers" parameter?
- Have I correctly created a Hash object from the JSON for context and partials?

I have marked specific test groups to be skipped for tags that are not implemented (echo, liquid and render).

### DIFF
--- a/src/DotLiquid.Golden.Tests/DotLiquid.Golden.Tests.csproj
+++ b/src/DotLiquid.Golden.Tests/DotLiquid.Golden.Tests.csproj
@@ -1,0 +1,80 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net461;net6.0;netcoreapp2.0</TargetFrameworks>
+    <AssemblyName>DotLiquid.Golden.Tests</AssemblyName>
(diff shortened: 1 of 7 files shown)


microalps:
Let's collaborate how we can incorporate this in a way that allows us to test all existing functionality doesn't regress, all broken functionality fails as expected, and that bug fixes for those test can confirm the failing tests are moved from the fail bucket to the pass bucket. Also, we might want to have a bucket of tests that we know will result in slightly different output (such as floating precision related tests), essentially allowing for alternate "want" values that are acceptable.

RodneyRichardson:
> Let's collaborate how we can incorporate this in a way that allows us to test all existing functionality doesn't regress, all broken functionality fails as expected, and that bug fixes for those test can confirm the failing tests are moved from the fail bucket to the pass bucket. Also, we might want to have a bucket of tests that we know will result in slightly different output (such as floating precision related tests), essentially allowing for alternate "want" values that are acceptable.

I like your approach in https://github.com/microalps/dotliquid/tree/golden-test. I'm happy to adopt that, and kill this PR. Would you be able to open a PR for that?

PR #571, opened 2025-02-20
Fix Numeric filters

### LINKED ISSUE #570: abs throws an exception on nil variable
### Dotliquid version
2.3.18

### Expected behavior

`abs` should render `0` when the variable is nil (null in C#).

### Actual behavior

Throws a NullReferenceException.

Render the following template

```
{%- assign my_variable = nil -%}
"{{ my_variable }}"
-{{ my_variable | abs }}-
```

RodneyRichardson:
Handles numeric parameters consistently across standard filters
* Fixes filters that control the range of numbers: Abs, AtMost, AtLeast, Round, Ceil, Floor
* Fixes numeric operation filters: Plus, Minus, DividedBy, Modulo, Times
* Adds Sum filter.

Uses the following logic:
* Performs conversion from string representation of numbers (including Thousands separators)
* Non-numeric parameters are treated as zero (null, non-numeric strings, etc).
* The return type matches the input type, when possible, or the next size type to handle overflow. 

Closes #570 and #483.


microalps:
Is this bug exclusive to Abs method? I noted several methods with the same issue in #566

RodneyRichardson:
> Is this bug exclusive to Abs method? I noted several methods with the same issue in #566

It also shows up in at_least and at_most, but they don't return the expected behaviour when fixed! I think I will fix this bug on them, and add it to this PR, and add tests for the current incorrect behaviour. We can then fix that when we re-integrate your golden tests and resolve the issues one at a time. Not sure how you want to approach changing the behaviour when current behaviour doesn't match the Ruby/shopify behaviour.

microalps:
> > Is this bug exclusive to Abs method? I noted several methods with the same issue in #566
> 
> It also shows up in at_least and at_most, but they don't return the expected behaviour when fixed! I think I will fix this bug on them, and add it to this PR, and add tests for the current incorrect behaviour. We can then fix that when we re-integrate your golden tests and resolve the issues one at a time. Not sure how you want to approach changing the behaviour when current behaviour doesn't match the Ruby/shopify behaviour.

I think the behavior there is that null/"abc" as the input is treated the same as 0. Hence `null | at_least "5"` = 5 and `null | at_least "-5"` = 0

Also see my latest push for Golden tests, there is an entire set of tests marked failing for "Non-integer arguments for numeric methods"

microalps:
Comments in https://github.com/dotliquid/dotliquid/pull/571#issuecomment-2674956771 were not addressed.

RodneyRichardson:
> Comments in [#571 (comment)](https://github.com/dotliquid/dotliquid/pull/571#issuecomment-2674956771) were not addressed.

Ah, I missed the abs/ceil/floor comment. I'll sort that too.

RodneyRichardson:
Fixed for Ceil and Floor. I've also tweaked Abs to be consistent in implementation (e.g. return decimal, not double).

RodneyRichardson:
`Round` also seems to fit into this, but is a bit more complicated. 

General behaviour seems to be
- if `input` is invalid, then return 0
- if `places` is invalid, then it rounds to zero decimal places. 

Note: The Ruby implementation for a negative value for `places` returns 0, but this feels inconsistent with the other invalid `places`, so I've implemented it differently. I round to zero decimal places. 

Similarly, 28 places is the maximum .NET will round to. I don't know what Ruby implementation does, but I round to zero. We could round to 28 if it's greater than 28, or return 0.

src/DotLiquid/StandardFilters.cs
```diff
@@ -766,21 +766,27 @@ public static object Times(Context context, object input, object operand)
...
             {
-                return null;
+                const decimal MinDecimalPlaces = 0m;
+                const decimal MaxDecimalPlaces = 16m;
+                if (decimal.TryParse(places?.ToString(), NumberStyles.Any, context.CurrentCulture, out decimal p))
+                {
+                    p = Math.Max(MinDecimalPlaces, Math.Min(MaxDecimalPlaces, p));
+                    int decimals = (int)Math.Floor(p);
+                    return Math.Round(d, decimals);
+                }
```
daviburg, line 786:
```suggestion
                }


```

src/DotLiquid/StandardFilters.cs
```diff
@@ -766,21 +766,27 @@ public static object Times(Context context, object input, object operand)
...
+                return Math.Round(d);
             }
+            return 0m;
```
daviburg, line 789:
```suggestion

            return 0m;
```

src/DotLiquid/StandardFilters.cs
```diff
@@ -920,10 +926,12 @@ public static IEnumerable Uniq(object input)
         /// </summary>
         /// <param name="context">The DotLiquid context</param>
         /// <param name="input">Input to be transformed by this filter</param>
-        public static double Abs(Context context, object input)
+        public static object Abs(Context context, object input)
         {
-            Double n;
-            return Double.TryParse(input.ToString(), NumberStyles.Number, context.CurrentCulture, out n) ? Math.Abs(n) : 0;
+            if (decimal.TryParse(input?.ToString(), NumberStyles.Any, context.CurrentCulture, out decimal d))
```
daviburg, line 931: double is very different from decimal in its handling of precision error.  Changing between them will generate different results which can be a regression for some customers. This change should only be done behind a version flag to avoid breaking DotLiquid consumers.

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -1438,11 +1448,13 @@ public void TestFloor()
             {
                 Helper.AssertTemplateResult("1", "{{ 1.2 | floor }}");
                 Helper.AssertTemplateResult("2", "{{ 2.0 | floor }}");
-                Helper.AssertTemplateResult("183", "{{ 183.357 | floor }}");
+                Helper.AssertTemplateResult("183", "{{ 183.57 | floor }}");
```
daviburg, line 1451: why change the test input?

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -1409,10 +1409,18 @@ public void TestRound()
...
+                Helper.AssertTemplateResult("1.235", "{{ 1.234678 | round: 3.1 }}");
                 Helper.AssertTemplateResult("1", "{{ 1 | round }}");
-
-                Assert.That(StandardFilters.Round("1.2345678", "two"), Is.Null);
+                Helper.AssertTemplateResult("1", "{{ 1.234678 | round }}");
+                Helper.AssertTemplateResult("1", "{{ 1.234678 | round: -3 }}");
+                Helper.AssertTemplateResult("1", "{{ 1.234678 | round: nonesuch }}");
+                Helper.AssertTemplateResult("0", "{{ nonesuch | round }}");
+
+                Assert.That(StandardFilters.Round(_contextV20, "1.2345678", "two"), Is.EqualTo(1m));
```
daviburg, line 1421: This kind of change may have unintended consequences, e.g. breaking DotLiquid consumers who have built a dependency on the output.

daviburg:
> > Is this bug exclusive to Abs method? I noted several methods with the same issue in #566
> 
> It also shows up in at_least and at_most, but they don't return the expected behaviour when fixed! I think I will fix this bug on them, and add it to this PR, and add tests for the current incorrect behaviour. We can then fix that when we re-integrate your golden tests and resolve the issues one at a time. Not sure how you want to approach changing the behaviour when current behaviour doesn't match the Ruby/shopify behaviour.

Fixing mathematics in DotLiquid was one of my earliest contribution to the code base, and got me in trouble because 'bug fixing' meant breaking consumers of the library who took a dependency on the current output. Hence the later introduction of compatibility versioning in the project. IMHO, this is one such case that needs to go behind a version flag.

daviburg:
> `Round` also seems to fit into this, but is a bit more complicated.
> 
> General behaviour seems to be
> 
> * if `input` is invalid, then return 0
> * if `places` is invalid, then it rounds to zero decimal places.
> * if `places` is a valid number, do `floor(places)` then clamp it in the range [0, 16]
> 
> Ruby seems to limit `places` to 16 decimal places, so I've done this too: https://replit.com/@rrjames/Liquid-Tests#main.rb
> 
> Note: The Ruby implementation for a negative value for `places` returns 0, but this feels inconsistent with the other invalid `places`, so I've implemented it differently and I round to zero decimal places.
> 
> @microalps, @daviburg : Thoughts?

We should not artificially limit ourselves to fewer decimal places than our underlying implementation language supports. Otherwise, Liquid would be driven by the lowest common denominator amongst all implementations.

src/DotLiquid/StandardFilters.cs
microalps: I had the same concerns, with math functions we did double work, catching an overflow exception before reverting to decimal types.

src/DotLiquid.Tests/StandardFilterTests.cs
daviburg: But we already have 3.5 floor on the very next test line to distinguish from round.

src/DotLiquid.Tests/StandardFilterTests.cs
daviburg: Right, I picked this example too fast - string for number input was never valid.

src/DotLiquid/StandardFilters.cs
daviburg: I can get behind a catch exception and recover from error; that should not change output for currently working calls, and we don't need to version-protect valid input failures turned into success.

src/DotLiquid/StandardFilters.cs
RodneyRichardson: Done.

src/DotLiquid/StandardFilters.cs
```diff
@@ -766,21 +766,27 @@ public static object Times(Context context, object input, object operand)
...
-                var i = Convert.ToDecimal(input);
-                return Math.Round(i, p);
-            }
-            catch (Exception)
+            if (decimal.TryParse(input?.ToString(), NumberStyles.Any, context.CurrentCulture, out decimal d))
             {
-                return null;
+                const decimal MinDecimalPlaces = 0m;
+                const decimal MaxDecimalPlaces = 16m;
+                if (decimal.TryParse(places?.ToString(), NumberStyles.Any, context.CurrentCulture, out decimal p))
```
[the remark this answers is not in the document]
RodneyRichardson: Done.

src/DotLiquid/StandardFilters.cs
RodneyRichardson: Done

RodneyRichardson:
> > > Is this bug exclusive to Abs method? I noted several methods with the same issue in #566
> > 
> > 
> > It also shows up in at_least and at_most, but they don't return the expected behaviour when fixed! I think I will fix this bug on them, and add it to this PR, and add tests for the current incorrect behaviour. We can then fix that when we re-integrate your golden tests and resolve the issues one at a time. Not sure how you want to approach changing the behaviour when current behaviour doesn't match the Ruby/shopify behaviour.
> 
> Fixing mathematics in DotLiquid was one of my earliest contribution to the code base, and got me in trouble because 'bug fixing' meant breaking consumers of the library who took a dependency on the current output. Hence the later introduction of compatibility versioning in the project. IMHO, this is one such case that needs to go behind a version flag.

Is there a document somewhere stating what each version flag changes/affects?

RodneyRichardson:
> > > Is this bug exclusive to Abs method? I noted several methods with the same issue in #566
> > 
> > 
> > It also shows up in at_least and at_most, but they don't return the expected behaviour when fixed! I think I will fix this bug on them, and add it to this PR, and add tests for the current incorrect behaviour. We can then fix that when we re-integrate your golden tests and resolve the issues one at a time. Not sure how you want to approach changing the behaviour when current behaviour doesn't match the Ruby/shopify behaviour.
> 
> Fixing mathematics in DotLiquid was one of my earliest contribution to the code base, and got me in trouble because 'bug fixing' meant breaking consumers of the library who took a dependency on the current output. Hence the later introduction of compatibility versioning in the project. IMHO, this is one such case that needs to go behind a version flag.

I've added Abs, Ceil and Floor behind a SyntaxCompatibilty flag, and added appropriate tests. It's a lot of work to maintain backwards compatibility! I will have a look at doing the same for the other filters.

src/DotLiquid/StandardFilters.cs
RodneyRichardson: I've added Abs, Ceil and Floor behind a SyntaxCompatibilty flag, and added appropriate tests. It's a lot of work to maintain backwards compatibility! I will have a look at doing the same for the other filters.

RodneyRichardson: I've change the behaviour to try to preserve the data type, unless it's a string. If it's a string, it will use the shared coercion to number, and then change the type as determined. 

I can do a similar thing to Round. 

AtLeast and AtMost (and presumably other mathematical functions with 2 arguments) - how do we want to handle these when the inputs are different types? Do we have a preferred order of coercion (e.g. int32, int64, decimal, double)?

daviburg: For all the mathematical functions, when the input types are different, the type with the most range and precision that covers the range and precision of the other type is preferred. And possibly in case of over/under-flow, the type is promoted to one that can handle the range.

daviburg: Here's a double that overflows on conversion to decimal:
https://dotnetfiddle.net/TgoY1q

daviburg: Here's a double that loses precision on conversion to decimal:
https://dotnetfiddle.net/99Gntw

RodneyRichardson: Unfortunately, the large number can't be parsed from a string (although I found 1e203 to work). And the precise number happily gets converted to a decimal without loss of precision. Interestingly: The precision loss only looks to be an issue with .NET Core, and not .NET Framework.

I've added all the SyntaxCompatibility flags to maintain old behaviour, and added appropriate tests. I've made AtLeast and AtMost return decimal, which allows all the golden liquid tests to pass. If you think it's worth maintaining the type in these two, it might be better to raise an issue so this can be merged.

src/DotLiquid/SyntaxCompatibilityEnum.cs
```diff
@@ -24,5 +24,15 @@ public enum SyntaxCompatibility
         /// Behavior as of DotLiquid 2.2a
         /// </summary>
         DotLiquid22a = 221,
+
+        /// <summary>
+        /// Behavior as of DotLiquid 2.2b
+        /// </summary>
+        DotLiquid22b = 222,
```
microalps, line 31: Current version of DotLiquid is 2.3. I would support just calling it DotLiquid24 = 240 and before merging bumping the version flag on the build system.

src/DotLiquid/StandardFilters.cs
```diff
@@ -805,10 +846,23 @@ public static object Ceil(Context context, object input)
         /// <returns>The rounded value; null if an exception have occurred</returns>
         public static object Floor(Context context, object input)
         {
-            if (decimal.TryParse(input.ToString(), NumberStyles.Any, context.CurrentCulture, out decimal d))
-                return Math.Floor(d);
-            else
-                return null;
+            if (context.SyntaxCompatibilityLevel < SyntaxCompatibility.DotLiquid22b)
```
microalps, line 849: I wonder if we should make a new attribute on the method specifying the Liquid compatibility, such as:

```c#
public static object Floor(Context context, object input)
        {
        ...
        }

[LiquidCompatibility(name:="Floor",minVersion=SyntaxCompatibility.DotLiquid24]
public static object FloorV24(Context context, object input)
        {
        ...
        }
```
This would then be utilized by Strainer.Create to ignore the incompatible ones. Caveat: Param tag would be unable to change the compatibility in middle of template.

src/DotLiquid/Util/StringExtensionMethods.cs
```diff
@@ -0,0 +1,81 @@
...
+using System.Text.RegularExpressions;
+
+namespace DotLiquid.Util
+{
+    public static class StringExtensionMethods
+    {
+        private static readonly Regex IntegerRegex = R.C(R.Q(@"^([+-]?\d+)$"));
+        private static readonly Regex NumericRegex = R.C(R.Q(@"^([+-]?\d[\d\.|\,]+)$"));
+
+        public static bool TryParseToNumericType(this string value, IFormatProvider formatProvider, out object convertedValue)
```
microalps, line 15: I like the idea, but maybe we should have INumeric and expose Math.Floor etc. on the INumeric interface. This would avoid need of type-casting.

src/DotLiquid/StandardFilters.cs
microalps: Give me a day or two to mock up how this would be. I'm wary to add more conditionals and want to test the performance impact.

daviburg: Oooh, very interesting idea. I like that we wouldn't have a bunch of if conditions in the middle of a method implementing a Liquid filter. Much better code readability, and easier to avoid mistakes regressing earlier compatibility implementations.

microalps: See #577 @daviburg , it does cleanup DotLiquid base, but I think it makes DotLiquid.Tests messier as direct method calls do not trigger this logic.

src/DotLiquid/SyntaxCompatibilityEnum.cs
RodneyRichardson: I'll change this to DotLiquid24 (which will be the same as the one on the render branch).

src/DotLiquid/Util/StringExtensionMethods.cs
RodneyRichardson: Sounds like another feature request! :)

src/DotLiquid/StandardFilters.cs
RodneyRichardson: I like this idea too. Makes it much more self-documenting. 

Tests could do something similar to Strainer to return the version of a function matching a compatibility level, and you could annotate the tests in a similar way, and match them up! You'd need separate test functions for the specific versions of functions where things change, but that will also make it more obvious when syntax has changed. I find reading tests is a great way to understand how to use a code base. 

I don't suppose this extends to non-filter changes in syntax (e.g. there are checks in Context and Tokenizer). But it's an improvement.

src/DotLiquid/SyntaxCompatibilityEnum.cs
RodneyRichardson: Done.

RodneyRichardson:
I consider this ready for review/merge.

RodneyRichardson:
This will have conflicts with #555

src/DotLiquid.Tests/Util/StringExtensionMethodsTests.cs
```diff
@@ -0,0 +1,107 @@
...
+
+        static IEnumerable ErrorTestCaseSource()
+        {
+            IFormatProvider invariantFormatProvider = CultureInfo.InvariantCulture;
+
+            yield return new object[] { null, null };
+            yield return new object[] { null, invariantFormatProvider };
+            yield return new object[] { string.Empty, invariantFormatProvider };
+            yield return new object[] { "banana", invariantFormatProvider };
+            // yield return new object[] { "-12,567", invariantFormatProvider }; // This returns -12567m
```
RodneyRichardson, line 104: I think we have a bug here when numbers contain a "thousands separator", but I don't want to change it yet from existing behaviour.

src/DotLiquid/StandardFilters.cs
```diff
@@ -706,49 +706,73 @@ public static object Minus(Context context, object input, object operand)
...
+            if (input is string inputString)
+            {
+                input = inputString.CoerceToNumericType(context.FormatProvider, 0);
+            }
+
+            if (input is decimal inputDecimal) { return Math.Ceiling(inputDecimal); }
+            else if (input is double inputDouble) { return Math.Ceiling(inputDouble); }
+            else if (input is int inputInt32) { return inputInt32; }
+            else if (input is long inputInt64) { return inputInt64; }
+            else return 0;
```
microalps: what happens to uint or the other types defined as IsNumeric?

RodneyRichardson: Good point. I am only handling the types that can be returned from CoerceToNumeric, but obviously they could pass in something else.

RodneyRichardson:
I'm still not completely happy with Round/AtLeast/AtMost. I'm coverting to decimal, so there's a small chance that a decimal may not contain the number(s). I'll rework it tomorrow so that it falls back to double.

src/DotLiquid.Tests/Filters/StandardFiltersV20Tests.cs
```diff
@@ -100,5 +97,195 @@ public void TestReplaceFirstRegexWorks()
...
+            Helper.AssertTemplateResult("1.235", "{{ 1.234678 | round: 2.7 }}", syntax: SyntaxCompatibilityLevel);
+            Helper.AssertTemplateResult("1.235", "{{ 1.234678 | round: 3.1 }}", syntax: SyntaxCompatibilityLevel);
+
+            Helper.AssertTemplateResult("", "{{ 1.234678 | round: -3 }}", syntax: SyntaxCompatibilityLevel);
+        }
+
+        [Test]
+        public void TestAtLeastTypes()
+        {
+            Assert.That(AtLeast(5, 5), Is.EqualTo(5).And.TypeOf(typeof(double)));
```
daviburg: Why are simple integer inputs resulting into a floating type (double)?

src/DotLiquid.Tests/Util/NumericConverterTests.cs
```diff
@@ -0,0 +1,177 @@
...
+            Assert.That(15.0.CoerceToReal(CultureInfo.InvariantCulture, 0), Is.EqualTo(15).And.TypeOf(typeof(double)));
+            Assert.That("15".CoerceToReal(CultureInfo.InvariantCulture, 0), Is.EqualTo(15m).And.TypeOf(typeof(decimal)));
+            Assert.That("-15".CoerceToReal(CultureInfo.InvariantCulture, 0), Is.EqualTo(-15m).And.TypeOf(typeof(decimal)));
+            Assert.That("15.0".CoerceToReal(CultureInfo.InvariantCulture, 0), Is.EqualTo(15m).And.TypeOf(typeof(decimal)));
+        }
+
+        [Test]
+        public void TestCoerceToRealOverflow()
+        {
+            double largeValue = double.Parse("1e203");
```
daviburg: consider using `var` for obvious types, e.g.

```suggestion
            var largeValue = double.Parse("1e203");
```

src/DotLiquid/Util/NumericConverter.cs
```diff
@@ -0,0 +1,165 @@
...
+using System.Linq.Expressions;
+using System.Reflection;
+using System.Text.RegularExpressions;
+
+namespace DotLiquid.Util
+{
+    /// <summary>
+    /// ExtensionMethods for converting to numeric values.
+    /// </summary>
+    public static class NumericConverter
```
daviburg, line 12: The typical naming convention for a class containing extensions is `<Extended class>Extensions`. This makes the extension methods easier to discover because if you see some code calling an extension method on say a string, you know you will find the implementation in the file StringExtensions.cs

src/DotLiquid/Util/NumericConverter.cs
```diff
@@ -0,0 +1,165 @@
...
+                    result = Convert.ToDouble(convertedFloat);
+                }
+                else
+                {
+                    // https://docs.microsoft.com/en-us/dotnet/standard/base-types/conversion-tables
+                    // All integer types can be widened to decimal
+                    result = Convert.ToDecimal(convertedObject);
+                }
+            }
+            return result;
```
daviburg:
```suggestion

            return result;
```

src/DotLiquid/Util/NumericConverter.cs
```diff
@@ -0,0 +1,165 @@
...
+        /// <returns>The coerced value as decimal or double type, or <paramref name="defaultValue"/> if coercion fails.</returns>
+        public static object CoerceToReal(this object value, IFormatProvider formatProvider, decimal defaultValue)
+        {
+            object result = defaultValue;
+            if (value != null)
+            {
+                object convertedObject = value.CoerceToNumericType(formatProvider, defaultValue);
+                if (convertedObject is decimal convertedDecimal)
+                {
+                    result = convertedDecimal;
```
daviburg:
```suggestion
                    return convertedDecimal;
```
I find it easier to read when one returns right away instead of having to scroll through all the subsequent blocks to see that all what happens is the return of the assigned value. This would also allow writing immutable code.

src/DotLiquid/Util/NumericConverter.cs
```diff
@@ -0,0 +1,165 @@
...
+        /// <summary>
+        /// Coerce an object into a decimal or double value
+        /// </summary>
+        /// <param name="value">The value to coerce. This can be a numeric type, or a string representation of a numeric type.</param>
+        /// <param name="formatProvider">The format provider for converting floating point numbers.</param>
+        /// <param name="defaultValue">The value to return if coercion fails.</param>
+        /// <returns>The coerced value as decimal or double type, or <paramref name="defaultValue"/> if coercion fails.</returns>
+        public static object CoerceToReal(this object value, IFormatProvider formatProvider, decimal defaultValue)
+        {
+            object result = defaultValue;
```
daviburg: consider just returning the default value directly in the fall back cases so that the code is immutable.

src/DotLiquid/Util/NumericConverter.cs
```diff
@@ -0,0 +1,165 @@
...
+                {
+                    result = convertedDecimal;
+                }
+                else if (convertedObject is double convertedDouble)
+                {
+                    result = convertedDouble;
+                }
+                else if (convertedObject is float convertedFloat)
+                {
+                    // https://docs.microsoft.com/en-us/dotnet/standard/base-types/conversion-tables
```
daviburg: ??? Float isn't in that table of lossles conversion, is it?

src/DotLiquid/Util/NumericConverter.cs
```diff
@@ -0,0 +1,165 @@
...
+                    // Float be widened to double
```
daviburg: Why do that?

src/DotLiquid/Util/NumericConverter.cs
```diff
@@ -0,0 +1,165 @@
...
+                    result = Convert.ToDouble(convertedFloat);
+                }
+                else
+                {
+                    // https://docs.microsoft.com/en-us/dotnet/standard/base-types/conversion-tables
+                    // All integer types can be widened to decimal
```
daviburg: That they can, but should we endure the compute overhead of systematically promoting everything to Decimal whether they need it or not?

src/DotLiquid/Util/NumericConverter.cs
```diff
@@ -0,0 +1,165 @@
...
+        /// <param name="value">The string to coerce.</param>
+        /// <param name="formatProvider">The format provider for converting floating point numbers.</param>
+        /// <param name="defaultValue">The value to return if coercion fails.</param>
+        /// <returns>The coerced value as int, long, double or decimal type, or <paramref name="defaultValue"/> if coercion fails.</returns>
+        public static object CoerceToNumericType(this object value, IFormatProvider formatProvider, object defaultValue)
+        {
+            object result = defaultValue;
+            if (value != null)
+            {
+                bool converted = value.TryCoerceToNumericType(formatProvider, out object convertedValue);
```
daviburg: don't declare single-use variables. Simply inline the code.

src/DotLiquid/Util/NumericConverter.cs
```diff
@@ -0,0 +1,165 @@
...
+        /// </summary>
+        /// <param name="value">The string to parse.</param>
+        /// <param name="formatProvider">The format provider for converting floating point numbers.</param>
+        /// <param name="convertedValue">The coerced value as int, long, double or decimal type, or null if parsing fails.</param>
+        /// <returns>true if parsing was successful; Otherwise, false.</returns>
+        public static bool TryCoerceToNumericType(this object value, IFormatProvider formatProvider, out object convertedValue)
+        {
+            if (value != null)
+            {
+                if (IsNumeric(value))
```
daviburg:
```suggestion
                if (NumericConverter.IsNumeric(value))
```
Qualify static method calls with the class name so it is easier to distinguish between instance method calls (this.FooBar) and static method calls (Class.FooBar) in a systematic manner.

src/DotLiquid/Util/NumericConverter.cs
```diff
@@ -0,0 +1,165 @@
...
+
+namespace DotLiquid.Util
+{
+    /// <summary>
+    /// ExtensionMethods for converting to numeric values.
+    /// </summary>
+    public static class NumericConverter
+    {
+        private static bool IsReal(object o) => o is double || o is float || o is decimal;
+        private static bool IsInteger(object o) => o is int || o is uint || o is long || o is ulong || o is short || o is ushort || o is byte || o is sbyte;
```
daviburg: Shouldn't you have BigInteger here also?

src/DotLiquid/Util/NumericConverter.cs
RodneyRichardson: No, you're right - I think I can rewrite AtLeast and AtMost to use a BinaryExpression (like Plus), and change Round to be similar. I will need to check string conversion works as we want.

src/DotLiquid/Util/NumericConverter.cs
RodneyRichardson: I don't think we support BigInteger anywhere.

src/DotLiquid/Util/NumericConverter.cs
RodneyRichardson: It is the last in the widening table: Single to Double.

src/DotLiquid/Util/NumericConverter.cs
RodneyRichardson: It waa for simplicity in AtLeast/AtMost so I only had to deal with two datatypes. I will use the tyoe promotion code instead.

src/DotLiquid/Util/NumericConverter.cs
RodneyRichardson: Whereas my rule of thumb is to have a single return point when possible! 🤔

src/DotLiquid/Util/NumericConverter.cs
RodneyRichardson: This started as StringExtensions, and was renamed when I added objects. It shoukd probably just not use the Extension syntactic sugar.

src/DotLiquid.Tests/Filters/StandardFiltersV20Tests.cs
RodneyRichardson: This should be fixed now.

src/DotLiquid.Tests/Util/NumericConverterTests.cs
RodneyRichardson: Deleted.

src/DotLiquid/Util/NumericConverter.cs
RodneyRichardson: Deleted.

src/DotLiquid/Util/NumericConverter.cs
RodneyRichardson: The code base is pretty inconsistent with this, but I've tried to update this patch.

RodneyRichardson:
What started as a simple "check for null" on `Abs` has now evolved quite a bit!

I think this is ready, again. I've moved all the meth operations to use updated logic to convert strings to numbers, and to try to retain the data types. There was also a bug where integer overflow threw an exception rather than handling it like floating point. 

AtMost/AtLeast should now retain the data type passed in, if possible. 

There are a few lines I've not been able to get coverage of - there are a few default switches that are essentially unreachable code, but I think should stay in. And I couldn't get doubles to throw an OverflowException - seems they tend to return PositiveInfinity instead.

src/DotLiquid/StandardFilters.cs
```diff
@@ -839,21 +899,40 @@ public static double Abs(Context context, object input)
...
+                    return Math.Max(Convert.ToDecimal(val1), Convert.ToDecimal(val2));
+                case Type t when t == typeof(double):
+                    return Math.Max(Convert.ToDouble(val1), Convert.ToDouble(val2));
+                case Type t when t == typeof(float):
+                    return Math.Max(Convert.ToSingle(val1), Convert.ToSingle(val2));
+                case Type t when t == typeof(ulong):
+                    return Math.Max(Convert.ToUInt64(val1), Convert.ToUInt64(val2));
+                case Type t when t == typeof(long):
+                    return Math.Max(Convert.ToInt64(val1), Convert.ToInt64(val2));
+                case Type t when t == typeof(uint):
```
microalps: Besides being very verbose, this introduces a new bug. If I call `1 | at_least 2147483648`, I should get a long. Hence the suggestion to use type promotions to get the greater type of the two. Same goes if I have `1 | at_least 2.5`. Here is a simpler suggestion:

```
	var decMax = Math.Max(decInput, decAtLeast);
	if (decMax == decInput)
		return input;
	return atLeast;
```

See DoMathsOperation for case that one is double and can't be converted with OverflowException
[POCAtLeastNumerics.zip](https://github.com/user-attachments/files/19597312/POCAtLeastNumerics.zip)

src/DotLiquid/StandardFilters.cs
```diff
@@ -827,10 +872,25 @@ public static IEnumerable Uniq(object input)
...
+
+            if (input is string inputString)
+            {
+                input = NumericConverter.CoerceToNumericType(inputString, context.FormatProvider, 0);
+            }
+
+            if (input is decimal inputDecimal) { return Math.Abs(inputDecimal); }
+            else if (input is float inputFloat) { return Math.Abs(inputFloat); }
+            else if (input is double inputDouble) { return Math.Abs(inputDouble); }
+            else if (input is sbyte inputInt8) { return Math.Abs(inputInt8); }
```
microalps: C# is unable to handle sbyte.MinValue as the result exceeds the MaxValue. Same for all the remaining integer types.

```c#
public static object Abs(object input) {
	try 
	{
		var decInput = Convert.ToDecimal(input);
		var decResult = Math.Abs(decInput);
		
		try {
			return Convert.ChangeType(decResult, input.GetType());
		} catch (OverflowException) 
		{
			return Convert.ChangeType(decResult, NumericTypePromotions[input.GetType()][0]);
		}
	} 
	catch (OverflowException) 
	{
		var dblInput = Convert.ToDouble(input);
		var dblResult = Math.Abs(dblInput);
		return Convert.ChangeType(dblResult, input.GetType());
	}
}
```

Attached updated 
[POCAtLeastNumerics.zip](https://github.com/user-attachments/files/19597783/POCAtLeastNumerics.zip)

microalps:
Based on my proof of concept code, I've tackled [rewriting](https://github.com/microalps/dotliquid/commit/362a97095dfb6d34ad79cb6f79370c373bc9836b) some math functions. This is not ready for review, but to serve as inspiration. However, I think the next step is to detail ALL of the issues with current code regarding math, not just the linked ticket.

src/DotLiquid/StandardFilters.cs
RodneyRichardson: `1 | atleast 2147483648` does result in `long` - the call to `GetBinaryResultType` does the type promotion (as per `DoMathsOperation`). 

But, `2147483648` is actually a `uint` so it should probably return a `uint`. We can simplify this with using dynamic types.

RodneyRichardson:
> Based on my proof of concept code, I've tackled [rewriting](https://github.com/microalps/dotliquid/commit/362a97095dfb6d34ad79cb6f79370c373bc9836b) some math functions. This is not ready for review, but to serve as inspiration. However, I think the next step is to detail ALL of the issues with current code regarding math, not just the linked ticket.

There are definitely more issues than just the listed ones. I've fixed a few related to combinations of types and MinValue/overflows, but we probably need a more robust set of tests to check all the edge conditions. Maybe something for v3?

src/DotLiquid/StandardFilters.cs
```diff
@@ -667,55 +764,89 @@ public static object Minus(Context context, object input, object operand)
...
-                return Math.Round(i, p);
-            }
-            catch (Exception)
-            {
-                return null;
-            }
+            // Math.Round can handle at most 28 decimals, so clamp place into [0, 28].
+            const int MaxDecimalPlaces = 28;
+            int decimals = Convert.ToInt32(
+                StandardFilters.Floor(context,
```
microalps: Seems a lot of method calls for naught. AI says Clamp is `(value < min) ? min : (value > max) ? max : value;` where value is Convert.ToInt32(places)

src/DotLiquid/LegacyFilters.cs
microalps, line 1: I dislike comparisons where changes are just spaces. It's impossible to find changes. Likewise for changes in position for code cleanliness, while it's ideal it's not a good idea in a random PR.

src/DotLiquid/StandardFilters.cs
```diff
@@ -20,6 +20,102 @@ namespace DotLiquid
...
+                    // Retry as Doubles (if they weren't already)
+                    if (input is double && operand is double)
+                    {
+                        throw;
+                    }
+
+                    input = Convert.ToDouble(input);
+                    operand = Convert.ToDouble(operand);
+
+                    return ExpressionUtility
```
microalps: seems like this method can be better restated with 3x ExpressionUtility reference.

src/DotLiquid/StandardFilters.cs
```diff
@@ -644,10 +740,10 @@ public static object Last(IEnumerable array)
         /// <param name="context">The DotLiquid context</param>
         /// <param name="input">Input to be transformed by this filter</param>
         /// <param name="operand">Number to be added to input</param>
-        [LiquidFilter(MinVersion = SyntaxCompatibility.DotLiquid21)]
+        [LiquidFilter(MinVersion = SyntaxCompatibility.DotLiquid24)]
         public static object Plus(Context context, object input, object operand)
         {
-            return DoMathsOperation(context, input, operand, Expression.AddChecked);
+            return StandardFilters.DoMathsOperation(context, input, operand, Expression.AddChecked);
```
microalps, line 650: seems like an unnecessary change since this is in the same class. I know it differentiates from LegacyFilters, but I think that's already true and triggers IDE0002 Name can be simplified.

src/DotLiquid/StandardFilters.cs
```diff
@@ -827,10 +913,27 @@ public static IEnumerable Uniq(object input)
...
-            Double n;
-            return Double.TryParse(input.ToString(), NumberStyles.Number, context.CurrentCulture, out n) ? Math.Abs(n) : 0;
+            dynamic value = NumericConverter.CoerceToNumericType(input, context.FormatProvider, 0);
+            if (value >= 0)
+            {
+                return value;
+            }
+
+            // Try multiplying my -1 to get the absolute value.
+            dynamic minusOne = Convert.ChangeType(-1, value.GetType());
```
microalps: I like the new method as it avoids type casting back and forth, but why not
```suggestion
            int minusOne = -1;
```
per MS you can combine math ops with int + float just not two differing types of float...

src/DotLiquid/Util/NumericConverter.cs
```diff
@@ -0,0 +1,190 @@
...
+        /// </summary>
+        /// <param name="value">The string to parse.</param>
+        /// <param name="formatProvider">The format provider for converting floating point numbers.</param>
+        /// <param name="convertedValue">The coerced value as int, long, double or decimal type, or null if parsing fails.</param>
+        /// <returns>true if parsing was successful; Otherwise, false.</returns>
+        public static bool TryParseToNumericType(string value, IFormatProvider formatProvider, out object convertedValue)
+        {
+            if (value != null)
+            {
+                if (int.TryParse(value, NumberStyles.Integer | NumberStyles.AllowThousands, formatProvider, out int intValue))
```
microalps, line 143: This is a change from the Regex in that thousands is now allowed while previously it was not. An improvement, certainly, but should be documented.

src/DotLiquid.Tests/Util/NumericConverterTests.cs
```diff
@@ -0,0 +1,189 @@
...
+            yield return new object[] { "123", invariantFormatProvider, 123 };
+            yield return new object[] { "-123", invariantFormatProvider, -123 };
+
+            // Int32 with thousands separator
+            yield return new object[] { String.Format(invariantFormatProvider, "{0:#,##0}", 12567),
+                invariantFormatProvider, 12567 };
+            yield return new object[] { String.Format(frenchFormatProvider, "{0:#,##0}", 12567),
+                frenchFormatProvider, 12567 };
+            // Note: For fallback to happen, the number must be big enough to contain 2 separators,
+            // otherwise it will be interprested as a French floating point number.
```
microalps: nit: spelling of interpreted

src/DotLiquid.Tests/Util/NumericConverterTests.cs
```diff
@@ -0,0 +1,189 @@
...
+
+            // Int32
+            yield return new object[] { "0", null, 0 };
+            yield return new object[] { "0", invariantFormatProvider, 0 };
+            yield return new object[] { "123", invariantFormatProvider, 123 };
+            yield return new object[] { "-123", invariantFormatProvider, -123 };
+
+            // Int32 with thousands separator
+            yield return new object[] { String.Format(invariantFormatProvider, "{0:#,##0}", 12567),
+                invariantFormatProvider, 12567 };
```
microalps: nit: @daviburg would say either place each parameter on its own line or one line. commas sometimes in middle do not make for easy readability.

src/DotLiquid.Tests/Util/NumericConverterTests.cs
```diff
@@ -0,0 +1,189 @@
...
+
+        private static IEnumerable ErrorTestCaseSource()
+        {
+            IFormatProvider invariantFormatProvider = CultureInfo.InvariantCulture;
+
+            yield return new object[] { null, null };
+            yield return new object[] { null, invariantFormatProvider };
+            yield return new object[] { string.Empty, invariantFormatProvider };
+            yield return new object[] { "banana", invariantFormatProvider };
+        }
```
microalps, line 240: what about 1.023.456 or 1.023,456 (in invariant)?

src/DotLiquid/StandardFilters.cs
RodneyRichardson: `Convert.ToInt32()` isn't gauranteed to work for all inputs, and neither are comparison operators if we tried to use dynamic, unfortunately. We could try using `Convert.ToDouble`, which should work, but could potentially give the wrong value if we lose precision in the conversion. 

This way does exactly the same, but defers sorting out types to the other methods. I reckon we may as well make use of the edge case handling we've already got than try to reproduce it here.

src/DotLiquid/LegacyFilters.cs
RodneyRichardson: I know - it's really annoying. I think it's the line endings that have changed, making comparisons fail. I'm not sure how to undo these, but I'll have a look.

src/DotLiquid/StandardFilters.cs
RodneyRichardson: Ah - @daviburg's review said exactly the opposite (hence the change). We need to decide a coding standard. I quite like being explicit, personally.

src/DotLiquid/Util/NumericConverter.cs
RodneyRichardson: Added `<remarks>` to XML documentation.

src/DotLiquid/StandardFilters.cs
RodneyRichardson: I'm not sure what you're suggesting here. It's not a very clean method, but it needs to use decimal for precision, and retry Overflow or DivideByZero exceptions - sometimes two retries.  The legacy version only has 2 calls, but has a bug where it doesn't handle integer operation overflow in all cases.

src/DotLiquid/LegacyFilters.cs
RodneyRichardson: @microalps You can tell github to ignore whitespace changes by clicking the little settings icon, and shoosing "Hide Whitespace"
image

src/DotLiquid/StandardFilters.cs
RodneyRichardson: I've taken out a redundant `else` statement, but I think we still need 3 calls.

RodneyRichardson:
Ready for another review. I've addressed most of your comments - simplified some functions, and fixed a few small edge case bugs. I've left the class name on calls to static functions, as per David's request, and I haven't been able to reduce DoMathsOperation further.

src/DotLiquid.Tests/Util/ExpressionUtilityTests.cs
```diff
@@ -0,0 +1,72 @@
...
+
+namespace DotLiquid.Tests.Util
+{
+    [TestFixture]
+    public class ExpressionUtilityTests
+    {
+        [Test]
+        public void CreateExpression_AddsTwoIntegers()
+        {
+            var del = ExpressionUtility.CreateExpression(Expression.AddChecked, typeof(int), typeof(int));
```
daviburg: del as in delegate? consider explicit variable name _delegate that wouldn't conflict with the reserved keyword delegate.

src/DotLiquid/Util/NumericConverter.cs
```diff
@@ -0,0 +1,190 @@
...
+        /// Check if the object is a numeric type.
+        /// </summary>
+        /// <param name="value">The value to check</param>
+        /// <returns>True if the value is an integer or floating point number; false otherwise.</returns>
+        public static bool IsNumeric(object value) => IsReal(value) || IsInteger(value);
+
+        /// <summary>
+        /// A dictionary of numeric types and their allowable conversions.
+        /// Based on the promotion table at
+        /// https://docs.microsoft.com/en-us/dotnet/standard/base-types/conversion-tables
```
daviburg:
```suggestion
        /// https://docs.microsoft.com/dotnet/standard/base-types/conversion-tables

```
By removing the locale URL segment, the user will get whatever locale they configured for their browsing experience.

src/DotLiquid/Util/NumericConverter.cs
```diff
@@ -0,0 +1,190 @@
...
+            { typeof(Single), new Type[] { typeof(Double) } },
+            { typeof(Decimal), new Type[] { typeof(Single), typeof(Double) } },
+            { typeof(Double), new Type[] { } },
+        });
+
+        /// <summary>
+        /// Get the return type for an operation between the two numbers of the specified types
+        /// </summary>
+        /// <param name="left">Type of the left parameter</param>
+        /// <param name="right">Type of the left parameter</param>
```
daviburg:
```suggestion
        /// <param name="right">Type of the right parameter</param>

```

src/DotLiquid/Util/NumericConverter.cs
```diff
@@ -0,0 +1,190 @@
...
+        /// <returns>The return type</returns>
+        /// <exception cref="ArgumentException">Thrown if a parameter is not a supported numeric type.</exception>
+        public static Type GetBinaryResultType(Type left, Type right)
+        {
+            if (left == right)
+                return left;
+
+            if (!NumericConverter.NumericTypePromotions.TryGetValue(left, out Type[] leftTypes))
+                throw new System.ArgumentException("Argument is not numeric", nameof(left));
```
daviburg:
```suggestion
                throw new ArgumentException(message: "Argument is not numeric", paramName: nameof(left));
```

As there is already a `using System;` above?

Also adding the named parameters as both are of the same type, hence the order is not obvious to a human reader.

src/DotLiquid/Util/NumericConverter.cs
```diff
@@ -0,0 +1,190 @@
...
+            if (!NumericConverter.NumericTypePromotions.TryGetValue(right, out Type[] rightTypes))
+                throw new System.ArgumentException("Argument is not numeric", nameof(right));
+
+            // Test left to right promotion
+            if (rightTypes.Contains(left))
+                return left;
+            if (leftTypes.Contains(right))
+                return right;
+            return rightTypes.First(t => leftTypes.Contains(t));
```
daviburg:
```suggestion
            return rightTypes.First(_type => leftTypes.Contains(_type));

```
Using English terms to disambiguate and for readability, with _ prefix to avoid clash with language reserved keyword.

src/DotLiquid.Tests/Util/ExpressionUtilityTests.cs
RodneyRichardson: I was just testing out how good Github Copilot was at creating unit tests, and didn't think to change it's naming convention. As it turns out, Copilot isn't very good at creating tests! I had to rewrite most of them, and they weren't specifically designed around coverage or testing edge cases.

src/DotLiquid/StandardFilters.cs
```diff
@@ -922,6 +1044,44 @@ public static IEnumerable Where(IEnumerable input, string propertyName, object t
...
+            // Flatten values to handle nested arrays
+            object sum = 0;
+            var values = input.Flatten().Cast<object>();
+            foreach (object value in values)
+            {
+                if (value != null)
+                {
+                    dynamic valueToAdd = NumericConverter.CoerceToNumericType(value, context.FormatProvider, 0);
+                    if (valueToAdd != 0)
+                    {
```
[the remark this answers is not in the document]
RodneyRichardson: No. We explicitly want to keep the type that is supplied, when possible.

src/DotLiquid/Util/ExpressionUtility.cs
```diff
@@ -11,60 +11,10 @@ namespace DotLiquid.Util
...
-            if (NumericTypePromotions[left].Contains(right))
-                return right;
-            return NumericTypePromotions[right].First(p => NumericTypePromotions[left].Contains(p));
-        }
-
         private static void Cast(Expression lhs, Expression rhs, Type leftType, Type rightType, Type resultType, out Expression castLhs, out Expression castRhs)
         {
-            castLhs = leftType == resultType ? lhs : (Expression)Expression.Convert(lhs, resultType);
-            castRhs = rightType == resultType ? rhs : (Expression)Expression.Convert(rhs, resultType);
+            castLhs = leftType == resultType ? lhs : Expression.Convert(lhs, resultType);
```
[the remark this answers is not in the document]
RodneyRichardson: Added doc comment.

src/DotLiquid/Context.cs
```diff
@@ -443,39 +441,9 @@ private object Resolve(string key, bool notifyNotFound = true)
...
-                            if (decimal.TryParse(match.Groups[1].Value, NumberStyles.Number | NumberStyles.Float, FormatProvider, out decimal parsedDecimalCurrentCulture))
-                                return parsedDecimalCurrentCulture;
-                            if (decimal.TryParse(match.Groups[1].Value, NumberStyles.Number | NumberStyles.Float, CultureInfo.InvariantCulture, out decimal parsedDecimalInvariantCulture))
-                                return parsedDecimalInvariantCulture;
-                            if (double.TryParse(match.Groups[1].Value, NumberStyles.Number | NumberStyles.Float, FormatProvider, out double parsedDouble))
-                                return parsedDouble;
-                            return double.Parse(match.Groups[1].Value, NumberStyles.Number | NumberStyles.Float, CultureInfo.InvariantCulture);
-                        }
+                        // Numeric values.
+                        if (NumericConverter.TryParseToNumericType(key, FormatProvider, out object numericValue))
```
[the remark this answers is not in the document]
RodneyRichardson: No. I don't think comments should capture change history.

src/DotLiquid.Tests/Util/ExpressionUtilityTests.cs
daviburg: Ugh. Do you have the "report a problem" button in the UX where you attempted to use Github Copilot? To let the corresponding team know they need to improve the model.

RodneyRichardson: Ah - there's a "Was this helpful?" with a thumbs up/down. It tend not to use that as it seems a bit pointless, but maybe I will in future. That seems to redirect to "Report a problem" if I do a thumbs down.
image

jimcullenaus:
Is this still in progress? Would love to be able to take advantage of the `sum` filter. Especially with the "if a property is not found then it's treated as zero" from #590.

daviburg:
> Is this still in progress? Would love to be able to take advantage of the `sum` filter. Especially with the "if a property is not found then it's treated as zero" from #590.

Sorry, looks like we lost track.

cyrildurand:
Any update on this ? How can we help ? 
I'm interested in the `sum` filter and would love to have it.

daviburg (commented):
[Dobby] AI-generated review from blind independent GPT-5.6 Sol and Claude Opus 5 passes, followed by cross-model validation. Five medium-impact, high-confidence findings are annotated inline.

src/DotLiquid/StandardFilters.cs
```diff
@@ -667,55 +668,78 @@ public static object Minus(Context context, object input, object operand)
...
+        /// <param name="context">The DotLiquid context</param>
         /// <param name="input">Input to be transformed by this filter</param>
         /// <param name="places">Number of decimal places for rounding</param>
-        /// <returns>The rounded value; null if an exception have occurred</returns>
-        public static object Round(object input, object places = null)
+        /// <returns>The rounded value; zero if input is invalid, or rounded to 0 decimals if places is invalid</returns>
+        /// <remarks>Behaviour differs from Ruby implementation for negative places values.
+        /// This will treat it as any other invalid places value, and round to closest integer.</remarks>
+        [LiquidFilter(MinVersion = SyntaxCompatibility.DotLiquid24)]
+        public static object Round(Context context, object input, object places = null)
```
daviburg, line 687: [Dobby] [AI-generated review: GPT-5.6 Sol + Claude Opus 5; blind independent passes with cross-model validation]

**Medium impact, high confidence:** this changes two existing public CLR APIs independently of the Liquid compatibility routing. `StandardFilters.Round(object, object)` is removed in favor of `Round(Context, object, object)`, and the sibling `StandardFilters.Abs(Context, object)` changes its return type from `double` to `object`. Existing source callers can stop compiling, and precompiled callers can fail with `MissingMethodException`; `SyntaxCompatibility` does not protect CLR consumers. Could we preserve the old public signatures and register the V24 implementations under different CLR method names via `LiquidFilter(Name = ...)`, or explicitly treat this as a major-version API break with an API compatibility check and changelog entry?

src/DotLiquid/StandardFilters.cs
```diff
@@ -667,55 +668,78 @@ public static object Minus(Context context, object input, object operand)
...
         {
-            try
-            {
-                var p = places == null ? 0 : Convert.ToInt32(places);
-                var i = Convert.ToDecimal(input);
-                return Math.Round(i, p);
-            }
-            catch (Exception)
+            // Math.Round can handle at most 28 decimals, so clamp places into [0, 28].
```
daviburg, line 689: [Dobby] [AI-generated review: GPT-5.6 Sol + Claude Opus 5; blind independent passes with cross-model validation]

**Medium impact, high confidence:** the 28-digit limit is specific to `Math.Round(decimal, int)`. Because `inputValue` is dynamic, `double`/`float` select `Math.Round(double, int)`, whose maximum is 15. Against this exact head, `Round(context, 1.5d, 20)` throws `ArgumentOutOfRangeException`, and `{{ 1e300 | round: 20 }}` renders a Liquid error because `1e300` parses as `double`. Please clamp according to the runtime input type (and handle non-finite `places` explicitly), with boundary tests for double/float at 15 and 16 and decimal at 28 and 29.

src/DotLiquid/Context.cs
daviburg, line 445: [Dobby] [AI-generated review: GPT-5.6 Sol + Claude Opus 5; blind independent passes with cross-model validation]

**Medium impact, high confidence:** this integer-first parser is used for every syntax compatibility level, so it changes legacy arithmetic before the versioned filter dispatch runs. In the base branch, invariant `1,000` matches the numeric path and is parsed as `decimal`; here it becomes `int`. Consequently a DotLiquid20 template such as `{{ 1,000 | divided_by: 3 }}` now renders `333` rather than taking the prior decimal-division path. Please preserve the previous resolver ordering for legacy modes or gate the new resolver behavior by syntax version, and add a cross-version test that asserts both the resolved CLR type and division output for grouped literals.

src/DotLiquid.Tests/Embedded/golden_rules.json
```diff
@@ -209,22 +170,34 @@
         "liquid.golden.whitespace_control - suppress whitespace surrounding an empty capture block"
     ],
     "alternate_test_expectations": {
+        // DotLiquid allows strings, and treats invalid values as zero
+        "liquid.golden.divided_by_filter - arg string not a number": "∞",
+        "liquid.golden.divided_by_filter - divied by zero": "∞",
```
daviburg, line 175: [Dobby] [AI-generated review: GPT-5.6 Sol + Claude Opus 5; blind independent passes with cross-model validation]

**Medium impact, high confidence:** this expected string is globalization-dependent. `Variable.Render` formats the resulting `double.PositiveInfinity` with the render provider, and the built net6.0 head renders `{{ 1 | divided_by: 0 }}` with `InvariantCulture` as `Infinity`, not `∞`. Other target frameworks and globalization modes can also use different infinity symbols, so this hardcoded golden value is not portable across the supported test legs. Please derive the expectation from the chosen provider `PositiveInfinitySymbol`, or pin the harness and expectation to one explicit globalization behavior.

src/DotLiquid/StandardFilters.cs
```diff
@@ -922,6 +1044,44 @@ public static IEnumerable Where(IEnumerable input, string propertyName, object t
...
+            }
+
+            // Flatten values to handle nested arrays
+            object sum = 0;
+            var values = input.Flatten().Cast<object>();
+            foreach (object value in values)
+            {
+                if (value != null)
+                {
+                    dynamic valueToAdd = NumericConverter.CoerceToNumericType(value, context.FormatProvider, 0);
```
daviburg, line 1074: [Dobby] [AI-generated review: GPT-5.6 Sol + Claude Opus 5; blind independent passes with cross-model validation]

**Medium impact, high confidence:** every nonzero element calls `DoMathsOperation`, which reaches a fresh `Expression.Lambda(...).Compile()` and `DynamicInvoke`. A Release/net6.0 probe against this exact head showed approximately 5 MB allocated for 1,000 integers and 50 MB plus 858 ms for 10,000 integers, scaling linearly with the array size. This turns one template filter into an unbounded expression-compilation loop. Please cache compiled delegates by operation and operand types, or use a typed accumulator that avoids compilation per element, and add a focused allocation/timing or compile-count regression test.

src/DotLiquid/StandardFilters.cs
daviburg: Human review - yes, this is exactly the type of breaking change we want to avoid. We do need to preserve old signatures and avoid build breaks.

daviburg:
> Any update on this ? How can we help ? I'm interested in the `sum` filter and would love to have it.

@cyrildurand would you be able to take over this PR and address the findings I posted through the agent review?

src/DotLiquid/StandardFilters.cs
microalps: Sorry I disagree with this review. Signature of filters, while declared as public, are not intended for public use directly. Instead, they are used via liquid templates and the caller is indifferent to many signature changes. Adding/removing context as 1st parameter, or broadening the possible output from double, are both fair game. Particularly with these methods being opt-in via the compatibility flag, the fact that an int may return and stringified result not include a decimal is exactly how this library is designed. @daviburg  justify any change here please.

daviburg: oooh, dang. Thanks for catching that.

src/DotLiquid/StandardFilters.cs
microalps: This seems to be a work-item of its own.

Update: Logged #617 to track this pre-existing issue

PR #573, opened 2025-02-20
Return right-hand value in Append and Prepend when left-hand value is null

### LINKED ISSUE #572: Append outputs empty string when left-hand operand is nil
`{{ nonesuch | append: 'hi' }}` should output "hi", but outputs an empty string.

### Dotliquid version
2.3.18

### Expected behavior

It should output the right-hand operand (i.e. "hi")

### Actual behavior

Outputs an empty string.

Use this template: `{{ nonesuch | append: 'hi' }}`

RodneyRichardson:
Closes #572.

Picked up by Golden Liquid tests.


src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -1512,19 +1512,25 @@ public void TestTimesStringV21()
...
 
         [Test]
         public void TestPrepend()
         {
             Hash assigns = Hash.FromAnonymousObject(new { a = "bc", b = "a" });
-            Helper.AssertTemplateResult("abc", "{{ a | prepend: 'a'}}", assigns);
-            Helper.AssertTemplateResult("abc", "{{ a | prepend: b}}", assigns);
+            Helper.AssertTemplateResult(expected: "abc", template: "{{ a | prepend: 'a'}}", localVariables: assigns);
+            Helper.AssertTemplateResult(expected: "abc", template: "{{ a | prepend: b}}", localVariables: assigns);
+            Helper.AssertTemplateResult(expected: "bcbc", template: "{{ a | prepend: a}}", localVariables: assigns);
```
microalps, line 1531: nit: this test doesn't add any more than the previous

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -1512,19 +1512,25 @@ public void TestTimesStringV21()
         [Test]
         public void TestAppend()
         {
-            Hash assigns = Hash.FromAnonymousObject(new { a = "bc", b = "d" });
+            Hash assigns = Hash.FromAnonymousObject(new { a = "bc", b = "d", d = "x" });
```
microalps, line 1515: Not clear where d is used below. Seems like this and next are unnecessary changes and just the last two lines are valid.

microalps (changes requested):
Please clean up the tests as noted and this is OK to merge.

microalps:
Quick testing on my end indicates that this fixes 2 tests:
liquid.golden.append_filter - undefined left value
liquid.golden.prepend_filter - undefined left value

src/DotLiquid.Tests/StandardFilterTests.cs
RodneyRichardson: This is testing that the lhs and rhs can contain the same argument. I thought that was a fairly common edge case to test. I'm happy to remove it if you think it's unnecessary.

src/DotLiquid.Tests/StandardFilterTests.cs
RodneyRichardson: d was added to confirm that the parser didn't get confused between `'d'` and `d` in a template. Again, not necessary, but I thought it a useful.

RodneyRichardson:
Tests removed - should be good to go.

microalps:
@daviburg giving you an opportunity to review. As I see it, it's a bug fix and should not be with a version flag.

src/DotLiquid.Tests/StandardFilterTests.cs
microalps: Good thinking. Leave it and comment so nobody removes it later.

src/DotLiquid/StandardFilters.cs
```diff
@@ -575,7 +575,7 @@ public static string RemoveFirst(Context context, string input, string @string)
         public static string Append(string input, string @string)
         {
             return input == null
-                ? input
+                ? @string
                 : input + @string;
         }
```
daviburg, line 580: Condition becomes meaningless. If we do this change, we should simplify to 
```csharp
public static string Append(string input, string @string) => input +@string;
```

src/DotLiquid/StandardFilters.cs
```diff
@@ -587,7 +587,7 @@ public static string Append(string input, string @string)
         public static string Prepend(string input, string @string)
         {
             return input == null
-                ? input
+                ? @string
                 : @string + input;
         }
```
daviburg, line 592: Condition becomes meaningless. If we do this change, we should simplify to 
```csharp
public static string Prepend(string input, string @string) => @string + input;
```

daviburg:
> @daviburg giving you an opportunity to review. As I see it, it's a bug fix and should not be with a version flag.

The question of a version flag is whether anyone would have built a dependency on the current behavior, such that updating the behavior will break whichever application or service customers have built with the current behavior. I cannot say if customers of Liquid may have built an expectation that null prepended or appended remains null.

It concerns me that the current is very intentionally testing for null. Did author though that null string concatenation would result in an exception? Or did they actually see value in null taking precedence over the action, such that the action should not take effect?

daviburg:
> Closes #572.
> 
> Picked up by Golden Liquid tests.

Thank you for joining the project and congratulations on your first contribution!

microalps:
> > @daviburg giving you an opportunity to review. As I see it, it's a bug fix and should not be with a version flag.
> 
> The question of a version flag is whether anyone would have built a dependency on the current behavior, such that updating the behavior will break whichever application or service customers have built with the current behavior. I cannot say if customers of DotLiquid may have built an expectation that null prepended or appended remains null.
> 
> It concerns me that the current is very intentionally testing for null. Did author though that null string concatenation would result in an exception? Or did they actually see value in null taking precedence over the action, such that the action should not take effect?

It is my opinion that the authors of DotLiquid had limited access to the Ruby runtime and based it on existing test cases in Ruby's liquid and the limited public documentation. Much of the code is reproducing behaviors that the underlying Ruby itself is handling and the Liquid authors didn't even test for. At the end of the day, Ruby is not .NET and there is no perfect port.

src/DotLiquid/StandardFilters.cs
microalps: what is the intended result of null + null? blank or null?

microalps: It seems Ruby is blank in this case as well. While it seems above works, it might be clearer if we used string interpolation.

```c#
public static string Prepend(string input, string @string) => $"{@string}{input}";
```

RodneyRichardson: Well spotted! I've updated, and added a test.

RodneyRichardson:
I think a lot of the differences with the Ruby implementation are in the error conditions and edge cases - e.g. when a parameter is invalid or nil. I would expect people would be relying more on the non-edge case usage than the error cases, so there's probably less impact of changing this. And it's probably beneficial if most of the world is using the Ruby or Javascript versions, and so most of the examples people find have that behaviour.

daviburg (approved):

Issue #568, opened 2025-02-19
Issues with appveyor CI build


RodneyRichardson:
I've just submitted my first PR, and the appveyor build output (https://ci.appveyor.com/project/microalps/dotliquid/builds/51545325) has a number of issues:

- Hundreds of warnings raised during the build step. This makes it difficult to find any real problems. Suggest we at least disable the warning about missing XML comments, or do an automated generation of XML comments, to remove a majority.
- Warnings in nuget package build, including deprecated 'PackageIconUrl'/'iconUrl', and missing README.
- No coverage results generated (even though tests are executed). May require generation of PDB debug information.

PR #574, opened 2025-02-21
Fix or suppress build warnings

RodneyRichardson:
Fix or Suppress some of the build warnings. Should be no change to functionality. See #568.

I've tried to do each change as a single commit, so we can easily pull any you don't want.


RodneyRichardson:
I consider this ready for review/merge.

We may want to decide if we want to suppress the "Missing XML Comment" warnings (CS1591), as per this PR, or actually supply comments on all public functions.

RodneyRichardson:
There are a few remaining warnings
* `IContextAware.Context { set; }` - suggests replacing with `SetContext` method, but this would be a breaking change. We could do it being an Obsolete attribute? Or suppress the warning.
* `RubyDateFormat` - doesn't like this being a public field. Again, breaking change.
* netstandard1.3 is no longer supported, and could be dropped.

RodneyRichardson:
I think this will need to be redone on the develop branch after a few more PRs are merged (#571, #604)

Issue #549, opened 2024-06-18
The "include" tag with a "with" operator does not work correctly


tiritas:
### Dotliquid version
5.5.0
### Expected behavior
1. The `include` with a `with` operator should not be iterating IEnumerable objects. The `for` operator is for iterating the template for each item in an IEnumerable, but the `with` operator is only supposed to pass the variable to the included template. It was added because liquid was unable to pass variables that had the same name as the template itself.
### Actual behavior
The `with` operator acts just like `for`. This is particularly problematic, when a string is passed, because of a second issue: DotLiquid treats strings as arrays of characters (because they implement IEnumerable), resulting in the included template being rendered once for each character in the string.

I have fixed both of these issued and would like to make a pull request. Do I need to fork the project, or can I push a new branch directly to this project?

microalps:
@tiritas you would need to fork the project and submit a PR for review. Please make sure to fetch the latest upstream before branching to ensure merge success

microalps:
Related to this behavior is calling first and last filters on a string. Golden Liquid identifies that it should return empty.

microalps:
StandardFilterTests.TestFirstLast has explicit tests that are the opposite of the behavior of Ruby. First and last of a string returns first or last character. Added in #388 (https://github.com/dotliquid/dotliquid/pull/388/files#diff-1bc0a20d2eab851ec2bcfcd0e8c67ba185dd3a5e529a6d8086d7a30c3667382dR610) by @daviburg 

To be fair, I think the Ruby implementation isn't consistent. You can't get the first/last of a string - you get blank. You can loop over a string in a for statement - it is treated as a single element array.

```ruby
      if collection.is_a?(String)
        return collection.empty? ? [] : [collection]
```

microalps:
Opened a ticket with Golden Liquid as the first/last tests are actually wrong in the event activesupport/rails is enabled.

PR #575, opened 2025-02-21
Golden Liquid Test Implementation [skip-ci]

microalps:
Added a test project to execute the Golden Liquid test suite, as per #566.


src/DotLiquid.Tests/Embedded/golden_rules.json
```diff
@@ -0,0 +1,227 @@
...
+        "liquid.golden.illegal - no multiplication operator",
+        "liquid.golden.illegal - no subtraction operator",
+        "liquid.golden.include_tag - bound variable with alias",
+        "liquid.golden.include_tag - break from include",
+        "liquid.golden.include_tag - break from nested include",
+        "liquid.golden.join_filter - joining a string is a noop",
+        "liquid.golden.join_filter - joining an int is a noop",
+        "liquid.golden.last_filter - array of things",
+        "liquid.golden.last_filter - last of a hash",
+        "liquid.golden.last_filter - last of a string",
```
microalps, line 162: Looks like this category is DotLiquid #549

src/DotLiquid.Tests/Embedded/golden_rules.json
```diff
@@ -0,0 +1,227 @@
...
+        "liquid.golden.default_filter - allow false",
+        "liquid.golden.default_filter - allow false from context",
+        "liquid.golden.default_filter - false keyword argument before positional",
+        "liquid.golden.default_filter - true keyword argument before positional",
+        "liquid.golden.default_filter - zero is not falsy with allow_false",
+        "liquid.golden.default_filter - zero is not falsy with allow_false",
+        "liquid.golden.default_filter - render a default given a literal false with 'allow false' equal to false",
+        "liquid.golden.default_filter - render a default given a literal false with 'allow false' equal to true",
+        // DotLiquid #457
+        "liquid.golden.date_filter - seconds since epoch format directive",
```
microalps, line 40: This is failing due to timezone. As such CI doesn't pick up on this error since it operates in UTC. New group of flaky tests?

RodneyRichardson:
> @RodneyRichardson thank you for working together on this goal. Let me know if you have any further suggestions here.

I've sent a PR (https://github.com/microalps/dotliquid/pull/4) about checking for the right type of Exception. I guess it depends what we want these to do - if we happen to throw any kind of exception, and essentially end up with the correct response, is that acceptable? I think it would be better to know it's failing in the correct way, rather than just failing, but that probably means more work.

src/DotLiquid.Tests/GoldenLiquidTests.cs
```diff
@@ -0,0 +1,128 @@
...
+                }
+                else
+                {
+                    Assert.That(Template.Parse(test.Template, syntax).Render(parameters).Replace("\r\n", "\n"), Is.EqualTo(test.Want), test.UniqueName);
+                }
+            });
+        }
+
+        [Test]
+        public void ExecuteGoldenLiquidFailingTests()
```
RodneyRichardson, line 116: I suggest this takes a TestCaseSource too. See microalps#4.

PR #576, opened 2025-02-24
Corrections for string types in for/include tags

### LINKED ISSUE #549: The "include" tag with a "with" operator does not work correctly

microalps:
Replaces #515 resolve #549

### DIFF
--- a/src/DotLiquid.Tests/Tags/IncludeTagTests.cs
+++ b/src/DotLiquid.Tests/Tags/IncludeTagTests.cs
@@ -195,5 +195,24 @@ namespace DotLiquid.Tests.Tags
             }
             Assert.That(fileSystem.CacheHitTimes, Is.EqualTo(1));
         }
+
+        [Test]
+        public void TestIncludeTagWithStringVariable()
+        {
+            Helper.LockTemplateStaticVars(Template.NamingConvention, () => {
+                Template.FileSystem = new DictionaryFileSystem(new Dictionary<string, string>() { { "product", "Product: {{ product }}" } });
+                Assert.Multiple(() =>
+                {
+                    var expected = "Product: foo";
+                    Helper.AssertTemplateResult(expected, "{% include 'product' with 'foo' %}");
+                    Helper.AssertTemplateResult(expected, "{% include 'product' for 'foo' %}");
+
+                    Helper.AssertTemplateResult(expected, "{% include 'product' with var %}", Hash.FromAnonymousObject(new { var = "foo" }));
+                    Helper.AssertTemplateResult(expected, "{% include 'product' for var %}", Hash.FromAnonymousObject(new { var = "foo" }));
+                });
+            });
+
+        }
+
     }
 }
--- a/src/DotLiquid.Tests/Tags/IncludeTagTests.cs
+++ b/src/DotLiquid.Tests/Tags/IncludeTagTests.cs
@@ -211,8 +211,6 @@ namespace DotLiquid.Tests.Tags
                     Helper.AssertTemplateResult(expected, "{% include 'product' for var %}", Hash.FromAnonymousObject(new { var = "foo" }));
                 });
             });
-
         }
-
     }
 }
--- a/src/DotLiquid.Tests/Tags/StandardTagTests.cs
+++ b/src/DotLiquid.Tests/Tags/StandardTagTests.cs
@@ -165,6 +165,25 @@ namespace DotLiquid.Tests.Tags
         }
 
         [Test]
+        public void TestForWithString()
+        {
+            // Based on Liquid integration test_for_tag_string
+            var expected = "test string";
+            Helper.AssertTemplateResult(expected, "{%for val in str%}{{val}}{%endfor%}", Hash.FromAnonymousObject(new { str = expected }));
+            Helper.AssertTemplateResult(expected, "{%for val in str limit:1%}{{val}}{%endfor%}", Hash.FromAnonymousObject(new { str = expected }));
+            Helper.AssertTemplateResult(
+                expected: "val-str-1-1-0-1-0-true-true-test string",
+                template: "{%for val in str%}{{forloop.name}}-{{forloop.index}}-{{forloop.length}}-{{forloop.index0}}-{{forloop.rindex}}-{{forloop.rindex0}}-{{forloop.first}}-{{forloop.last}}-{{val}}{%endfor%}",
+                Hash.FromAnonymousObject(new { str = expected }));
+
+            // Additional tests for null and empty strings
+            string nullString = null;
+            Helper.AssertTemplateResult(string.Empty, "{%for val in str%}{{forloop.index}}{%endfor%}", Hash.FromAnonymousObject(new { str = nullString }));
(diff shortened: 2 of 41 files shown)


src/DotLiquid.Tests/Tags/StandardTagTests.cs
```diff
@@ -164,6 +164,25 @@ public void TestForWithRange()
...
+            Helper.AssertTemplateResult(
+                expected: "val-str-1-1-0-1-0-true-true-test string",
+                template: "{%for val in str%}{{forloop.name}}-{{forloop.index}}-{{forloop.length}}-{{forloop.index0}}-{{forloop.rindex}}-{{forloop.rindex0}}-{{forloop.first}}-{{forloop.last}}-{{val}}{%endfor%}",
+                Hash.FromAnonymousObject(new { str = expected }));
+
+            // Additional tests for null and empty strings
+            string nullString = null;
+            Helper.AssertTemplateResult(string.Empty, "{%for val in str%}{{forloop.index}}{%endfor%}", Hash.FromAnonymousObject(new { str = nullString }));
+            Helper.AssertTemplateResult(string.Empty, "{%for val in str%}{{forloop.index}}{%endfor%}", Hash.FromAnonymousObject(new { str = string.Empty }));
+
```
daviburg, line 183:
```suggestion

```

src/DotLiquid/Tags/For.cs
```diff
@@ -200,6 +200,9 @@ public override void Render(Context context, TextWriter result)
 
         private static List<object> SliceCollectionUsingEach(Context context, IEnumerable collection, int from, int? to)
         {
+            if (collection is string @string && @string.Length > 0)
+                return new List<object> { @string };
```
daviburg, line 204: I agree that this will make DotLiquid consistent with Ruby implementation. However, if a user of DotLiquid was transforming a string character by character through the for loop, this change will break them, right?

microalps: Yes, it will. I didn't imagine you would allow this, but I'm just prepping for the inevitable 3.0 branch to break free of these constraints.

daviburg: Ok. Sorry, I don't mean to be a pain in the rear, just looking out for the impact of code changes.

microalps: @daviburg Please approve my other PR for new Liquid syntax as it also includes DotLiquid 2.4 syntax. That will go to a new devel branch where myself and @RodneyRichardson will commit other syntax changes with 2.4. Once done, we'll merge that all into master. I'm ready to add a syntax check here since anyways context is available.

microalps:
@RodneyRichardson If you can review in lieu of @daviburg

src/DotLiquid.Tests/Tags/IncludeTagTests.cs
```diff
@@ -195,5 +195,24 @@ public void TestIncludeFromTemplateFileSystem()
...
+            Helper.LockTemplateStaticVars(Template.NamingConvention, () => {
+                Template.FileSystem = new DictionaryFileSystem(new Dictionary<string, string>() { { "product", "Product: {{ product }}" } });
+                Assert.Multiple(() =>
+                {
+                    var expected = "Product: foo";
+                    Helper.AssertTemplateResult(expected, "{% include 'product' with 'foo' %}");
+                    Helper.AssertTemplateResult(expected, "{% include 'product' for 'foo' %}");
+
+                    Helper.AssertTemplateResult(expected, "{% include 'product' with var %}", Hash.FromAnonymousObject(new { var = "foo" }));
+                    Helper.AssertTemplateResult(expected, "{% include 'product' for var %}", Hash.FromAnonymousObject(new { var = "foo" }));
```
RodneyRichardson, line 211: Rename variable `var`, as it's a reserved word. Maybe use `value`?

RodneyRichardson (approved):
Seems good to me. One small update suggested - variable name,

src/DotLiquid.Tests/Tags/IncludeTagTests.cs
```diff
@@ -195,5 +195,24 @@ public void TestIncludeFromTemplateFileSystem()
...
+                });
+            });
+
```
daviburg:
```suggestion
```

src/DotLiquid.Tests/Tags/IncludeTagTests.cs
```diff
@@ -195,5 +195,24 @@ public void TestIncludeFromTemplateFileSystem()
...
+        }
+
```
daviburg:
```suggestion
```

src/DotLiquid/Context.cs
```diff
@@ -698,7 +698,7 @@ private static object Liquidize(object obj)
             {
                 return liquidizableObj.ToLiquid();
             }
-            if (obj is string || obj is IEnumerable || obj is decimal || obj is DateTime || obj is DateTimeOffset || obj is TimeSpan || obj is Guid || obj is Enum
+            if (obj is IEnumerable /* string is IEnumerable */ || obj is decimal || obj is DateTime || obj is DateTimeOffset || obj is TimeSpan || obj is Guid || obj is Enum
```
daviburg:
```suggestion
            if (obj is IEnumerable /* string is IEnumerable */ || obj is decimal || obj is DateTime || obj is DateTimeOffset || obj is TimeSpan || obj is Guid || obj is Enum
```

(Trim trailing whitespace)

src/DotLiquid/Tags/Include.cs
```diff
@@ -60,9 +60,9 @@ public override void Render(Context context, TextWriter result)
                 foreach (var keyValue in _attributes)
                     context[keyValue.Key] = context[keyValue.Value];
 
-                if (variable is IEnumerable)
+                if (variable is IEnumerable enumerable && !(variable is string))
```
daviburg, line 63: I see you added new tests for include tag, but wouldn't this code line change the behavior compared to before? Should this also be behind the 2.4 syntax compatibility check?

daviburg (approved):

src/DotLiquid/Tags/Include.cs
microalps: I agreed with you on the for loop that people might have expected to be able to iterate through a string. but to include a template for a single char is not even a thought.

src/DotLiquid.Tests/Tags/IncludeTagTests.cs
microalps: There are ~45 instances of `var =` in anonymous type declarations. Apparently, a variable can't be named var, but a property or dictionary key can be.

PR #577, opened 2025-02-27
Refactor compatibility flags for filters

### LINKED ISSUE #558: Filter truncate_words naming inconsistent with Ruby
### Dotliquid version
All

### Expected behavior
```
{{ "Ground control to Major Tom." | truncatewords: 3 }}
```
should truncate as [documented](https://shopify.github.io/liquid/filters/truncatewords/)

### Actual behavior
Filter is ignored. Expects `truncate_words`.

microalps:
Allows for flagging filter methods as compatible with specific syntax compatibility flags. Also allows for aliases without creating new pointer methods. This branch will soon include Truncatewords alias and accompanying Golden Liquid tests.


src/DotLiquid/Context.cs
```diff
@@ -28,10 +28,20 @@ public class Context
...
         /// <summary>
         /// Liquid syntax flag used for backward compatibility
         /// </summary>
-        public SyntaxCompatibility SyntaxCompatibilityLevel { get; set; }
+        public SyntaxCompatibility SyntaxCompatibilityLevel {
+            get => _syntaxCompatibilityLevel;
+            set
+            {
+                if (_strainer != null)
+                    throw new System.ArgumentException("");
```
microalps, line 41: This needs a proper exception to disallow changing the compatibility level after strainers are computed.

microalps:
@RodneyRichardson would this structure work for your Numeric filters task? I think I still want to separate legacy filters to a separate class completely so in the future we can easily exclude compatibility by default.

RodneyRichardson:
I think this approach would work for a:
* Adding a new filter (like render) - this can be tagged with the version it was added, and be excluded in earlier versions. I think the current implementation will allow render, even if you set Syntax to DotLiquid20.
* Changing the behaviour of a filter - like Abs, Ceil, Floor, etc. These can be split into separate functions, and tagged appropriately.

We may need a (compile time? unit test?) check that at most one version of each filter was enabled for a specific syntax.
I think we could also default to DotLiquidLatest, and allow people to pin the syntax version explicitly if they didn't want new behaviour (although maybe we only have a default in the tests - I'd need to check).

microalps:
> I think this approach would work for a:
> 
> * Adding a new filter (like render) - this can be tagged with the version it was added, and be excluded in earlier versions. I think the current implementation will allow render, even if you set Syntax to DotLiquid20.

Render is a tag. I did this for filters only for now. However, I would need to implement if I want to change for tag behavior for strings without breaking changes for @daviburg 

> * Changing the behaviour of a filter - like Abs, Ceil, Floor, etc. These can be split into separate functions, and tagged appropriately.

Exactly, but let me restructure this first a bit.

> 
> We may need a (compile time? unit test?) check that at most one version of each filter was enabled for a specific syntax. I think we could also default to DotLiquidLatest, and allow people to pin the syntax version explicitly if they didn't want new behaviour (although maybe we only have a default in the tests - I'd need to check).

If there are two with the same signature, DotLiquid overwrites in the order of registered tags.

microalps:
Resolve #558

microalps:
Food for thought: opting out of legacy filters is not possible now. However, performance can be gained by leaving it out. Simple benchmark:

```
|              Method |        Job | BuildConfiguration | syntaxCompatibilityLevel |        Mean |      Error |     StdDev | Ratio | RatioSD |    Gen 0 |    Gen 1 |   Gen 2 | Allocated |
|-------------------- |----------- |------------------- |------------------------- |------------:|-----------:|-----------:|------:|--------:|---------:|---------:|--------:|----------:|
|               Parse | Job-UQALTN |              NuGet |              DotLiquid24 |    41.61 us |   5.095 us |   5.451 us |  1.00 |    0.00 |   2.9297 |   0.0610 |       - |     18 KB |
|               Parse | Job-HVSSAJ |            Default |              DotLiquid24 |    34.36 us |   1.311 us |   1.226 us |  0.81 |    0.11 |   2.9297 |   0.0610 |       - |     18 KB |
|                     |            |                    |                          |             |            |            |       |         |          |          |         |           |
|            ParseBig | Job-UQALTN |              NuGet |              DotLiquid24 |   124.18 us |   3.510 us |   3.605 us |  1.00 |    0.00 |  16.6016 |   1.2207 |       - |    102 KB |
|            ParseBig | Job-HVSSAJ |            Default |              DotLiquid24 |   128.52 us |   4.537 us |   4.660 us |  1.04 |    0.05 |  16.6016 |   1.2207 |       - |    102 KB |
|                     |            |                    |                          |             |            |            |       |         |          |          |         |           |
|    RenderNestedHash | Job-UQALTN |              NuGet |              DotLiquid24 | 5,270.22 us | 686.117 us | 734.138 us |  1.00 |    0.00 | 437.5000 |  93.7500 | 23.4375 |  2,749 KB |
|    RenderNestedHash | Job-HVSSAJ |            Default |              DotLiquid24 | 3,967.74 us | 253.068 us | 270.780 us |  0.77 |    0.11 | 437.5000 | 101.5625 | 27.3438 |  2,777 KB |
|                     |            |                    |                          |             |            |            |       |         |          |          |         |           |
| RenderILiquidizable | Job-UQALTN |              NuGet |              DotLiquid24 | 5,864.02 us | 563.711 us | 603.164 us |  1.00 |    0.00 | 437.5000 | 125.0000 | 23.4375 |  2,750 KB |
| RenderILiquidizable | Job-HVSSAJ |            Default |              DotLiquid24 | 4,482.84 us |  50.538 us |  54.075 us |  0.77 |    0.08 | 437.5000 |  93.7500 | 23.4375 |  2,777 KB |
```

RodneyRichardson:
How will this work with the `render` tag? This creates a new context, reads in a sub-template, parses it, renders it, then discards the context. In theory, that sub-template could have something like `{% param syntax='DotLiquid22' %}`. Would the new Context.Parse() set the syntax? Similarly, does the `include` tag throw an error if the sub-template tries to change the syntax?

microalps:
> How will this work with the `render` tag? This creates a new context, reads in a sub-template, parses it, renders it, then discards the context. In theory, that sub-template could have something like `{% param syntax='DotLiquid22' %}`. Would the new Context.Parse() set the syntax? Similarly, does the `include` tag throw an error if the sub-template tries to change the syntax?

I checked your PR and you have the following in NewIsolatedContext method: `subContext._strainer = _strainer;` that makes sense but won't allow included templates to change compatibility.

microalps:
@daviburg What are your thoughts about the split filter? Although it now matches ruby implementation, I don't expect a single .NET developer to understand what's under the hood. It seems as though the Ruby developers addressed a particular issue in their own applications and then imposed that solution on the broader community.

Empty string as separator splits into a char array (not changed). No trimming occurs.
Space as a separator implies two differences, namely (1) that it splits on control characters ascii 9-13 and space (2) that empty entries at beginning middle or end are removed.
Any other separator implies that empty entries at the end are removed.

Sources:
https://github.com/ruby/ruby/blob/master/doc/string/split.rdoc
https://www.rubydoc.info/stdlib/core/String:split

src/DotLiquid/Strainer.cs
```diff
@@ -31,16 +34,26 @@ public static void GlobalFilter(string rawName, object target, MethodInfo method
...
+                strainer.Extend(context.SyntaxCompatibilityLevel, typeof(LegacyFilters));
+
+            strainer.Extend(context.SyntaxCompatibilityLevel, typeof(StandardFilters));
 
             foreach (var keyValue in Filters)
-                strainer.Extend(keyValue.Value);
+                strainer.Extend(context.SyntaxCompatibilityLevel, keyValue.Value);
 
             foreach (var keyValue in FilterFuncs)
                 strainer.AddMethodInfo(keyValue.Key, keyValue.Value.Item1, keyValue.Value.Item2);
```
RodneyRichardson, line 55: I know this isn't changed, but using a Tuple instead of a custom class/struct/record makes this fairly unreadable. I had to go to the definition, then check where is was being populated to understand what's going on.

src/DotLiquid/Strainer.cs
```diff
@@ -18,7 +20,8 @@ public class Strainer
     {
         private static readonly Dictionary<string, Type> Filters = new Dictionary<string, Type>();
         private static readonly Dictionary<string, Tuple<object, MethodInfo>> FilterFuncs = new Dictionary<string, Tuple<object, MethodInfo>>();
```
RodneyRichardson, line 22: Related to comment below - can we replace Tuple with a type with more meaningly properties? Seeing Item1 and Item2 in the code makes it hard to read and understand, particularly when using `var`.

src/DotLiquid/Strainer.cs
```diff
@@ -1,8 +1,10 @@
 using System;
```
RodneyRichardson, line 1: Can we add some StrainerTests? I know the code tends to test things indirectly through the "system level" functions (e.g. Parse, Render), but it would be good to have dedicated tests for some of the main classes.

src/DotLiquid/Properties/Resources.resx
```diff
@@ -264,4 +264,7 @@
   <data name="DateFilterRubyDirectiveInvalid" xml:space="preserve">
     <value>Invalid directive passed to {0}</value>
   </data>
+  <data name="ContextPropertyReadonly" xml:space="preserve">
+    <value>Property '{0}' cannot be modified after the context has entered its current state.</value>
```
RodneyRichardson, line 268: I'm not sure what "entered its current state" means. Is there something more specific we can say here? Does it have explicit states?

src/DotLiquid/Strainer.cs
```diff
@@ -31,16 +34,26 @@ public static void GlobalFilter(string rawName, object target, MethodInfo method
...
+        /// Creates a new instance of the <see cref="Strainer"/> class.
+        /// </summary>
+        /// <param name="context">The DotLiquid context to be used for creating the <see cref="Strainer"/>.</param>
+        /// <returns>A new instance of the <see cref="Strainer"/> class.</returns>
         public static Strainer Create(Context context)
         {
-            Strainer strainer = new Strainer(context);
+            var strainer = new Strainer(context);
+
+            if (context.SyntaxCompatibilityLevel < SyntaxCompatibility.DotLiquidLatest)
```
RodneyRichardson, line 46: Should this be explicitly `< SyntaxCompatibility.DotLiquid24`?

src/DotLiquid/StandardFilters.cs
```diff
@@ -22,7 +22,7 @@ public static class StandardFilters
     {
```
RodneyRichardson, line 22: I would expect we should not have any code that now has a comparison to SyntaxCompatibility, but I think there are a few remaining (which I can't comment on directly in the PR for some reason) - e.g. Slice (line 79), ReplaceFirst, Date, etc. These need to be split into a completely separate function (even if that means duplicating code) to make the change clean.

src/DotLiquid/StandardFilters.cs
```diff
@@ -726,14 +706,10 @@ public static object Last(IEnumerable array)
         /// <param name="context">The DotLiquid context</param>
         /// <param name="input">Input to be transformed by this filter</param>
         /// <param name="operand">Number to be added to input</param>
+        [LiquidFilter(Name = "Plus", MinVersion = SyntaxCompatibility.DotLiquid21)]
```
RodneyRichardson, line 709: Is the `Name = "Plus"` still in the code? That should probably be removed.

src/DotLiquid/StandardFilters.cs
```diff
@@ -845,7 +814,7 @@ public static string Default(string input, string @defaultValue)
 
         private static bool IsReal(object o) => o is double || o is float || o is decimal;
 
-        private static object DoMathsOperation(Context context, object input, object operand, Func<Expression, Expression, BinaryExpression> operation)
+        internal static object DoMathsOperation(Context context, object input, object operand, Func<Expression, Expression, BinaryExpression> operation)
```
RodneyRichardson, line 817: Can we move this to another class, rather than LegacyFilters calling into StandardFilters? Maybe merge `ExpressiongUtility` and this into a `MathsOperation` class?

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -120,7 +123,20 @@ public void TestEscapeOnce()
         }
 
         [Test]
-        public void TestTruncateWords()
+        public void TestTruncateWordsV20()
```
RodneyRichardson, line 126: Can we split these into LegacyFilterTests?

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -90,7 +90,6 @@ public void TestEscape()
             Assert.That(StandardFilters.Escape(null), Is.EqualTo(null));
             Assert.That(StandardFilters.Escape(""), Is.EqualTo(""));
             Assert.That(StandardFilters.Escape("<strong>"), Is.EqualTo("&lt;strong&gt;"));
-            Assert.That(StandardFilters.H("<strong>"), Is.EqualTo("&lt;strong&gt;"));
 
             Helper.AssertTemplateResult(
```
RodneyRichardson, line 94: Maybe something for another day, but I find combining `Helper.AssertTemplateResult` and direct calls to `StandardTemplate.XXX` to be a little mixed. 

The latter definitely belongs in `StandardFilterTests` as unit tests, but maybe the `Helper.Assert` tests belong somewhere else as a system level test?

RodneyRichardson (changes requested):
I like the idea, but still needs a little work to make a clean separation from different versions of functions (i.e. remove all syntax checks from within function code).

src/DotLiquid/Strainer.cs
```diff
@@ -61,24 +78,38 @@ public Strainer(Context context)
...
         /// methods from the specified type and use them instead.
         /// </summary>
-        /// <param name="type"></param>
-        public void Extend(Type type)
+        /// <param name="syntaxCompatibilityLevel">The Liquid syntax flag used for backward compatibility</param>
+        /// <param name="type">The type from which static methods will be extracted and mixed to the current <see cref="Strainer"/>.</param>
+        public void Extend(SyntaxCompatibility syntaxCompatibilityLevel, Type type)
         {
             // Calls to Extend replace existing filters with the same number of params.
             var methods = type.GetRuntimeMethods().Where(m => m.IsPublic && m.IsStatic);
```
RodneyRichardson, line 86: Should this be `.Where(m => m.IsPublic && !m.IsStatic && !m.IsSpecialName)` to match code in Drop?

src/DotLiquid/Strainer.cs
microalps: When we introduce DotLiquid25, anything that is 2.4 specific would have been moved to legacy. Unless no change to filters happened (e.g. parse/tags). I'll go your way and comment that it should be adjusted as needed.

src/DotLiquid/ConstructorTagFactory.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        public ConstructorTagFactory(string tagName)
+        {
+            TagName = tagName;
+        }
+
+        /// <inheritdoc />
+        public Tag Create() => new T();
+    }
+}
```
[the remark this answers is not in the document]
microalps: Rolled back this breaking change, will eventually separate out to its own PR.

src/DotLiquid/StandardFilters.cs
microalps: I applied same logic to Slice, ReplaceFirst, and RemoveFirst. Leaving Date method alone for now.

src/DotLiquid.Tests/StandardFilterTests.cs
microalps: Please see new structure for version specific tests. It adds more tests (every version is tested) and is limited to unit tests only. It's WIP.

src/DotLiquid/Strainer.cs
microalps: The line you reference wasn't changed so I'm not going to change that.

daviburg:
> @daviburg What are your thoughts about the split filter? Although it now matches ruby implementation, I don't expect a single .NET developer to understand what's under the hood. It seems as though the Ruby developers addressed a particular issue in their own applications and then imposed that solution on the broader community.
> 
> Empty string as separator splits into a char array (not changed). No trimming occurs. Space as a separator implies two differences, namely (1) that it splits on control characters ascii 9-13 and space (2) that empty entries at beginning middle or end are removed. Any other separator implies that empty entries at the end are removed.
> 
> Sources: https://github.com/ruby/ruby/blob/master/doc/string/split.rdoc https://www.rubydoc.info/stdlib/core/String:split

Behaviors that are unique to an implementation, and do not translate to other implementation languages, are challenging. IIRC, there is already something around date time to choose between the Ruby behavior and the .NET behavior `Context.UseRubyDateFormat`. Split might call for another such context parameter (😖) so DotLiquid consumers can opt between wanting the same behavior they are used to from Ruby, or as .NET developers something that match their past experience. Overlay your versioning filter attribute for changing the default behavior to align with Liquid Gold test suite (i.e. when consumers do not explicitly opt for .NET like behavior).

daviburg:
(Requested a Copilot review, let's see if it has good suggestions for us)

src/DotLiquid.Tests/Filters/StandardFiltersV20Tests.cs
```diff
@@ -0,0 +1,104 @@
...
+using System.Globalization;
+using NUnit.Framework;
+
+namespace DotLiquid.Tests.Filters
+{
+    [TestFixture]
+    public class StandardFiltersV20Tests : StandardFiltersTestsBase
+    {
+        public override SyntaxCompatibility SyntaxCompatibilityLevel => SyntaxCompatibility.DotLiquid20;
+        public override CapitalizeDelegate Capitalize => i => LegacyFilters.Capitalize(_context, i);
```
daviburg, line 12: nit: avoid single letter character variable names. Consider using full English terms.

microalps:
> Behaviors that are unique to an implementation, and do not translate to other implementation languages, are challenging. IIRC, there is already something around date time to choose between the Ruby behavior and the .NET behavior `Context.UseRubyDateFormat`. Split might call for another such context parameter (😖) so DotLiquid consumers can opt between wanting the same behavior they are used to from Ruby, or as .NET developers something that match their past experience. Overlay your versioning filter attribute for changing the default behavior to align with Liquid Gold test suite (i.e. when consumers do not explicitly opt for .NET like behavior).

I think this would be a separate work item to allow custom aliasing. In this case, we would have an ExtendedFilter.RubySplit that works the .NET way and an opt-in to have split alias RubySplit method (at least so the golden liquid test can pass). Need to think about this and see if there is a way to massage the current APIs to serve this purpose.

src/DotLiquid/Tokenizer.cs
```diff
@@ -17,7 +17,7 @@ internal static class Tokenizer
         private static readonly HashSet<char> SearchQuoteOrVariableEnd = new HashSet<char> { '}', '\'', '"' };
         private static readonly HashSet<char> SearchQuoteOrTagEnd = new HashSet<char> { '%', '\'', '"' };
         private static readonly char[] WhitespaceCharsV20 = new char[] { '\t', ' ' };
-        private static readonly char[] WhitespaceCharsV22 = new char[] { '\t', '\n', '\v', '\f', '\r', ' ' };
+        internal static readonly char[] WhitespaceCharsV22 = new char[] { '\t', '\n', '\v', '\f', '\r', ' ' };
```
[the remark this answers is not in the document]
RodneyRichardson: I think this one should be addressed.

src/DotLiquid/Strainer.cs
RodneyRichardson: This could be raised as an issue (or added to #585) to be done later.

src/DotLiquid/StandardFilters.cs
RodneyRichardson: This could be raised as an issue, to be addressed later.

src/DotLiquid/Tokenizer.cs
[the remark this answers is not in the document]
microalps: I actually looked into this again - Shopify's implementation uses lstrip and rstrip for whitespace control. Shopify's implementation for split uses Ruby's split. 0The underlying Ruby implementation for all three uses `ascii_isspace`

microalps:
> > @daviburg @RodneyRichardson which comments are blockers from merging this into develop branch?
> 
> I have commented on my issues to say whether to fix or delay. There's only one I'd like to see fixed, as it affects the public interface.
> 
> There are also two "low confidence" issues that were picked up by the Copilot PR review above regarding comments that should be addressed too.

One low confidence issue is correct, I've used AI to rewrite the comments on the split method entirely. The other (context comment) is a false positive.

src/DotLiquid.Tests/ConditionTests.cs
```diff
@@ -356,12 +357,31 @@ public void TestDictionaryHasKey()
...
+        [Test]
+        public void TestExpandoHasKey()
+        {
+            _context = new Context(CultureInfo.InvariantCulture);
+            dynamic testDictionary = new ExpandoObject();
+            testDictionary.title = "Vacuum";
+            testDictionary.type = "cleaning";
+            _context["dictionary"] = testDictionary;
+            
+
```
daviburg, line 382:
```suggestion
```
Nit: remove duplicate empty line

src/DotLiquid.Tests/ConditionTests.cs
```diff
@@ -371,12 +391,31 @@ public void TestDictionaryHasValue()
...
+        [Test]
+        public void TestExpandoHasValue()
+        {
+            _context = new Context(CultureInfo.InvariantCulture);
+            dynamic testDictionary = new ExpandoObject();
+            testDictionary.title = "Vacuum";
+            testDictionary.type = "cleaning";
+            _context["dictionary"] = testDictionary;
+
+
```
daviburg, line 416:
```suggestion
```

src/DotLiquid.Tests/ContextTests.cs
```diff
@@ -233,6 +233,24 @@ public void TestVariables()
...
+        {
+            List<int> list = new List<int> { 1, 2, 3, 4, 5 };
+            _context["list"] = list;
+            Assert.That(_context["list"], Is.EqualTo(list));
+            Assert.That(_context["list[0]"], Is.EqualTo(1));
+            Assert.That(_context["list[-1]"], Is.EqualTo(5));
+            Assert.That(_context["list[12]"], Is.Null);
+            Assert.That(_context["list[-12]"], Is.Null);
+
+            List<string> emptyList = new List<string>();
```
daviburg, line 247:
```suggestion
            var emptyList = new List<string>();
```
nit: use `var` for obvious types.

src/DotLiquid.Tests/ContextTests.cs
```diff
@@ -233,6 +233,24 @@ public void TestVariables()
             Assert.That(_context["nil"], Is.EqualTo(null));
         }
 
+        [Test]
+        public void TestVariablesArray()
+        {
+            List<int> list = new List<int> { 1, 2, 3, 4, 5 };
```
daviburg, line 239:
```suggestion
            var list = new List<int> { 1, 2, 3, 4, 5 };
```

src/DotLiquid.Tests/ContextTests.cs
```diff
@@ -376,12 +394,19 @@ public void TestAddFilter()
...
+            
             context = new Context(CultureInfo.InvariantCulture);
             Assert.That(context.Invoke("hi", new List<object> { "hi?" }), Is.EqualTo("hi?"));
-            context.SyntaxCompatibilityLevel = SyntaxCompatibility.DotLiquid22;
+        }
+
+        [Test]
+        public void TestAddFilter_NotFoundException()
+        {
+            Context context = new Context(CultureInfo.InvariantCulture) { SyntaxCompatibilityLevel = SyntaxCompatibility.DotLiquid22 };
```
daviburg, line 405:
```suggestion
            var context = new Context(CultureInfo.InvariantCulture) { SyntaxCompatibilityLevel = SyntaxCompatibility.DotLiquid22 };
```
(I do see that this code has not made use of `var` for joint declaration and assignment - this is only a nit, feel free to ignore)

RodneyRichardson: Maybe I'm old school but I have a preference to specify the type if you know it, rather than `var`. That way it's obvious to the reader. I wouldn't think the compiler cares either way (if it can infer the type).

I'm curious when you'd choose `var` over a known type.

src/DotLiquid.Tests/Filters/StandardFiltersV20Tests.cs
microalps: Not going to fix this, it is in test suite within a lambda expression to proxy parameter 1, 2, 3, n. Using a single letter is acceptable in this case for easier readability of the critical part here - LEGACYFILTERS vs STANDARDFILTERS.

src/DotLiquid/StandardFilters.cs
microalps: I don't want to mess with Math ops at the moment. Agree to defer.

src/DotLiquid/Strainer.cs
microalps: Agree to defer, although there is definitely opportunity to add unit tests to test the new functionality.

src/DotLiquid.Tests/ContextTests.cs
RodneyRichardson: I've looked into this further - it seems common practice to use `var` for constructor assignment (exactly like `var context = new Context()`) where the type is written on the same line. That seems a good rule of thumb.

RodneyRichardson:
@microalps : This looks good to me. What's the intention with the `develop` branch? Will it be merged into master, or is it a v3? When should I target PRs to it, and when to master?

I have one PR #584 targeting `develop`, and it seems maybe #590 could go there (to use the new attribute) and maybe #571 and #571, depending on the branch's purpose.

microalps:
> @microalps : This looks good to me. What's the intention with the `develop` branch? Will it be merged into master, or is it a v3? When should I target PRs to it, and when to master?
> 
> I have one PR #584 targeting `develop`, and it seems maybe #590 could go there (to use the new attribute) and maybe #571 and #571, depending on the branch's purpose.

develop is intended to allow for 2.4 compatibility flag to be fully defined before we merge to master. Otherwise this version flag is a moving target.

v3 will be a separate repository completely with different build settings and push to NuGet as v3.0.x-dev. @tgjones would be nice to be granted admin rights on the dotliquid organization to do so, otherwise I will need to fork to my own account.

PR #578, opened 2025-02-28
Check for list index out of bounds

### LINKED ISSUE #531: IndexOutOfRangeException
### Dotliquid version
2.2.508

### Expected behavior
value=null or value=string.Empty

### Actual behavior
throw IndexOutOfRangeException

TryEvaluateHashOrArrayLikeObject methods in Context class does not check if the index is a valid key and throws IndexOutOfRangeException.

image

RodneyRichardson:
Closes #531.

### DIFF
--- a/src/DotLiquid.Tests/ContextTests.cs
+++ b/src/DotLiquid.Tests/ContextTests.cs
@@ -242,6 +242,7 @@ namespace DotLiquid.Tests
             Assert.That(_context["list[0]"], Is.EqualTo(list[0]));
             Assert.That(_context["list[-1]"], Is.EqualTo(list[list.Count - 1]));
             Assert.That(_context["list[12]"], Is.Null);
+            Assert.That(_context["list[-12]"], Is.Null);
 
             List<string> emptyList = new List<string>();
             _context["empty_list"] = emptyList;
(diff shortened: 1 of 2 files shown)


src/DotLiquid.Tests/ContextTests.cs
```diff
@@ -233,6 +233,24 @@ public void TestVariables()
...
         }
 
+        [Test]
+        public void TestVariablesArray()
+        {
+            List<int> list = new List<int> { 1, 2, 3, 4, 5 };
+            _context["list"] = list;
+            Assert.That(_context["list"], Is.EqualTo(list));
+            Assert.That(_context["list[0]"], Is.EqualTo(list[0]));
+            Assert.That(_context["list[-1]"], Is.EqualTo(list[list.Count - 1]));
```
microalps, line 243:
```suggestion
            Assert.That(_context["list[0]"], Is.EqualTo(1));
            Assert.That(_context["list[-1]"], Is.EqualTo(5));
```
expectations should be as concrete as possible.

microalps (commented):
This looks good to go with small changes. We should track though that this is not evaluated in golden liquid itself.

microalps (approved):

PR #579, opened 2025-02-28
Fix Condition when right operand is null (contains, startswith, endswith)

RodneyRichardson:
Prevents exceptions being thrown when right = null. Also adds test for when left operand is null, but these were handle correctly already.

Replaces stale #513.

### DIFF
--- a/src/DotLiquid/Condition.cs
+++ b/src/DotLiquid/Condition.cs
@@ -26,10 +26,10 @@ namespace DotLiquid
             { ">", (left, right) => left != null && right != null && Comparer<object>.Default.Compare(left, Convert.ChangeType(right, left.GetType())) == 1 },
             { "<=", (left, right) => left != null && right != null && Comparer<object>.Default.Compare(left, Convert.ChangeType(right, left.GetType())) <= 0 },
             { ">=", (left, right) => left != null && right != null && Comparer<object>.Default.Compare(left, Convert.ChangeType(right, left.GetType())) >= 0 },
-            { "contains", (left, right) => ((left is string) ? ((string)left).Contains((string)right) : (left is IEnumerable) ? Any((IEnumerable) left, (element) => element.BackCompatSafeTypeInsensitiveEqual(right)) : false) },
-            { "startsWith", (left, right) => (left is IList) ? EqualVariables(((IList) left).OfType<object>().FirstOrDefault(), right) : ((left is string) ? ((string)left).StartsWith((string) right) : false) },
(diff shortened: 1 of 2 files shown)


src/DotLiquid/Condition.cs
```diff
@@ -26,10 +26,10 @@ public class Condition
...
             { ">=", (left, right) => left != null && right != null && Comparer<object>.Default.Compare(left, Convert.ChangeType(right, left.GetType())) >= 0 },
-            { "contains", (left, right) => ((left is string) ? ((string)left).Contains((string)right) : (left is IEnumerable) ? Any((IEnumerable) left, (element) => element.BackCompatSafeTypeInsensitiveEqual(right)) : false) },
-            { "startsWith", (left, right) => (left is IList) ? EqualVariables(((IList) left).OfType<object>().FirstOrDefault(), right) : ((left is string) ? ((string)left).StartsWith((string) right) : false) },
-            { "endsWith", (left, right) => (left is IList) ? EqualVariables(((IList) left).OfType<object>().LastOrDefault(), right) : ((left is string) ? ((string)left).EndsWith((string) right) : false) },
-            { "hasKey", (left, right) => (left is IDictionary) ? ((IDictionary) left).Contains(right) : false },
+            { "contains", (left, right) => ((left is string && right != null) ? ((string)left).Contains((string)right) : (left is IEnumerable) ? Any((IEnumerable) left, (element) => element.BackCompatSafeTypeInsensitiveEqual(right)) : false) },
```
microalps, line 29: Besides rebasing, did you consider the below alternative?
```suggestion
            { "contains", (left, right) => ((left is string leftString) ? leftString.Contains((string)right ?? string.Empty) : (left is IEnumerable) ? Any((IEnumerable) left, (element) => element.BackCompatSafeTypeInsensitiveEqual(right)) : false) },

```

Emphasis on ` ?? string.Empty`. If you think about it, it doesn't mean the same thing, since string is also IEnumerable and will evaluate the second half if right is null. Same for the next few.

RodneyRichardson: I just tried this, and tests fail for `'bob' contains null`, and equivalent.

RodneyRichardson:
Rebased and ready to go...

microalps:
I've reviewed one more case and [confirmed](https://replit.com/@microalps/LiquidUnitTests#main.rb) that liquid returns TRUE for empty string on the right. Adding test cases and cleaning up some messages.

RodneyRichardson:
Looks good. I fixed a variable name, and added some more tests.

I've noticed we're inconsistent with our string/numeric conversions in these - I've added tests to explicitly show the quirk. But it means `contains` will return false, but `endswith`/`startswith` might return true for the same arguments.

I don't propose we change this now, but maybe in a v3?

microalps (approved):

PR #581, opened 2025-03-04
Change all R.Scan() calls to use Regex

RodneyRichardson:
I noticed only one of the overloads were made Obsolete. 

Replace R.Scan(string, string, Action) with R.Scan(string, Regex, Action).

### DIFF
--- a/src/DotLiquid.Tests/RegexpTests.cs
+++ b/src/DotLiquid.Tests/RegexpTests.cs
@@ -30,6 +30,11 @@ namespace DotLiquid.Tests
         }
 #endif
 
+        private static List<string> Run(string input, string pattern)
+        {
+            return R.Scan(input, new Regex(pattern));
+        }
+
         [Test]
         public void TestEmpty()
         {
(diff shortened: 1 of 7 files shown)


src/DotLiquid/Liquid.cs
```diff
@@ -38,10 +37,12 @@ public static class Liquid
         private static readonly Lazy<Regex> LazyDirectorySeparatorsRegex = new Lazy<Regex>(() => R.C(DirectorySeparators), LazyThreadSafetyMode.ExecutionAndPublication);
         private static readonly Lazy<Regex> LazyLimitRelativePathRegex = new Lazy<Regex>(() => R.C(LimitRelativePath), LazyThreadSafetyMode.ExecutionAndPublication);
         private static readonly Lazy<Regex> LazyVariableSegmentRegex = new Lazy<Regex>(() => R.B(R.Q(@"\A\s*(?<Variable>{0}+)\s*\Z"), Liquid.VariableSegment), LazyThreadSafetyMode.ExecutionAndPublication);
+        private static readonly Lazy<Regex> LazyTagAttributesRegex = new Lazy<Regex>(() => R.B(R.Q(@"(\w+)\s*\:\s*({0})"), QuotedFragment), LazyThreadSafetyMode.ExecutionAndPublication);
 
         internal static Regex DirectorySeparatorsRegex => LazyDirectorySeparatorsRegex.Value;
         internal static Regex LimitRelativePathRegex => LazyLimitRelativePathRegex.Value;
         internal static Regex VariableSegmentRegex => LazyVariableSegmentRegex.Value;
+        internal static Regex TagAttributesRegex => LazyTagAttributesRegex.Value;
```
microalps, line 45: This isn't new and the above Regex have the same limitation, but I noticed that Template.RegexTimeout would be evaluated only once. This is an unexpected behavior in general. Is there anyway to avoid this? It's not critical as it isn't thread-safe anyways to alter the value between runs.

src/DotLiquid/Util/R.cs
```diff
@@ -59,6 +59,19 @@ public static List<string> Scan(string input, Regex regex)
...
+        /// Overload that only works when the pattern contains two groups. The callback
+        /// is called for each match, passing the two group values.
+        /// </summary>
+        /// <param name="input"></param>
+        /// <param name="regex"></param>
+        /// <param name="callback"></param>
+        public static void Scan(string input, Regex regex, Action<string, string> callback)
+        {
+            foreach (Match match in regex.Matches(input))
+                callback(match.Groups[1].Value, match.Groups[2].Value);
```
microalps, line 72: Do we really need this, or can we inline this to the 3 callers? Seems easy enough to do. And if yes, I question the public part (which is why we need to obsolete and not delete the old method)

src/DotLiquid/Util/R.cs
```diff
@@ -79,7 +92,7 @@ public static List<string> Scan(string input, string pattern)
         /// <param name="input"></param>
         /// <param name="pattern"></param>
         /// <param name="callback"></param>
-        /// <returns></returns>
+        [Obsolete("Use Scan(string, Regex, Action) instead.")]
```
microalps, line 95: Should we add a call to this in RegexpTests to avoid the code coverage drop?

src/DotLiquid/Util/R.cs
RodneyRichardson: As a general rule, I think if it's used in at least three places, it's a good candidate for refactor/function. So I'd keep it.

src/DotLiquid/Util/R.cs
RodneyRichardson: Ah - I didn't see any tests for the old function, so it didn't register that the coverage would drop, but we're testing it indirectly... I'll add a test.

src/DotLiquid/Liquid.cs
RodneyRichardson: Hmm, tricky. Looks like MatchTimeout can't be specified when Match is called, only when the Regex is created. Many of the Regex are `static readonly` variables (apart from the Lazy ones), so will be evaluated close to when the assembly is loaded.

I see several options (in order of complexity):

1. Do nothing. It's not causing a problem. 
2. Remove the RegexTimeOut variable, and just use a hardcoded constant (possibly Timeout.Infinite, possibly per-Regex).
3. Change all Regex to be Lazy<Regex>, which will delay setting the timeout to the first time it's used. 
4. Replace Lazy<Regex> with an equivalent class that allows the Value to be discarded. Discard all Regex when RegexTimeOut is changed, so that the new timeout value will be picked up on use. 

I think Option 4 is probably the most correct, but I think Option 2 is probably the most sensible and easiest for something that's probably not causing a problem.

RodneyRichardson: Either way, it should probably be logged as an issue and addressed separately.

src/DotLiquid.Tests/RegexpTests.cs
```diff
@@ -74,9 +79,146 @@ public void TestQuotedWordsInTheMiddle()
...
+            R.Scan(markup, Liquid.TagAttributesRegex, (key, value) => d[key] = value);
+
+            Assert.That(d.Count, Is.EqualTo(2));
+            Assert.That(d["hello"], Is.EqualTo("world"));
+            Assert.That(d["hello_again"], Is.EqualTo("another_world"));
         }
+
+        #region Tests of obsolete functions
+
+        [Obsolete("Tests obsolete function")]
```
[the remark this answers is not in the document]
RodneyRichardson: I think it actually prevents the warning raised by calling the obsolete function inside, and means we don't need to add a #pragma to suppress the warning. I thought it was cleaner than how I did it on my other PR about fixing warnings (which I will update).

I think it's because we don't actually call the function itself - it's all done via Reflection in the NUnit test framework.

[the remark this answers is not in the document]
RodneyRichardson: I checked this by doing a Rebuild Solution and then searched for "obsolete" in the Error List. I only warnings obsolete warnings about RethrowErrors, and Context constructor without CancelationToken.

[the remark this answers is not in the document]
microalps: Compared output in AppVeyor and you seem correct.

src/DotLiquid/Util/R.cs
microalps: Add 5 lines of code to reduce 3 places from 2 lines to 1. Net total of 2 extra lines. I disagree, especially since it isn't more readable (what is action<string, string>?) and there is overhead of n+1 stacktrace changes depending on how many matches are found.

RodneyRichardson: No worries. I've removed it, but still marked the old one Obsolete.

PR #582, opened 2025-03-04
Allow non-printable whitespace in quoted blocks

RodneyRichardson:
Fixes some of the failing golden liquid tests related to newline/whitespace.

I also tweaked the Golden Liquid tests to allow either `\r\n` or `\n` in expected result (rather than always replacing `\r\n`).


microalps:
My approach was to drop regex completely in #561. What is the performance impact of this change?

RodneyRichardson:
Ah - I hadn't reaslised you were working on this as well! 

I haven't checked performance, but I'd be suprised if there was any change. I don't think performance is much of an issue with DotLiquid, though, is it?

I'll have a look at your PR #561.

PR #584, opened 2025-03-05
Drop unsupported netstandard1.0

RodneyRichardson:
Remove TargetFramework netstandard1.0, and associated code (`#if CORE`, `#if NETSTANDARD_1_3`). Support stopped in 2020.

Anyone starting new code should not be using this, and anyone maintaining older code can stick with older releases of DotLiquid.

Also: Add net8.0 support, and bump net461 to net462 (which is the oldest supported). We'll leave in net6.0 for now, as support has only just ended (Nov 2024) and there are probably some projects still using it.

### DIFF
--- a/src/DotLiquid/Exceptions/LiquidException.cs
+++ b/src/DotLiquid/Exceptions/LiquidException.cs
@@ -3,7 +3,7 @@ using System;
 namespace DotLiquid.Exceptions
 {
     [Serializable]
-    public abstract class LiquidException : ApplicationException
+    public abstract class LiquidException : Exception
     {
         protected LiquidException(string message, Exception innerException)
             : base(message, innerException)
--- a/src/DotLiquid/Exceptions/LiquidException.cs
+++ b/src/DotLiquid/Exceptions/LiquidException.cs
@@ -2,15 +2,8 @@ using System;
 
 namespace DotLiquid.Exceptions
 {
-#if !CORE
     [Serializable]
-#endif
-    public abstract class LiquidException :
-#if CORE
-        Exception
-#else
-        ApplicationException
-#endif
+    public abstract class LiquidException : ApplicationException
     {
         protected LiquidException(string message, Exception innerException)
             : base(message, innerException)
--- a/src/DotLiquid/Exceptions/RenderException.cs
+++ b/src/DotLiquid/Exceptions/RenderException.cs
@@ -1,16 +1,9 @@
-﻿using System;
+using System;
 
 namespace DotLiquid.Exceptions
 {
-#if !CORE
     [Serializable]
-#endif
-    public abstract class RenderException :
-#if CORE
-        Exception
-#else
-        ApplicationException
-#endif
+    public abstract class RenderException : LiquidException
     {
         protected RenderException(string message, Exception innerException)
             : base(message, innerException)
--- a/src/DotLiquid/StandardFilters.cs
+++ b/src/DotLiquid/StandardFilters.cs
@@ -23,33 +23,6 @@ namespace DotLiquid
         private static readonly Lazy<Regex> StripHtmlBlocks = new Lazy<Regex>(() => R.C(@"<script.*?</script>|<!--.*?-->|<style.*?</style>", RegexOptions.Singleline | RegexOptions.IgnoreCase), LazyThreadSafetyMode.ExecutionAndPublication);
         private static readonly Lazy<Regex> StripHtmlTags = new Lazy<Regex>(() => R.C(@"<.*?>", RegexOptions.Singleline), LazyThreadSafetyMode.ExecutionAndPublication);
 
-#if NETSTANDARD1_3
-        private class StringAwareObjectComparer : IComparer
-        {
-            private readonly StringComparer _stringComparer;
-
-            public StringAwareObjectComparer(StringComparer stringComparer)
-            {
-                _stringComparer = stringComparer;
-            }
-
-            public int Compare(Object x, Object y)
-            {
-                if (x == y)
-                    return 0;
-                if (x == null)
-                    return -1;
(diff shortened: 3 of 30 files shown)


RodneyRichardson:
I think the code coverage change/decrease is a false positive, as I've only deleted code. Maybe it's not great when combining code coverage of different TargetFrameworks?

src/DotLiquid/Strainer.cs
```diff
@@ -142,19 +142,8 @@ public object Invoke(string method, List<object> args)
...
-                            .IsAssignableFrom(
-                                convertibleArg
-                                    .GetType()
-#if NETSTANDARD1_3
-                                    .GetTypeInfo()
-#endif
-                                    )
-                        )
+                    if (convertibleArg.GetType() != parameterType &&
+                        !parameterType.IsInstanceOfType(convertibleArg))
```
RodneyRichardson, line 146: OK - I did change this line of code from `IsAssignableFrom` to `IsInstanceOfType`. I'll change it back, just so there's no code changes at all in this PR.

RodneyRichardson:
I've reverted the minor changes I made, plus some of the formatting ones to make the diff smaller, and hopefully please the code coverage...

src/DotLiquid/StandardFilters.cs
```diff
@@ -643,11 +611,7 @@ public static string Date(Context context, object input, string format)
             }
             else if ((input is decimal) || (input is double) || (input is float) || (input is int) || (input is uint) || (input is long) || (input is ulong) || (input is short) || (input is ushort))
             {
-#if CORE
-                dateTimeOffset = DateTimeOffset.FromUnixTimeSeconds(Convert.ToInt64(input)).ToLocalTime();
```
microalps, line 647: seems like this should have been adjusted to be used for .NET 6.0 and higher builds.

src/DotLiquid/Util/R.cs
```diff
@@ -32,9 +32,7 @@ public static Regex B(string format, params string[] args)
         /// <returns>the regex</returns>
         public static Regex C(string pattern, RegexOptions options = RegexOptions.None)
         {
-#if !CORE
             options = options | RegexOptions.Compiled;
-#endif
             var regex = new Regex(pattern, options, Template.RegexTimeOut);
```
microalps, line 36: Remove previous line and
```suggestion
            var regex = new Regex(pattern, options | RegexOptions.Compiled, Template.RegexTimeOut);
```

microalps (commented):
Couple of notes:
1. Appveyor is for reference only, not actually impacted by change. Need to remember to change Appveyor itself before merging.
2. Probably should merge into devel for a future 2.4 build
3. Review changes to code coverage not sure if it's caused by your changes or not.

src/DotLiquid/Exceptions/LiquidException.cs
```diff
@@ -2,15 +2,8 @@
...
-#if !CORE
     [Serializable]
-#endif
-    public abstract class LiquidException :
-#if CORE
-        Exception
-#else
-        ApplicationException
-#endif
+    public abstract class LiquidException : Exception
```
rolandbaer, line 6: Shouldn't this be ApplicationException (as always the !CORE part is used, the else part should be used here)

src/DotLiquid/Exceptions/RenderException.cs
```diff
@@ -1,16 +1,9 @@
...
-#if !CORE
     [Serializable]
-#endif
-    public abstract class RenderException :
-#if CORE
-        Exception
-#else
-        ApplicationException
-#endif
+    public abstract class RenderException : Exception
```
rolandbaer, line 6: Shouldn't this be ApplicationException (as always the !CORE part is used, the else part should be used here)

src/DotLiquid/Exceptions/LiquidException.cs
microalps: https://stackoverflow.com/questions/5685923/what-is-applicationexception-for-in-net

RodneyRichardson: Well spotted - so technically, this is a change. @microalps : Do we want this PR to be essentially a "no change"? Or do we want to swap to `Exception` for better practice?

src/DotLiquid/Exceptions/RenderException.cs
RodneyRichardson: I think this should probably be a `LiquidException` in a future version (so we have a single base class for exceptions), but I'll change this to be the same as the comment above (whichever was @microalps prefers).

src/DotLiquid/Exceptions/LiquidException.cs
microalps: What if we keep LiquidException as inheriting from ApplicationException (for now) and swap the others to inherit from LiquidException? This would make a one line change to switch all to best practice, and provide guidance to developers about the "deprecation".

RodneyRichardson: OK - RenderException now inherits from LiquidException, and LiquidException from ApplicationException.

src/DotLiquid/StandardFilters.cs
RodneyRichardson: Adjusted in what way? The subsequent line (which is left in) seems fine to me, and the tests pass on net461, coreapp2.0 and net6.0.

RodneyRichardson:
@microalps : Did you want me to merge this into the `develop` branch instead of `master`? If so, we should add net8.0 (maybe in a different PR, if it causes any issues)

RodneyRichardson:
Recreated branch and a new PR #592 to resolve conflicts.

Issue #394, opened 2020-06-24
Implement Truthiness of types Unit Tests


daviburg:
### Dotliquid version
Current as of 6-23-2020

### Expected behavior
> All values in Liquid are truthy except nil and false.

Text should appear because 

> the text "Tobi" is not a boolean, but it is truthy in a conditional

### Actual behavior
Blank output, text does not appear

### Steps to reproduce the Problem (you can add files)
got to http://dotliquidmarkup.org/try-online
paste
```
{% assign name = "Tobi" %}

{% if name == true %}
  This text will always appear if "name" is defined.
{% endif %}
```

daviburg:
Internally the match condition is trying to convert the types of the two operands to find a common ground, settles on string (as boolean can be converted to string value 'True' or 'False'). But instead we should overwrite the behavior and special case when either of the operand is boolean and not the other, the other type needs to follow the Shopify 'Truthiness' spec.

daviburg:
For backward compatibility reasons we should probably put this and other spec compliance fixes behind a flag for strict Shopify specification compliance.

jolleekin:
That is expected. You're trying to compare `"Tobi"` and `true`, which are completely different (by type and by value). What you should write instead is
```
{% assign name = "Tobi" %}

{% if name %}
  This text will always appear if "name" is defined.
{% endif %}
```

daviburg:
That 'name' would be truthy is consistent with example in
https://shopify.github.io/liquid/basics/truthy-and-falsy/

> ### Truthy
> All values in Liquid are truthy except **nil** and **false**.
> 
> In the example below, the string “Tobi” is not a boolean, but it is truthy in a conditional:
```
{% assign tobi = "Tobi" %}

{% if tobi %}
  This condition will always be true.
{% endif %}
```
> Strings, even when empty, are truthy. The example below will result in empty HTML tags if **settings.fp_heading** is empty:
> 
> Input

```
{% if settings.fp_heading %}
  <h1>{{ settings.fp_heading }}</h1>
{% endif %}
```
> Output

`<h1></h1>`

But consider also same specification with additional details at https://shopify.dev/docs/themes/liquid/reference/basics/true-and-false
Which gives the explicit cross-type equality example:

> In the example below, the text "Tobi" is not a boolean, but it is truthy in a conditional:
```
{% assign name = "Tobi" %}

{% if name == true %}
  This text will always appear if "name" is defined.
{% endif %}
```
[...]

PR #586, opened 2025-03-11
SafeTypeInsensitiveEqual: Compare boolean and true/false strings

RodneyRichardson:
I think I'm dredging up a past issue here, related to #394 and #434.

Should we do comparison between boolean values, and string values containing with true/false? I've added the code to `SafeTypeInsensitiveEqual`, but it's obviously a change in behaviour and probably needs to be behind a SyntaxCompatibility flag.

I haven't updated all the tests yet, depending on how we proceed with this (hence why it's still draft).

I had a few questions though:
- `BackCompatSafeTypeInsensitiveEqual`: This feels like it should also be behind a SyntaxCompatibility flag.
- Do we want the behaviour in this PR? 

@daviburg , @microalps


src/DotLiquid/Util/ObjectExtensionMethods.cs
```diff
@@ -109,6 +109,20 @@ public static bool SafeTypeInsensitiveEqual(this object value, object otherValue
...
+            // NOTE(Rodney Richardson): If either value is boolean or the string "true" or "false", compare with Falsy/Truthy
+            // We can't use IsTruthy() directly, as everything that's not Falsy is considered Truthy
+            bool valueIsFalsy = value.IsFalsy();
+            bool valueIsTruthy = (value is bool boolValue && boolValue == true)
+                || (value is string stringValue && "true".Equals(stringValue, StringComparison.OrdinalIgnoreCase));
+            bool otherValueIsFalsy = otherValue.IsFalsy();
+            bool otherValueIsTruthy = (otherValue is bool boolOtherValue && boolOtherValue == true)
+                || (otherValue is string stringOtherValue && "true".Equals(stringOtherValue, StringComparison.OrdinalIgnoreCase));
+
+            if (valueIsFalsy) return otherValueIsFalsy;
```
daviburg, line 121: Please observe the code formatting convention of the rest of the file - explicit code block for if conditions.

```suggestion
            if (valueIsFalsy)
            {
                return otherValueIsFalsy;
            }
```

etc.

src/DotLiquid/Util/ObjectExtensionMethods.cs
```diff
@@ -109,6 +109,20 @@ public static bool SafeTypeInsensitiveEqual(this object value, object otherValue
                 return false;
             }
 
+            // NOTE(Rodney Richardson): If either value is boolean or the string "true" or "false", compare with Falsy/Truthy
+            // We can't use IsTruthy() directly, as everything that's not Falsy is considered Truthy
```
daviburg, line 113: but that contradicts the Shopify documentation referred to by the IsTruthy method.

daviburg:
> I think I'm dredging up a past issue here, related to #394 and #434.
> 
> Should we do comparison between boolean values and string values containing with true/false? I've added the code to `SafeTypeInsensitiveEqual`, but it's obviously a change in behaviour and probably needs to be behind a SyntaxCompatibility flag.
> 
> I haven't updated all the tests yet, depending on how we proceed with this (hence why it's still draft).
> 
> I had a few questions though:
> 
> * `BackCompatSafeTypeInsensitiveEqual`: This feels like it should also be behind a SyntaxCompatibility flag.
> * Do we want the behaviour in this PR?
> 
> @daviburg , @microalps

What is the rationale behind this change? The change appears to contradict what is described in the Shopify documentation as referred to by the current implementation.

src/DotLiquid/Util/ObjectExtensionMethods.cs
RodneyRichardson: I had it like that, then shrank it down because it looked too bulky! :)

RodneyRichardson:
> What is the rationale behind this change? The change appears to contradict what is described in the Shopify documentation as referred to by the current implementation.

I've been working through some of the golden liquid test failures with the assumption they all pass on Shopify/Ruby, and that we want to be more like them if we can - in this case the "one is not equal to true" and "zero is not equal to false" tests. 

Technically, the string `"false"` is considered Truthy (since it's a string) but DotLiquid considered it Falsy. I'll have a look at this again tomorrow!

daviburg:
We do want to align with Shopify's implementation indeed. When Shopify's implementation contradicts their docs however, then we need to first clarify with Shopify if it is a doc bug or an implementation bug.

RodneyRichardson:
This PR has change entirely, but I wanted to keep some of the above conversation. 

I think the reason we were failing some of the golden liquid tests is that we were considering the string "false" to be Falsy, but the documentation says that only boolean false and nil/null are Falsy. It states strings, including the empty string, are Truthy.

I've updated the code and tests, but left this as a Draft as we'll need to put this behind a SyntaxCompatibility flag. I just wanted to get feedback before working out how to do that, since this code doesn't have access to the context, so a bit of restructuring will be needed.

src/DotLiquid.Tests/StandardFilterTests.cs
```diff
@@ -2014,10 +2014,11 @@ public void TestWhere_NonStringCompare()
...
                 localVariables: Hash.FromAnonymousObject(new { products }));
 
-            // The products array has a price of 7.0f, compare to the string '7.0'
+            // The products array has a price of 7.0f, compare to the string '7'
+            // Expect string != number
             Helper.AssertTemplateResult(
-                expected: "\r\n\r\nCheapest products:\r\n\r\n- Spatula\r\n",
-                template: @"{% assign cheap_products = products | where: ""price"", '7.0' %}
+                expected: "\r\n\r\nCheapest products:\r\n",
+                template: @"{% assign cheap_products = products | where: ""price"", '7' %}
```
daviburg, line 2021: you are changing the test from explicit float to probably a defaulting to integer. It's ok to have a test with integer comparison to float, but you should have as an additional test then.

src/DotLiquid.Tests/Util/ObjectExtensionMethodsTests.cs
```diff
@@ -1,122 +1,229 @@
...
+                valueU64,
+                value8,
+                value16,
+                value32,
+                value64,
+                valueDecimal,
+                valueFloat,
+                valueDouble
+            };
+            for (int i = 0; i < equalValues.Length; i++)
```
daviburg, line 89:
```suggestion

            for (int i = 0; i < equalValues.Length; i++)
```

Separating logic from bloc of variables declaration.

daviburg:
> This PR has changed entirely, but I wanted to keep some of the above conversation.
> 
> I think the reason we were failing some of the golden liquid tests is that we were considering the string "false" to be Falsy, but the liquid documentation says that only boolean false and nil/null are Falsy. It states strings, including the empty string, are Truthy. The logic of changing one type to another using `ChangeType` also had some effects which takes it away from the standard (i.e. `0` converts to `false`, `"false"` converts to `false`, `"1.2"` converts to `1.2`, etc), and results in statements where `a == b` does not imply `b == a` (as noted in comments by David in the code).
> 
> There are a few side effects to fixing this:
> 
> * strings do not equate to numbers (ie `"1" != 1`)
> * strings do not equate to booleans (i.e. `"false" != false`)
> 
> I think it still needs a bit of work around the numeric comparison. We can probably be better with the logic of promoting types, rather than a blanket `ChangeType` call.
> 
> I've updated the code and tests, but left this as a Draft as we'll need to put this behind a SyntaxCompatibility flag. I just wanted to get feedback before working out how to do that, since this code doesn't have access to the context, so a bit of restructuring will be needed.

This is a big change in behavior. I understand that we are trying to align with the 'golden liquid' standard. Does that however align with the templates' authors intention? Do template authors intend that `"1"` isn't equal to `1`, and `"false"` isn't equal to `false`? The pedantic enforcement of types may run contrary to the value of the templates in solving programming goals in a quick manner. Thus, I don't know if this should be behind a version compatibility flag or a distinct flag to opt in strict type comparison, instead of a best effort to deliver on the template author's assumed intention.

src/DotLiquid.Tests/StandardFilterTests.cs
RodneyRichardson: It's a string, not an integer. And in the new code strings do not equate/match numbers (as per shopify). I changed it from `'7.0'` to match the test just above it (which uses `7`), just to show that it will match the integer but won't match the identical string. We already have a float and an int test. 

https://replit.com/@rrjames/Liquid-Tests shows strings do no equate to numbers.

src/DotLiquid/Util/ObjectExtensionMethods.cs
```diff
@@ -118,25 +126,41 @@ public static bool SafeTypeInsensitiveEqual(this object value, object otherValue
...
-                return Convert.ChangeType(otherValue, aType).Equals(value);
-            }
-            catch (Exception ex) when (ex is InvalidCastException || ex is FormatException || ex is OverflowException)
+            // NOTE(Rodney Richardson): When values are different numeric types we need to try if one can be converted to the other without loss or vice-versa
+            if (IsNumeric(value) && IsNumeric(otherValue))
             {
                 try
                 {
-                    return Convert.ChangeType(value, bType).Equals(otherValue);
+                    var newOtherValue = Convert.ChangeType(otherValue, aType);
```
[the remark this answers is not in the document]
RodneyRichardson: Ah - this was to aid debugging. Splitting into a separate staement allows a breakpoint to evaluation newOtherValue before returning. I'll change it back.

RodneyRichardson:
> This is a big change in behavior. I understand that we are trying to align with the 'golden liquid' standard. Does that however align with the templates' authors intention? Do template authors intend that `"1"` isn't equal to `1`, and `"false"` isn't equal to `false`? The pedantic enforcement of types may run contrary to the value of the templates in solving programming goals in a quick manner. Thus, I don't know if this should be behind a version compatibility flag or a distinct flag to opt in strict type comparison, instead of a best effort to deliver on the template author's assumed intention.

I agree it's a big change - maybe it's something we should leave to a v3.

I don't know how to easily answer the question of "do template authors expect this to work the same as shopify, or would they rather it follow a different internal logic". Doing the same as shopify is easier, but I don't know if it's correct.

It would be good to have a group of users who we could ask to vote about these things. I've only just discovered that it's used as part of Azure Logic Apps (DotLiquid 2.0.361), which seems to be what most of the stackoverflow questions relate to.

daviburg:
> > This is a big change in behavior. I understand that we are trying to align with the 'golden liquid' standard. Does that however align with the templates' authors intention? Do template authors intend that `"1"` isn't equal to `1`, and `"false"` isn't equal to `false`? The pedantic enforcement of types may run contrary to the value of the templates in solving programming goals in a quick manner. Thus, I don't know if this should be behind a version compatibility flag or a distinct flag to opt in strict type comparison, instead of a best effort to deliver on the template author's assumed intention.
> 
> I agree it's a big change - maybe it's something we should leave to a v3.
> 
> I don't know how to easily answer the question of "do template authors expect this to work the same as shopify, or would they rather it follow a different internal logic". Doing the same as shopify is easier, but I don't know if it's correct.
> 
> It would be good to have a group of users who we could ask to vote about these things. I've only just discovered that it's used as part of Azure Logic Apps (DotLiquid 2.0.361), which seems to be what most of the stackoverflow questions relate to.

Yes, Liquid transform is a very popular action in the Azure Logic Apps workflows. We are overdue to update its version. Maybe the Logic Apps Aviators community will be a forum where you can ask other users for input.

daviburg:
> > This is a big change in behavior. I understand that we are trying to align with the 'golden liquid' standard. Does that however align with the templates' authors intention? Do template authors intend that `"1"` isn't equal to `1`, and `"false"` isn't equal to `false`? The pedantic enforcement of types may run contrary to the value of the templates in solving programming goals in a quick manner. Thus, I don't know if this should be behind a version compatibility flag or a distinct flag to opt in strict type comparison, instead of a best effort to deliver on the template author's assumed intention.
> 
> I agree it's a big change - maybe it's something we should leave to a v3.
> 
> I don't know how to easily answer the question of "do template authors expect this to work the same as shopify, or would they rather it follow a different internal logic". Doing the same as shopify is easier, but I don't know if it's correct.
> 
> It would be good to have a group of users who we could ask to vote about these things. I've only just discovered that it's used as part of Azure Logic Apps (DotLiquid 2.0.361), which seems to be what most of the stackoverflow questions relate to.

Actually, we just got an anecdotical answer to this very question on the DotLiquid Google group:

> Basically I'm tying to add 2 values. But one of them is a in Text and the other is in Integer.
>
> This woks if the source is like "IntegerData": 77 but not when like  "IntegerData": "77"
> 
> Ay idea how to get this to work? I used the "to_i" filter to convert to integer but that did not work. 
> 
> Thank you
> 
> Source JSON
> 
> {
>     "content": {
>         "TextData": "44",
>         "IntegerData": "77"
>     }
> }
> 
> Liquid Map 
> 
> {%- assign sum = 10 -%}
> {%- assign sum  =  sum | Plus: content.IntegerData  -%}
> 
> {
>   "add": {{sum}}
> }

RodneyRichardson:
> Actually, we just got an anecdotical answer to this very question on the DotLiquid Google group:

Ooh, that is interesting. Sounds like maybe we need to develop and document a type conversion logic, and implement it consistently. Interchange between numbers and strings seems like a good place to start (i.e. allow strings with numeric representation to be passed in to numeric arguments). Number to string could be harder when dealing with Reals, but maybe we have a default format. And string to bool - I think we could have a flag to control whether `"false"` is considered Truthy or not or just pick the Ruby implementation?

The comparison operators are a little harder, as they don't have an expected type. We should try hard, though, to enforce the standard logical axioms (i.e. `A == B` implies `B == A`, `A == B` and `B == C` implies `A == C`, etc.).

daviburg:
> We should try hard, though, to enforce the standard logical axioms (i.e. `A == B` implies `B == A`, `A == B` and `B == C` implies `A == C`, etc.).

Yes, I agree with the need to follow standard logical axioms.

daviburg:
> Sounds like maybe we need to develop and document a type conversion logic, and implement it consistently. Interchange between numbers and strings seems like a good place to start (i.e. allow strings with numeric representation to be passed in to numeric arguments). Number to string could be harder when dealing with Reals, but maybe we have a default format. And string to bool - I think we could have a flag to control whether `"false"` is considered Truthy or not or just pick the Ruby implementation?

For our own sanity we need to keep order with the flags we are creating overtime, and their default. I think they are divided in categories:
1. Maintain backward compatibility, i.e. don't break users who just upgrade the library
2. Choose between conventions, e.g. Ruby vs .NET
I don't know if Liquid Gold is one of these conventions or a compatibility flag.

PR #588, opened 2025-03-14
Update golden tests to v0.25.0

RodneyRichardson:
Imports latest golden tests (0.25.0) from https://github.com/jg-rp/golden-liquid/blob/main/golden_liquid.json, and updates golden_rules.json to exclude failing new tests.

Also, change skipped_groups to alphabetical order.


microalps (approved):

microalps:
@RodneyRichardson please confirm my changes are good and we will merge this.

RodneyRichardson:
> @RodneyRichardson please confirm my changes are good and we will merge this.

Seems good with one small change. I swapped the order that you check the IDictionary types, so that we check the more specific type first. For example, if we passed in a `Dictionary<string, object>`, it would have followed the `IDictionary` branch instead of the better `IDictionary<string, object>` one.

RodneyRichardson:
I guess this is now changing more than just the golden tests, and will overlap with https://github.com/dotliquid/dotliquid/pull/579. Would it be better to put your changes into a different PR, maybe even #579?

microalps:
> @microalps : Seems good with one small change. I swapped the order that you check the IDictionary types, so that we check the more specific type first. For example, if we passed in a `Dictionary<string, object>`, it would have followed the `IDictionary` branch instead of the better `IDictionary<string, object>` one.

I followed the same order as Context.TryEvaluateHashOrArrayLikeObject. I even considered maybe tapping into the same method. As far as 'more specific', check Hash, we implement both. One is using generics introduced in .NET 3.5 and one predates that I guess. IDictionary<T1, T2> doesn't inherit from IDictionary.

> This is now changing more than just the golden tests, and will overlap with #579. Would it be better to put your changes into a different PR, maybe even #579?

I was forced to fix it otherwise I'd have to specify (incorrectly) the output for hasKey is "FALSE". So they overlap, but doing it separately would be more confusing.

RodneyRichardson:
The generic `IDictionary<K,V>` doesn't inherit from `IDictionary`, but `Dictionary<K, V>` implements both (as probably any concrete implementation would).

RodneyRichardson:
OK - How about that change? It shouldn't matter which order the evaluation is done, and it's in the same order as `Context.TryEvaluateHashOrArrayLikeObject`.

PR #589, opened 2025-03-18
Implement replace_last and remove_last filters

microalps:
Added in Shopify Liquid 5.2.0.
_Note: These do not match the replace_first and remove_first filters rather as identified by Golden Liquid tests._

### DIFF
--- a/src/DotLiquid/StandardFilters.cs
+++ b/src/DotLiquid/StandardFilters.cs
@@ -460,13 +460,17 @@ namespace DotLiquid
         /// <param name="input">Input to be transformed by this filter</param>
         /// <param name="string">Substring to be replaced</param>
         /// <param name="replacement">Replacement string to be inserted</param>
-        [LiquidFilter(MinVersion = SyntaxCompatibility.DotLiquid22)]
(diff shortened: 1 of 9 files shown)


src/DotLiquid/StandardFilters.cs
```diff
@@ -542,6 +542,23 @@ public static string ReplaceFirst(Context context, string input, string @string,
             }, RegexOptions.None, Template.RegexTimeOut);
         }
 
+        /// <summary>
+        /// Replace the first occurrence of a string with another
```
RodneyRichardson, line 546: Should say "Replace the last..."

RodneyRichardson:
I assume this is still a Draft, as it could do with some tests for StandardFilters.

microalps:
> I assume this is still a Draft, as it could do with some tests for StandardFilters.

Not anymore, ready for review now.

RodneyRichardson (changes requested):

microalps:
@RodneyRichardson Corrected code review comments. Merged develop branch and added replace_first and remove_first consistent with golden liquid. Also note, I noted that I incorrectly tagged LegacyFilters.ReplaceFirst as DotLiquid 2.0-2.1 while it should have only been 2.0 as noted in #419

PR #590, opened 2025-03-19
Implement sum filter

RodneyRichardson:
This implementation is slightly more forgiving than Ruby, where if a property is not found then it's treated as zero (whereas I think Ruby throws an error). It also does not flatten nested arrays (which seems like weird behaviour in Ruby).


src/DotLiquid/StandardFilters.cs
```diff
@@ -1015,6 +1017,65 @@ public static IEnumerable Where(IEnumerable input, string propertyName, object t
...
+                        }
+                        else if (double.TryParse(stringValue, out double doubleValue))
+                        {
+                            valueToAdd = doubleValue;
+                        }
+                    }
+                    else if (value is IEnumerable enumerableValue)
+                    {
+                        valueToAdd = Sum(context, enumerableValue);
+                    }
```
RodneyRichardson, line 1072: I'm not sure about this nested behaviour. I didn't have it initially, but a golden liquid test failed. Should we remove it? I think it could result in unexpected behaviours.

RodneyRichardson: Removed.

RodneyRichardson:
@daviburg @microalps Ready for review.

microalps:
It's time to merge my POC so this can go in the new filter test structure. As there is no apparent difference in behavior between contexts, I'm not sure we need to test every version flag.

RodneyRichardson:
> It's time to merge my POC so this can go in the new filter test structure. As there is no apparent difference in behavior between contexts, I'm not sure we need to test every version flag.

`DoMathsOperation` has some different behaviour related to conversion between strings and numbers, which varies by SyntaxCompatibility. Testing each context allowed me to uncover this, and hence put in specific logic for the string to number conversions.

It could possibly do with another context, to handle a different locale (e.g. fr-FR) which uses a different decimal separator, but I thought that was getting towards overkill.

I agree this could wait until the POC is merged, and then update the tests to match.

RodneyRichardson:
> It's time to merge my POC so this can go in the new filter test structure. As there is no apparent difference in behavior between contexts, I'm not sure we need to test every version flag.

@microalps I've rebased this to develop, and have moved my tests into the StandardFiltersTestsBase structure. I wanted some tests using a different culture (to test numeric string conversion), so added a `LatestFrench` test class. I'm not sure if there's a better way to test a matrix of Culture/SyntaxCompatibility?

src/DotLiquid.Tests/Filters/StandardFiltersLatestFrenchTests.cs
```diff
@@ -0,0 +1,21 @@
+using System;
+using System.Globalization;
+using NUnit.Framework;
+
+namespace DotLiquid.Tests.Filters
+{
+    [TestFixture]
+    public class StandardFiltersLatestFrenchTests : StandardFiltersV24Tests
```
daviburg, line 8: Is this a placeholder to add future tests that shows any output difference with French locale?

src/DotLiquid/StandardFilters.cs
```diff
@@ -955,6 +957,61 @@ public static IEnumerable Where(IEnumerable input, string propertyName, object t
...
+                        else if (decimal.TryParse(stringValue, NumberStyles.Float, context.FormatProvider, out decimal decimalValue))
+                        {
+                            valueToAdd = decimalValue;
+                        }
+                        else if (double.TryParse(stringValue, NumberStyles.Float, context.FormatProvider, out double doubleValue))
+                        {
+                            valueToAdd = doubleValue;
+                        }
+                    }
+                    sum = DoMathsOperation(context, sum, valueToAdd, Expression.AddChecked);
```
daviburg, line 1010:
```suggestion

                    sum = DoMathsOperation(context, sum, valueToAdd, Expression.AddChecked);
```

src/DotLiquid/StandardFilters.cs
```diff
@@ -955,6 +957,61 @@ public static IEnumerable Where(IEnumerable input, string propertyName, object t
...
+                }
+            }
+            return sum;
```
daviburg, line 1014:
```suggestion

            return sum;
```

daviburg (approved):

src/DotLiquid.Tests/Filters/StandardFiltersLatestFrenchTests.cs
RodneyRichardson: We could add extra tests, but it also runs the base tests. There is one base test for Sum which uses it for numeric to string testing.

daviburg: The thousand and fragment separators for French are the opposite from English, i.e. 1.000,00 vs 1,000.00

Could you check if that shows up as expected, given this PR is for a math operation?

RodneyRichardson: I think there's an existing bug/unknown around thousands separators, given we use custom Regex for determining if it's integer or real, and then default to InvariantCulture for reals. I will try to investigate what the existing behaviour is. There is inherent ambiguity in there already.

RodneyRichardson: Rasied as #600.

RodneyRichardson:
Merged into #571 to take advantage of NumericConverter.

PR #597, opened 2025-03-21
Add `icon` to nuspec file

RodneyRichardson:
This replaces the obsolete `iconUrl` with `icon` to remove the nuspec warning in the build. See #568.

I've also brought the DotLiquid.csproj file in line with these values, even though we don't use that. I think current best practice is to use the csproj file instead of the nuspec file, so that they are never out of sync. 

Q: Do we want to change the build process to use the csproj file, and delete the nuspec file? Or shall we delete the packaging tags from the csproj file?


microalps:
Should we get rid of LICENSE.txt completely since we are using expressions? Also, seems like iconUrl is recommended to be kept for backwards compatibility with clients that don't support it, and you did that in project file but not nuspec. 

> To maintain backward compatibility with clients and sources that don't yet support icon, specify both icon and iconUrl. Visual Studio supports icon for packages coming from a folder-based source.

Source: https://learn.microsoft.com/en-us/nuget/reference/nuspec

RodneyRichardson:
> Should we get rid of LICENSE.txt completely since we are using expressions? 

I've removed LICENSE.txt and the old readme.txt.

> Also, seems like iconUrl is recommended to be kept for backwards compatibility with clients that don't support it, and you did that in project file but not nuspec.

When I read that before, I thought that only applied to the csproj file. But I think it's both - so I've added it back in.

RodneyRichardson:
Annoyingly, the warning is still there after adding `icon`. But this is still the right thing to do.

microalps:
Can we try dropping the nuspec (and no other change) and compare the build output?

RodneyRichardson:
If I just delete the nuspec file, then the `nuget pack` command will just fail (which is the thing generating the warning). I could try replacing that with a `dotnet pack` command to us the csproj file instead, but I think this is fine.

microalps (approved):

PR #598, opened 2025-03-25
Update golden liquid schema

RodneyRichardson:
Schema of `golden_liquid.json` has changed significantly, although the tests are pretty much the same. I've updated the tests to use the new schema.

I also added test to check the contents of `golden_rules.json` matches actual tests to pick up name changes, removed test in future, etc.


microalps:
Before reviewing, I would like to understand - how do we validate that between definition updates tests are not being removed accidentally? Here are some scenarios that we need to consider:

1. New tests are added and they fail. Accidentally we copy all the failing to be excluded and not just the new ones.
2. Existing test was deleted upstream but was actually useful.
3. Test was renamed and is out of sync with golden rules - that I think you added tests for but not sure if it belongs here.
4. Someone manually commits changes to golden liquid that didn't actually come from upstream

It seems to me we need a script to compare old vs new with a clear process that catches some or all of these issues.

RodneyRichardson:
Can we pick up these manually in the PR diff when there's a change? This commit is a bit different, as there's a significant schema change, but that should work in future. Some of these might also be picked up if there's a difference in the indirect code coverage (i.e. there was a golden liquid test without a corresponding unit test, and now a specific thing is not being tested). 

I don't know enough about the github build system to be able to add an additional check to the PRs, but that might also be a good idea? Maybe just something that checks for deletions in golden-liquid.json?

PR #602, opened 2025-04-08
Fix line endings for .cs files

RodneyRichardson:
.gitattributes shows line endings for .cs and project files should be CRLF. This PR changes the line endings on files that contained LF. There are no code changes in this PR.


RodneyRichardson:
Hmm. I'm not sure this is changing line endings. Perhaps git is already sorting them automatically? I'd be interested to see what you see when checking this out.

PR #603, opened 2025-04-15
Add .NET 10 Target

RodneyRichardson:
Add .Net 8.0 and .Net 10 targets, remove unsupported .Net 6, bump .NET Framework to 4.6.2, and bump some website and analyzer dependencies.

Explicitly add some transitive dependencies to avoid vulnerable versions.

### DIFF
--- a/src/DotLiquid.Tests/DotLiquid.Tests.csproj
+++ b/src/DotLiquid.Tests/DotLiquid.Tests.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net462;netcoreapp3.1;net6.0</TargetFrameworks>
+    <TargetFrameworks>net462;netcoreapp3.1;net6.0;net8.0</TargetFrameworks>
     <AssemblyName>DotLiquid.Tests</AssemblyName>
     <AssemblyOriginatorKeyFile>../Formosatek-OpenSource.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
@@ -42,6 +42,11 @@
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'net6.0' ">
+    <PackageReference Include="NUnit" Version="4.3.2" />
+    <PackageReference Include="NUnit3TestAdapter" Version="5.0.0" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net8.0' ">
     <PackageReference Include="NUnit" Version="4.3.2" />
     <PackageReference Include="NUnit3TestAdapter" Version="5.0.0" />
   </ItemGroup>
--- a/src/DotLiquid.Tests/DotLiquid.Tests.csproj
+++ b/src/DotLiquid.Tests/DotLiquid.Tests.csproj
@@ -36,6 +36,11 @@
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.13.0" />
   </ItemGroup>
 
+  <ItemGroup>
+    <!-- Explicitly add System.Text.RegularExpressions 4.3.1 to avoid vulnerable 4.3.0 transitive dependency -->
+    <PackageReference Include="System.Text.RegularExpressions" Version="4.3.1" />
+  </ItemGroup>
+
   <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp3.1'">
     <PackageReference Include="NUnit" Version="3.14.0" />
     <PackageReference Include="NUnit3TestAdapter" Version="3.17.0" />
--- a/src/DotLiquid.Tests/DotLiquid.Tests.csproj
+++ b/src/DotLiquid.Tests/DotLiquid.Tests.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net462;netcoreapp3.1;net8.0;net10.0</TargetFrameworks>
+    <TargetFrameworks>net462;netcoreapp3.1;net6.0;net8.0;net10.0</TargetFrameworks>
     <AssemblyName>DotLiquid.Tests</AssemblyName>
     <AssemblyOriginatorKeyFile>../Formosatek-OpenSource.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
@@ -44,6 +44,11 @@
   <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp3.1'">
     <PackageReference Include="NUnit" Version="3.14.0" />
     <PackageReference Include="NUnit3TestAdapter" Version="3.17.0" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net6.0' ">
+    <PackageReference Include="NUnit" Version="4.3.2" />
+    <PackageReference Include="NUnit3TestAdapter" Version="5.0.0" />
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'net8.0' ">
--- a/src/DotLiquid.Tests/DotLiquid.Tests.csproj
+++ b/src/DotLiquid.Tests/DotLiquid.Tests.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net462;netcoreapp3.1;net6.0;net8.0</TargetFrameworks>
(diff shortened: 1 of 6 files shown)


RodneyRichardson:
Created #605 to bump test coverage for StrFTime.

RodneyRichardson:
@microalps Does this also require an update to appveyor?

src/DotLiquid.Tests/DotLiquid.Tests.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net462;netcoreapp3.1;net6.0</TargetFrameworks>
+    <TargetFrameworks>net462;netcoreapp3.1;net8.0;net10.0</TargetFrameworks>
```
microalps: We can use net6.0 still to test netstandard 2.0 build. This makes it compatible with .NET Core 2.0+, .NET 5.0+, and .NET Framework 4.6.1+

src/DotLiquid.Website/Views/Home/Index.cshtml
```diff
@@ -12,7 +12,7 @@
     </p>
 
     <h2>Requirements</h2>
-    <p>.NET Framework 4.5 or above, .NET Standard 2.0 or above, or .NET 6.0 or above.</p>
+    <p>.NET Framework 4.6.2 or above, .NET Standard 2.0 or above, or .NET 8.0 or above.</p>
```
microalps: Maybe we should spell out what .NET standard 2.0 means or link https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-0

microalps (changes requested):

microalps:
> @microalps Does this also require an update to appveyor?

Yes, this would require us bumping the image to Visual Studio 2026. We can then use net 6.0 as mentioned above and omit the install of netcoreapp 3.1 making the build even faster.

Specifically update https://github.com/dotliquid/dotliquid/blob/develop/appveyor.yml with version (2.4 not 2.2) image (Visual Studio 2026) and remove lines 21-24. Address the comments and I'll update appveyor accordingly.

microalps:
UPDATE: I noticed that dropping 6.0 also means those using a legacy version silently lose support of DateOnly/TimeOnly even though they are not 'dropped' in reality due to netstandard 2.0 support. This should be treated as a breaking change. I'm not sure if it's worth pursuing. Instead, we should consider dropping netstandard 2.0 at this point and keep 6.0 which happens to be part of the AppVeyor image.

src/DotLiquid.Website/Views/Home/Index.cshtml
RodneyRichardson: This is also a nice link (more visual): https://dotnet.microsoft.com/en-us/platform/dotnet-standard

src/DotLiquid.Tests/DotLiquid.Tests.csproj
RodneyRichardson: I added net6.0 back in.

src/DotLiquid.Website/Views/Home/Index.cshtml
RodneyRichardson: I've updated the text.

RodneyRichardson:
> UPDATE: I noticed that dropping 6.0 also means those using a legacy version silently lose support of DateOnly/TimeOnly even though they are not 'dropped' in reality due to netstandard 2.0 support. This should be treated as a breaking change. I'm not sure if it's worth pursuing. Instead, we should consider dropping netstandard 2.0 at this point and keep 6.0 which happens to be part of the AppVeyor image.

I think supporting the things that Microsoft currently supports is a good rule of thumb. .NET 6 ended support in November 2024. .NET Standard 2.0 will be supported indefinitely (as far as I know) and I believe is the mechanism for having a library that can be used in older unsupported versions of .NET, as well as other flavours (like Mono and Xamarin). 

I think dropping .NET 6 needs to be done at some point - but if considered a breaking change, can be pushed to v3 (hence why this is still draft).

RodneyRichardson:
> > @microalps Does this also require an update to appveyor?
> 
> Yes, this would require us bumping the image to Visual Studio 2026. We can then use net 6.0 as mentioned above and omit the install of netcoreapp 3.1 making the build even faster.
> 

Shall I also remove netcoreapp3.1? My thinking was that we want the tests to have as many target frameworks as is practical, and the library itself have just the currently supported frameworks.

microalps:
Part of this was included in #619 which introduced a compiler directive for net8.0. Also addressed the lack of netcoreapp3.1 on appveyor by building netstandard 2.0 but testing with netcoreapp3.1 or nearest major version.

RodneyRichardson:
I've updated Appveyor to use Visual Studio 2026 image for `dotliquid (dev)`, but this doesn't look to have been picked up. I'll leave it to you, @microalps, to update Appveyor.

src/DotLiquid/DotLiquid.csproj
```diff
@@ -5,7 +5,7 @@
     <AssemblyTitle>DotLiquid</AssemblyTitle>
     <NeutralLanguage>en-US</NeutralLanguage>
     <Authors>Tim Jones;Alessandro Petrelli</Authors>
-    <TargetFrameworks>netstandard2.0;net45;net6.0;net8.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net462;net8.0;net10.0</TargetFrameworks>
```
microalps: I thought we're keeping net6.0 build. Also unclear why we need to remove net45 build - with net462 we can just drop it entirely and rely on netstandard build since that covers 4.6.1+

src/DotLiquid.Website.Tests/DotLiquid.Website.Tests.csproj
```diff
@@ -25,7 +25,13 @@
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.13.0" />
     <PackageReference Include="NUnit" Version="4.3.2" />
     <PackageReference Include="NUnit3TestAdapter" Version="5.0.0" />
-    <PackageReference Include="System.Runtime.InteropServices.RuntimeInformation" Version="4.3.0" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <!-- Explicitly include System.Drawing.Common transitive dependency to avoid vulnerable v4.7.0 -->
+    <PackageReference Include="System.Drawing.Common" Version="4.7.3" />
```
microalps: Transient package updates should occur with centralized package management in Directory.Packages.Props. This has the benefit that should the dependency drop in the future it is not included.

microalps (changes requested):
> I've updated Appveyor to use Visual Studio 2026 image for `dotliquid (dev)`, but this doesn't look to have been picked up. I'll leave it to you, @microalps, to update Appveyor.

Not seeing a file change to AppVeyor.yml. Can you confirm that the latest commits have been pushed?

src/DotLiquid/DotLiquid.csproj
microalps: https://www.appveyor.com/docs/windows-images-software/ lists .NET 6.0 being included, but the build failed with the following:

> Testhost process for source(s) 'C:\projects\dotliquid\src\DotLiquid.Tests\bin\Release\net6.0\DotLiquid.Tests.dll' exited with error: You must install or update .NET to run this application.
App: C:\projects\dotliquid\src\DotLiquid.Tests\bin\Release\net6.0\testhost.exe
Architecture: x64
Framework: 'Microsoft.NETCore.App', version '6.0.0' (x64)
.NET location: C:\Program Files\dotnet
The following frameworks were found:
  8.0.23 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  8.0.28 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  9.0.12 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  10.0.2 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  10.0.9 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]

RodneyRichardson:
Ok - this looks good to me, once AppVeyor works. Looks like .NET 6.0 isn't on the VS2026 image!

src/DotLiquid.Website.Tests/DotLiquid.Website.Tests.csproj
RodneyRichardson: OK - I'll update this from #611.

src/DotLiquid/DotLiquid.csproj
RodneyRichardson: Can we try `choco install dotnet-6.0-sdk -y`? I've added it to the appveyor.yml

microalps (approved):

src/DotLiquid/DotLiquid.csproj
microalps: I modified to be conditional for when they eventually fix the image. Build success. Time took ~8 minutes while previous builds took 6-7 min.

PR #605, opened 2025-04-22
Add tests for unspecified format directives.

RodneyRichardson:
Adds tests to increase coverage flagged in #611.


microalps:
Closed - as per @RodneyRichardson this is merged into #611 to address drop in coverage.

PR #611, opened 2025-07-21
Patch vulnerabilities in transient packages.

RodneyRichardson:
NuGet reports vulnerabilities for various System packages we use indirectly in Tests and Website. This Pull Request explicitly bumps the versions to a non-vulnerable version.


RodneyRichardson:
Looks good to me. Consider merging back with #605 to prevent the drop in code coverage.

PR #619, opened 2026-09-11
Respect decimal scale in output

microalps:
This relates to #566 and our solution was to accept the output differences as correct. The cause for this is traced to efd59f61458219a4f8a8d698cc7a0288cdd2186a where the following comment was made:
> NOTE(David Burg): The decimal type default string serialization behavior adds non-significant trailing zeroes
to indicate the precision of the result.
This is not a desirable default for Liquid as it confuses the users as to why '12.5 |times 10' becomes '125.0'.
So we overwrite the default serialization behavior to specify a format with maximum significant precision.
Decimal type has a maximum of 29 significant digits. 

However, for syntax compatibility we should match liquid reference library even if this is not a 'desirable default'. Gated behind the future 2.4 flag.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -17,11 +17,7 @@ install:
     choco install opencover.portable -y
 
     choco install codecov -y
-
-    dotnet --list-runtimes | find "Microsoft.NETCore.App 3.1" >nul
-
-    if %errorlevel% neq 0 ( choco install dotnetcore-runtime.install --version=3.1.32 -y )
-
+    
     dotnet restore src/DotLiquid.sln
 cache:
 - C:\ProgramData\chocolatey\bin -> appveyor.yml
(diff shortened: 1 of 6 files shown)


microalps:
@daviburg since this behavior was originally made and noted by you, please review and confirm this is OK.

daviburg (approved):
[Dobby] Approved. The core change correctly gates reference-Liquid decimal rendering behind `DotLiquid24` while preserving the legacy default. The four review findings are non-blocking follow-ups:

1. Keep the shipping `net8.0` TFM expansion separate from this behavior fix; the universal `decimal.GetBits` path is sufficient, and PR #603 owns broader TFM policy.
2. Align the comments with the implemented contract: nonzero decimal scale retains at least one fractional digit, while additional trailing zeroes are trimmed.
3. Strengthen regression coverage for `DotLiquid22`/`DotLiquid22a`, the motivating `12.5 | times: 10` case, and non-invariant writer culture.
4. Remove or reconcile the informational `appveyor.yml` hunk with the secured effective script, and remove the introduced trailing whitespace.

These findings do not block approval. AppVeyor and Codecov are green at head `4ec0040`.

